PR #24, opened 2018-06-29
refactor: Apply Prettier, Improve Spelling, Markdown, etc.

dodekeract:
Initially based this on `⑆master`, then had to rebase to `⑆dev`... 😓

Changed a lot of minor issues that I noticed. Mostly tested and works fine for me.

Careful, I had to exclude `KtComment` because the `index.js` seems to be missing entirely.


.editorconfig
```diff
@@ -1,11 +1,11 @@
...
 charset = utf-8
-trim_trailing_whitespace = true
-insert_final_newline = true
 end_of_line = lf
-max_line_length = null
+indent_size = 2
+indent_style = tab
 insert_final_newline = true
+insert_final_newline = true
+max_line_length = 120
```
igeligel: Why is it not 80? Default is 80 and should stay like this.

dodekeract: @igeligel I'm fine with 80, but @ostoc likes longer lines. Given that he's the main author of this repo, I'd consider that to be more important than following an arbitrary standard.

I personally had no issues reading it.

igeligel: > arbitrary standard

This is the community standard. If you line-length reach more than 80 characters, then the code should be refactored since it is not clear anymore and easy to read.

package.json
```diff
@@ -74,5 +65,56 @@
...
+	},
+	"license": "MIT",
+	"lint-staged": {
+		"*.{js,json,css,scss,md,vue}": [
+			"prettier --write",
+			"git add"
+		]
+	},
+	"name": "@3yourmind/kotti",
+	"prettier": {
```
igeligel: Did not we decide company-wide for this configuration:

```json
{
  "useTabs": true,
  "singleQuote": true,
  "trailingComma": "all",
  "arrowParens": "always"
}
```

We should be consistent in our style in all our code bases in my opinion.

packages/kotti-style/package.json
```diff
@@ -1,25 +1,21 @@
...
-		"url": "https://github.com/3YOURMIND/kotti"
-	},
-	"keywords": ["element", "theme"],
 	"author": "3YOURMIND GmbH",
-	"license": "MIT",
 	"bugs": {
 		"url": "https://github.com/3YOURMIND/kotti/issues"
 	},
-	"homepage": "https://3yourmind.github.io/kotti/",
+	"contributors": [
```
igeligel: You can add yourself here.

.editorconfig
igeligel: Because this is what prettier has in mind 👍 

Read through their option philosophy here: https://prettier.io/docs/en/option-philosophy.html

I think with our configuration:

```json
{
  "useTabs": true,
  "singleQuote": true,
  "trailingComma": "all",
  "arrowParens": "always"
}
```

We do not have that many changes to the default prettier configuration.

package.json
dodekeract: @igeligel I copied most of it from the old `.prettierrc`. I'm fine with using `"arrowParens": "always"` though.

igeligel: Yeah, the old prettier configuration file was not up to date. Like I said, we should use the configuration which I posted here in all our projects.

PR #28, opened 2018-07-06
Refactor Table, Improve Accessibility, DX, UX

dodekeract:


packages/kotti-table/src/Table.vue
```diff
@@ -101,118 +78,85 @@ export default {
 		tdClasses: { default: () => [], types: [Array, String, Object] },
 		thClasses: { default: () => [], types: [Array, String, Object] },
 		trClasses: { default: () => [], types: [Array, String, Object] },
-		value: { default: undefined, type: Array }, // @TODO WTF is this?
```
ostoc: do you think we should call it `preSelectedRows`?

dodekeract: @ostoc `value` makes sense since it’s the property used by `v-model`. I also added a comment to the prop so that it’s more obvious what’s going on.

PR #44, opened 2019-01-14
some fixes while using KtTable with remote sort

amrdraz:
- fixed loading state leaving already loaded columns visible
- added support for passing 1 or -1 in the sortedColumn prop as a possible sortOrder value


amrdraz:
@ostoc my ie vm isn't connecting to the internet for some reason (probably I need to update vm ware) so I can't handle this now (other more pressing tasks on my plate)

PR #48, opened 2019-01-24
fix: fix table control on firefox

tomquinonero:
Fix hyphens issue on table controls on firefox.

### DIFF
--- a/packages/kotti-table/src/TableHeader.vue
+++ b/packages/kotti-table/src/TableHeader.vue
@@ -154,6 +154,7 @@ th.sortable .kt-table__controls {
 	top: 0;
 	width: 0.6rem;
 	height: 100%;
+	hyphens: none;
 }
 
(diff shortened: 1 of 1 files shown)


FlorianWendelborn:
@tomquinonero I’d suggest fixing this on yoco’s classes instead by using an `!important`, since this will **never** be good behavior in yoco.

ostoc:
@tomquinonero i found there is no issue on Safari and Firefox. Which version of Firefox you are using?

FlorianWendelborn:
@ostoc For an unknown reason, this bug is only triggered in AMPI. We don’t have any custom css that should interfere here though.

The bug appears in my Firefox Developer Edition.

image

FlorianWendelborn:
@ostoc It would be great if this fix could be moved to the `.yoco` class, as I suggested, or is there ever a reason why it should add hyphens in a ligature?

PR #51, opened 2019-01-29
fix comment component accepting html and opt in parsing

amrdraz:
CommentReply accepts HTML but Comment didn't. Added that and a parser option in order to parse new lines tags

also added docs and allows named exports in main kotti in order to be able to get the KtComment.defaultParser

but then I also realized we could generally make it such that KtComment always escapes and then runs the user passed parser (so that users don't forget and shoot themselves in the foot)

### DIFF
--- a/docs/pages/patterns/comments.vue
+++ b/docs/pages/patterns/comments.vue
@@ -107,14 +107,22 @@ KtComment will escape all tags by default but you can opt out and pass your own 
 you can use KtComment's default parser function with KtComment.defaultParser
 
 ```js
-function newLineParser (msg) { return escape(msg).replace(/\n/g, "<br>") }
+methods: {
+	dangerDefaultParserOverride: msg => escape(msg).replace(/\n/g, '<br>'),
+	// alternativly you could
+	dangerDefaultParserOverride: msg => escape(msg),
+	postEscapeParser: msg => msg.replace(/\n/g, '<br>'),
+	// or just
+	postEscapeParser: msg => msg.replace(/\n/g, '<br>'),
+}
 ```
 
 <div class="element-example">
 	<KtComment
 		v-for="comment in badComments"
 		:key="comment.id"
-	  :parser="parser"
+	  :dangerDefaultParserOverride="dangerDefaultParserOverride"
+		:postEscapeParser="postEscapeParser"
 		:id="comment.id"
 		:createdTime="comment.createdTime"
 		:userName="comment.userName"
@@ -148,7 +156,8 @@ function newLineParser (msg) { return escape(msg).replace(/\n/g, "<br>") }
 | `userId`  | id of user who made the comment to reply too | number, string | "2" | - |
 | `userName`  | name of user to display | string | "Jhone Doe" | - |
 | `message`  | the actual comment | string | "Hello" | - |
-| `parser`             | A function that processes the comment message before it is passed to the div that render it | (string) => string | Function | lodash escape function |
+| `dangerDefaultParserOverride`  | A function that processes and escapes the comment message before it is passed to the div that render it, as the name implies you're responsible for escaping if you use this | (string) => string | Function | lodash escape function |
+| `postEscapeParser`  | A function that processes the message after is has been escaped use this instead of `dangerDefaultParserOverride` | (string) => string | Function | (_) => _ |
 | `allowChange`  | wether this comment is editable | boolean | true,false | false |
 
 ### Event
@@ -242,7 +251,8 @@ export default {
 		},
 	},
 	methods: {
-		parser: msg => escape(msg).replace(/\n/g, '</br>'),
+		dangerDefaultParserOverride: msg => escape(msg),
+		postEscapeParser: msg => msg.replace(/\n/g, '</br>'),
 		handleEdit(payload) {
 			console.log(payload)
 		},
--- a/docs/pages/patterns/comments.vue
+++ b/docs/pages/patterns/comments.vue
@@ -95,12 +95,61 @@
 
 // Delete Payload
 {
-	id: Number
-	message: String
-	parentId: Number
 }
 
 ```
+
+## Parsing HTML 
+
+KtComment will escape all tags by default but you can opt out and pass your own parser by using the parser prop 
+
+> Remember to still escape tags to prevent XHR attacks,
+you can use KtComment's default parser function with KtComment.defaultParser
+
+```js
+function newLineParser (msg) { return escape(msg).replace(/\n/g, "<br>") }
+```
+
+<div class="element-example">
+	<KtComment
+		v-for="comment in badComments"
+		:key="comment.id"
+	  :parser="parser"
+		:id="comment.id"
+		:createdTime="comment.createdTime"
+		:userName="comment.userName"
+		:userAvatar="comment.userAvatar"
+		:userId="comment.userId"
+		:message="comment.message"
+		:replies="comment.replies"
+		:allowChange="comment.allowChange"
+		@delete="handelDelete($event)"
+		@edit="handleEdit($event)"
+		@submit="handleSubmit($event)"
+	/>
+	<KtCommentInput
+		class="mt-16px"
+		placeholder="Add comment"
+		userAvatar='https://picsum.photos/120'
+		@submit="handleSubmit($event)"
+	/>
+</div>
+
+### Props
+
+		allowChange: Boolean,
+
+| Attribute            | Description                                 | Type                        | Accepted values                 | Default |
+|:---------------------|:--------------------------------------------|:----------------------------|:--------------------------------|:--------|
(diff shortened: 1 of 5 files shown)


packages/kotti-comment/src/Comment.vue
```diff
@@ -9,7 +9,7 @@
 			<div
 				v-if="!isInlineEdit"
 				class="comment__message"
-				v-text="inlineMessage"
+				v-html="inlineMessage"
```
FlorianWendelborn: This means that Kotti, by default, would be cross-site-scripting vulnerable. I don’t think that’s a good idea since it will eventually lead to someone having a **critical** security vulnerability.

For example, if I wouldn’t have seen this PR and upgraded Kotti after this is merged, any user in AMPI could comment `<script>alert('pwned')</script>` and execute arbitrary javascript in the browser of anyone seeing their comment.

Quoting from The Vue documentation:

> Dynamically rendering arbitrary HTML on your website can be very dangerous because it can easily lead to XSS vulnerabilities. Only use HTML interpolation on trusted content and **never on user-provided content.**
> — https://vuejs.org/v2/guide/syntax.html#Raw-HTML

That’s also why I **really** prefer [React’s equivalent of `v-html`](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml):

```jsx
function createMarkup() {
  return {__html: 'First &middot; Second'};
}

function MyComponent() {
  return <div dangerouslySetInnerHTML={createMarkup()} />;
}
```

If you **absolutely** need this feature, you should make it opt-in and use a similarly "scary" name as React, for example by adding a `dangerouslyEnableHTML` property to `<KtComment/>`.

ostoc: @FlorianWendelborn  Correct me if I am wrong, since vue.js 2, the `v-html` already blocked dangerous scripts and only allow few html tags.

amrdraz: @FlorianWendelborn how's now

docs/pages/patterns/comments.vue
```diff
@@ -95,13 +95,69 @@
...
+## Parsing HTML 
+
+KtComment will escape all tags by default but you can opt out and pass your own parser by using the parser prop 
+
+> Remember to still escape tags to prevent XHR attacks,
+you can use KtComment's default parser function with KtComment.defaultParser
+
+```js
+methods: {
+	dangerDefaultParserOverride: msg => escape(msg).replace(/\n/g, '<br>'),
```
FlorianWendelborn: Should be `dangerouslyOverrideParser`
```suggestion
	dangerouslyOverrideParser: msg => escape(msg).replace(/\n/g, '<br>'),
```

docs/pages/patterns/comments.vue
```diff
@@ -95,13 +95,69 @@
...
+	// alternativly you could
+	dangerDefaultParserOverride: msg => escape(msg),
```
FlorianWendelborn:
```suggestion
	dangerouslyOverrideParser: msg => escape(msg),
```

docs/pages/patterns/comments.vue
```diff
@@ -95,13 +95,69 @@
...
+	// or just
+	postEscapeParser: msg => msg.replace(/\n/g, '<br>'),
+}
+```
+
+<div class="element-example">
+	<KtComment
+		v-for="comment in badComments"
+		:key="comment.id"
+	  :dangerDefaultParserOverride="dangerDefaultParserOverride"
```
FlorianWendelborn:
```suggestion
	:dangerouslyOverrideParser="dangerouslyOverrideParser"
```

docs/pages/patterns/comments.vue
```diff
@@ -167,6 +249,8 @@ export default {
 		},
 	},
 	methods: {
+		dangerDefaultParserOverride: msg => escape(msg),
```
FlorianWendelborn:
```suggestion
		dangerouslyOverrideParser: msg => escape(msg),
```

docs/pages/patterns/comments.vue
```diff
@@ -140,7 +222,7 @@ export default {
 							userName: 'Benni',
 							createdTime: '2018-03-20',
 							message:
-								'Join Bright Side Now! Join Bright Side Now! Join Bright Side Now! Join Bright Side Now!',
+								'Join Bright Side Now!</br>Join Bright Side Now! Join Bright Side Now! Join Bright Side Now!',
```
FlorianWendelborn:
```suggestion
								'Join Bright Side Now!<br/>Join Bright Side Now! Join Bright Side Now! Join Bright Side Now!',
```

packages/kotti-comment/src/Comment.vue
```diff
@@ -9,7 +9,7 @@
 			<div
 				v-if="!isInlineEdit"
 				class="comment__message"
-				v-text="inlineMessage"
+				v-html="postEscapeParser(dangerDefaultParserOverride(inlineMessage))"
```
FlorianWendelborn:
```suggestion
				v-html="postEscapeParser(dangerouslyOverrideParser(inlineMessage))"
```

packages/kotti-comment/src/Comment.vue
```diff
@@ -46,6 +46,8 @@
 					:message="reply.message"
 					:createdTime="reply.createdTime"
 					:allowChange="reply.allowChange"
+					:dangerDefaultParserOverride="dangerDefaultParserOverride"
```
FlorianWendelborn:
```suggestion
					:dangerouslyOverrideParser="dangerouslyOverrideParser"
```

packages/kotti-comment/src/Comment.vue
```diff
@@ -74,6 +77,8 @@ import KtCommentInput from './CommentInput.vue'
 export default {
 	name: 'KtComment',
 	props: {
+		dangerDefaultParserOverride: { default: escape, type: Function },
```
FlorianWendelborn:
```suggestion
		dangerouslyOverrideParser: { default: escape, type: Function },
```

packages/kotti-comment/src/CommentReply.vue
```diff
@@ -12,7 +12,12 @@
 					v-if="!isInlineEdit"
 					@click="$emit('_inlineReplyClick', { userName, userId })"
 				>
-					<span v-html="inlineMessage" /> <i class="yoco" v-text="'comment'" />
+					<span
+						v-html="
+							postEscapeParser(dangerDefaultParserOverride(inlineMessage))
```
FlorianWendelborn:
```suggestion
							postEscapeParser(dangerouslyOverrideParser(inlineMessage))
```

packages/kotti-comment/src/CommentReply.vue
```diff
@@ -53,6 +59,9 @@ export default {
 		KtButtonGroup,
 	},
 	props: {
+		dangerDefaultParserOverride: { default: escape, type: Function },
```
FlorianWendelborn:
```suggestion
		dangerouslyOverrideParser: { default: escape, type: Function },
```

docs/pages/patterns/comments.vue
```diff
@@ -95,13 +95,69 @@
...
-	parentId: Number
 }
 
 ```
 
+## Parsing HTML 
+
+KtComment will escape all tags by default but you can opt out and pass your own parser by using the parser prop 
+
+> Remember to still escape tags to prevent XHR attacks,
```
FlorianWendelborn:
```suggestion
> Remember to **escape malicious tags** to prevent [Cross-site-scripting](https://en.wikipedia.org/wiki/Cross-site_scripting) attacks,
```

docs/pages/patterns/comments.vue
```diff
@@ -95,13 +95,69 @@
...
+you can use KtComment's default parser function with KtComment.defaultParser
+
+```js
+methods: {
+	dangerDefaultParserOverride: msg => escape(msg).replace(/\n/g, '<br>'),
+	// alternativly you could
+	dangerDefaultParserOverride: msg => escape(msg),
+	postEscapeParser: msg => msg.replace(/\n/g, '<br>'),
```
FlorianWendelborn:
```suggestion
	postEscapeParser: msg => msg.replace(/\n/g, '<br/>'),
```

docs/pages/patterns/comments.vue
```diff
@@ -95,13 +95,69 @@
...
+	// or just
+	postEscapeParser: msg => msg.replace(/\n/g, '<br>'),
```
FlorianWendelborn:
```suggestion
	postEscapeParser: msg => msg.replace(/\n/g, '<br/>'),
```

packages/kotti-comment/src/Comment.vue
FlorianWendelborn: @ostoc It’s possible that `vue` attempted to block parts of the issue, but `` works.

PR #54, opened 2019-01-31
Feature/disable table row

amrdraz:
Add support for setting disabling of some rows in a table through a function


ostoc:
Nice job 👍

PR #56, opened 2019-02-01
suport multiline flag for inlineEdit component

amrdraz:
I thought just allowing the parser would have been enough but it seems that the inlineEdit component used input which wouldn't allow users to have multiple lines

Added a multiline flag that would switch the input field to a text area and default parser to newLineParser.

I ported the component into Yoda in order to test it in components so it's not urgent to publish a new version after this merges

### DIFF
--- a/packages/kotti-inline-edit/src/InlineEdit.vue
+++ b/packages/kotti-inline-edit/src/InlineEdit.vue
@@ -3,14 +3,16 @@
 		<label class="form-label" v-text="label" />
 		<div
 			v-if="!editMode"
-			v-html="postEscapeParser(dangerouslyOverrideParser(representValue))"
+			v-html="postParser(dangerouslyOverrideParser(representValue))"
 			@click="editMode = true"
 			:class="representTextClass"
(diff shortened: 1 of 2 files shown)


ostoc:
@amrdraz did you give an example in the docs ?

packages/kotti-inline-edit/src/InlineEdit.vue
```diff
@@ -37,8 +41,9 @@ export default {
 	},
 	props: {
 		dangerouslyOverrideParser: { default: escape, type: Function },
-		postEscapeParser: { default: _ => _, type: Function },
+		postEscapeParser: { default: DEFAULT_POST_PARSER, type: Function },
 		invalidMessage: { default: 'Click to edit', type: String },
+		multiline: { deafult: false, type: Boolean },
```
FlorianWendelborn: Typo here!

ostoc: please raise an issue, will do a hot fix

Issue #63, opened 2019-02-28
KtNavBar Border color change


justinline:
**Is your feature request related to a problem? Please describe.**
The borders for KtNavBar have opacity but doesn't work well with KtNavTheme overrides

**Describe the solution you'd like**
Would be great to have the Borders calculated based on whether the background color is light or dark :)

ostoc:
@justinline The API changes will change when next release, you don't need to specify the borderColor anymore. 

**Prvider name changed from `KtNavbarTheme` to `KtTheme`**

Default Theme Provider

```js
provide() {
	return {
		KtTheme: {
			navbarBackgroundColor: '#122C56',
			navbarTextColor: 'rgba(255,255,255,.54)',
			navbarTextActiveColor: 'rgba(255,255,255, 1)',
			logoUrl: LogoSvg,
			logoHeight: '40px',
		},
	}
},
```
This change should be also important to @FlorianWendelborn @tomquinonero

PR #64, opened 2019-02-28
feature: change KtNavbarTheme to KtTheme #63

ostoc:
The API changes will change when next release, you don't need to specify the borderColor anymore.

**Prvider name changed from `KtNavbarTheme` to `KtTheme`**

Default Theme Provider

```js
provide() {
	return {
		KtTheme: {
			navbarBackgroundColor: '#122C56',
			navbarTextColor: 'rgba(255,255,255,.54)',
			navbarTextActiveColor: 'rgba(255,255,255, 1)',
			logoUrl: LogoSvg,
			logoHeight: '40px',
		},
	}
},
```

### DIFF
--- a/docs/pages/patterns/navbar.vue
+++ b/docs/pages/patterns/navbar.vue
@@ -82,29 +82,31 @@ quickLinksData: {
 
 `KtNavbar` injects the theme provided by the parent or globally.
 
-When `KtNavbarTheme` object is provided, it will replace the default theme.
+When `KtTheme` object is not provided, it will replace the default theme.
 
 Currently supports:
 
-- `activeColor`: the color when menu is slected, the active class by default has `font-weight: 600`
-- `backgroundColor`: the background color of the navbar
+- `navbarTextColor`: the default menu text color of the navbar, make sure the `textColor` has enough contrast with background
+- `navbarTextActiveColor`: the color when menu is slected, the active class by default has `font-weight: 600`
+- `navbarBackgroundColor`: the background color of the navbar
 - `logoUrl`: the url of the logo image
-- `textColor`: the default menu text color of the navbar, make sure the `textColor` has enough contrast with background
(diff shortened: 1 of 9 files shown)


docs/pages/patterns/navbar.vue
```diff
@@ -82,29 +82,31 @@ quickLinksData: {
 
 `KtNavbar` injects the theme provided by the parent or globally.
 
-When `KtNavbarTheme` object is provided, it will replace the default theme.
+When `KtTheme` object is not provided, it will replace the default theme.
```
FlorianWendelborn: I don’t understand this sentence since you added the `not`.

packages/kotti-navbar/src/KtNavbar.vue
```diff
@@ -94,15 +98,12 @@ export default {
...
-			default: () => ({
-				backgroundColor: '#122C56',
-				textColor: 'rgba(255,255,255, 0.8)',
-				activeColor: 'rgba(255,255,255, 1)',
-				borderColor: '#rgba(255,255,255, 0.14)',
-				logoUrl: null,
-			}),
+		KtTheme: {
+			default: {
+				navbarBackgroundColor: '#122C56',
```
FlorianWendelborn: I’d suggest moving this (partially repeated, and logically belonging together) piece of code to a utility file (e.g. `default-theme.json`). You can just import that everywhere where you need the default theme then. 🙂

docs/pages/patterns/navbar.vue
ostoc: when there is no `KtTheme` object provided. Kotti will use default theme (the dark blue one)

packages/kotti-navbar/src/KtNavbar.vue
ostoc: yep, a good idea. can do when we have more theme need to be configured.

PR #72, opened 2019-04-16
Add checkbox component (fix #71)

### LINKED ISSUE #71: Support KtCheckbox component
We actually do not have any checkbox components.
It is, although, mentioned there: <https://3yourmind.github.io/kotti/components/inputs>.
I think that it is just been forgot.

tomquinonero:
Add a component for checkboxes. 
Also add the documentation for it and add it to the NavBar.

### DIFF
--- a/docs/pages/components/checkbox.vue
+++ b/docs/pages/components/checkbox.vue
@@ -22,7 +22,7 @@ We can also specify the label as a default scope:
 ```html
 <div class="element-example">
 	<KtCheckbox v-model="value2">
-		I agree to the <a href="http://google.com">Terms of usage</a>
+		I agree to the <a href="https://3yourmind.com">Terms of usage</a>
 	</KtCheckbox>
 </div>
 ```
--- a/docs/pages/components/checkbox.vue
+++ b/docs/pages/components/checkbox.vue
@@ -1,0 +1,42 @@
+<template lang="md">
+
+# Radio
+
+## Basic
+
+<div class="element-example">
+	<KtCheckbox :value="true" label="Good Morning" />
+	<KtCheckbox :value="false">
+		Good Night <a href="https://en.wikipedia.org/wiki/Good_Night">Link</a>
+	</KtCheckbox>
+	<h4>Value: {{radio1}}</h4>
+</div>
+
+</template>
+
+<script>
+export default {
+	name: 'KtCheckboxDoc',
+	data() {
+		return {
+			radio1: 'morning',
+			radio2: 'option1',
+			title: 'Test',
+			options: [
+				{
+					label: 'Good Day',
+					value: 'option1',
+				},
+				{
+					label: 'Good Night',
+					value: 'option2',
+				},
+				{
+					label: 'Good Morning',
+					value: 'option3',
+				},
+			],
+		}
+	},
+}
+</script>
--- a/docs/pages/components/checkbox.vue
+++ b/docs/pages/components/checkbox.vue
@@ -30,7 +30,7 @@ We can also specify the label as a default scope:
 	<KtCheckbox v-model="value2">
 		I agree to the <a href="http://google.com">Terms of usage</a>
 	</KtCheckbox>
-		<h4>Value: {{value2}}</h4>
+	<h4>Value: {{value2}}</h4>
 </div>
 
 ### Provide an `initialValue`
--- a/docs/pages/components/checkbox.vue
+++ b/docs/pages/components/checkbox.vue
@@ -28,7 +28,7 @@ We can also specify the label as a default scope:
 ```
 <div class="element-example">
 	<KtCheckbox v-model="value2">
-		I agree to the <a href="http://google.com">Terms of usage</a>
+		I agree to the <a href="https://3yourmind.com">Terms of usage</a>
 	</KtCheckbox>
 	<h4>Value: {{value2}}</h4>
 </div>
--- a/docs/pages/components/checkbox.vue
+++ b/docs/pages/components/checkbox.vue
@@ -4,7 +4,7 @@
 
 ## Basic usage
 
-The `KtCheckbox` component needs a `v-model` and a `label` props. An optional `name` can be used.
+The `KtCheckbox` component requires the `v-model` and `label` properties. Other arguments (e.g. `name`) get directly passed to the `<input>` element.
 
 ```html
 <div class="element-example">
--- a/docs/pages/components/checkbox.vue
+++ b/docs/pages/components/checkbox.vue
@@ -33,22 +33,6 @@ We can also specify the label as a default scope:
 	<h4>Value: {{value2}}</h4>
 </div>
 
-### Provide an `initialValue`
-
-In certain cases we may wat to have a default value in case the one used by `v-model` is `undefined` (while it's retrieved for example).
-We can then define `initialValue` props:
-``` html
-<div class="element-example">
-	<KtCheckbox v-model="value3" :initialValue="true" label="Will you give everything up for kotti?" />
-	<h4>Value: {{value3}}</h4>
-</div>
-```
-<div class="element-example">
-	<KtCheckbox v-model="value3" :initialValue="true" label="Will you give everything up for kotti?" />
-	<h4>Value: {{value3}}</h4>
-</div>
-In the example above, `value3` is `undefined` and will default to `true`.
-
 ## Properties
 
 | Attribute            | Description                                   | Type      | Accepted values | Default |
@@ -56,11 +40,8 @@ In the example above, `value3` is `undefined` and will default to `true`.
 | `label`              | label input field                             | `String`  | —               | —       |
 | `name`               | name attribute for the field                  | `String`  | —               | —       |
 | `value`              | value                                         | `Boolean` | —               | —       |
-| `initialValue`       | initial value (in case `value` is not defined)| `Boolean` | —               | —       |
-
 
 </template>
-
 <script>
 export default {
 	name: 'KtCheckboxDoc',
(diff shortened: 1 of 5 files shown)


docs/pages/components/checkbox.vue
```diff
@@ -0,0 +1,75 @@
+<template lang="md">
+
+# Checkbox
+
+## Basic usage
+
+The `KtCheckbox` component needs a `v-model` and a `label` props. An optional `name` can be used.
```
FlorianWendelborn: fix grammar pls

FlorianWendelborn: `$attrs` is better than explicitly supporting some attributes to be passed

docs/pages/components/checkbox.vue
```diff
@@ -0,0 +1,75 @@
...
+	<KtCheckbox v-model="value2">
+		I agree to the <a href="http://google.com">Terms of usage</a>
+	</KtCheckbox>
+</div>
+```
+<div class="element-example">
+	<KtCheckbox v-model="value2">
+		I agree to the <a href="http://google.com">Terms of usage</a>
+	</KtCheckbox>
+		<h4>Value: {{value2}}</h4>
```
FlorianWendelborn:
```suggestion
	<h4>Value: {{value2}}</h4>
```

docs/pages/components/checkbox.vue
```diff
@@ -0,0 +1,75 @@
...
+</div>
+
+### Provide an `initialValue`
+
+In certain cases we may wat to have a default value in case the oe used by `v-model` is `undefined` (while it's retrieved for example).
```
FlorianWendelborn:
```suggestion
In certain cases we may wat to have a default value in case the one used by `v-model` is `undefined` (while it's retrieved for example).
```

docs/pages/components/checkbox.vue
```diff
@@ -0,0 +1,75 @@
...
+We can then define `initialValue` props:
+``` html
+<div class="element-example">
+	<KtCheckbox v-model="value3" :initialValue="true" label="Will you give everything up for kotti?" />
```
FlorianWendelborn: Why does `initialValue` exist if you have a `v-model`?

docs/pages/components/checkbox.vue
```diff
@@ -0,0 +1,75 @@
...
+In the example above, `value3` is `undefined` and will default to `true`.
+
+## Properties
+
+| Attribute            | Description                                   | Type      | Accepted values | Default |
+|:---------------------|:----------------------------------------------|:----------|:----------------|:--------|
+| `label`              | label input field                             | `String`  | —               | —       |
+| `name`               | name attribute for the field                  | `String`  | —               | —       |
+| `value`              | value                                         | `Boolean` | —               | —       |
+| `initialValue`       | initial value (in case `value` is not defined)| `Boolean` | —               | —       |
```
FlorianWendelborn: I think this should be removed. And *if* you support this it should be explicit (as in `null`).

packages/kotti-checkbox/src/Checkbox.vue
```diff
@@ -0,0 +1,42 @@
...
+	<label class="form-checkbox">
+		<input
+			class="form-checkbox__input"
+			v-model="model"
+			:name="name"
+			type="checkbox"
+		/>
+		<i class="form-icon" />
+		<span class="form-checkbox__label">
+			<slot>{{ label }}</slot>
```
FlorianWendelborn:
```suggestion
			<slot v-text="label"/>
```

packages/kotti-checkbox/src/Checkbox.vue
```diff
@@ -0,0 +1,42 @@
...
+		</span>
+	</label>
+</template>
+
+<script>
+export default {
+	name: 'KtCheckbox',
+	props: {
+		label: { types: [String] },
```
FlorianWendelborn:
```suggestion
		label: { required: true, type: String },
```

packages/kotti-checkbox/src/Checkbox.vue
```diff
@@ -0,0 +1,42 @@
...
+		name: { types: [String] },
```
FlorianWendelborn:
```suggestion
		name: { default: null, types: [String, null] },
```

packages/kotti-checkbox/src/Checkbox.vue
```diff
@@ -0,0 +1,42 @@
...
+		value: { types: [Boolean] },
```
FlorianWendelborn:
```suggestion
		value: { required: true, type: Boolean },
```

or

```suggestion
		value: { default: null, types: [Boolean, null] },
```

packages/kotti-checkbox/src/Checkbox.vue
```diff
@@ -0,0 +1,42 @@
...
+		initialValue: { types: [Boolean] },
```
FlorianWendelborn:
```suggestion
```

packages/kotti-checkbox/src/Checkbox.vue
```diff
@@ -0,0 +1,42 @@
...
+	},
+	computed: {
+		model: {
+			get() {
+				let value
```
FlorianWendelborn: If you for whatever reason don’t want to remove support for `initialValue`, at least do this:

```suggestion
				return this.value === null ? this.initialValue : this.value
```

packages/kotti-checkbox/src/Checkbox.vue
```diff
@@ -0,0 +1,42 @@
...
+				if (typeof this.value == 'boolean') {
```
FlorianWendelborn:
```suggestion
				if (typeof this.value === 'boolean') {
```

packages/kotti-checkbox/src/Checkbox.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+	<label class="form-checkbox">
+		<input v-model="model" :name="name" :value="value" type="checkbox" />
```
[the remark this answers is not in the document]
tomquinonero: Only using v-model here, should be fine.

packages/kotti-checkbox/src/Checkbox.vue
tomquinonero: The initial value was asked by @amrdraz.
Anyway I think your suggestion differs too much from the original goal. Just imagine value being `"I'm a String"`. Plus the style here is explicit does what it seems to do, even if it needs few lines for it.

tomquinonero: Screen Shot 2019-04-17 at 11 44 49 AM

docs/pages/components/checkbox.vue
```diff
@@ -0,0 +1,56 @@
...
+	<KtCheckbox v-model="value1" label="Will you give everything up for kotti?" />
+	<h4>Value: {{value1}}</h4>
+</div>
+
+We can also specify the label as a default scope:
+
+```html
+<div class="element-example">
+	<KtCheckbox v-model="value2">
+		I agree to the <a href="http://google.com">Terms of usage</a>
```
FlorianWendelborn:
```suggestion
		I agree to the <a href="https://3yourmind.com">Terms of usage</a>
```

docs/pages/components/checkbox.vue
```diff
@@ -0,0 +1,56 @@
...
+	</KtCheckbox>
+</div>
+```
+<div class="element-example">
+	<KtCheckbox v-model="value2">
+		I agree to the <a href="http://google.com">Terms of usage</a>
```
FlorianWendelborn:
```suggestion
		I agree to the <a href="https://3yourmind.com">Terms of usage</a>
```

docs/pages/components/checkbox.vue
```diff
@@ -0,0 +1,56 @@
...
+	<h4>Value: {{value2}}</h4>
+</div>
+
+## Properties
+
+| Attribute            | Description                                   | Type      | Accepted values | Default |
+|:---------------------|:----------------------------------------------|:----------|:----------------|:--------|
+| `label`              | label input field                             | `String`  | —               | —       |
+| `name`               | name attribute for the field                  | `String`  | —               | —       |
+| `value`              | value                                         | `Boolean` | —               | —       |
```
FlorianWendelborn: Please re-format table. It’s too wide.

docs/pages/components/checkbox.vue
```diff
@@ -0,0 +1,56 @@
...
+
+</template>
+<script>
+export default {
+	name: 'KtCheckboxDoc',
+	data() {
+		return {
+			value1: true,
+			value2: false,
+			value3: undefined,
```
FlorianWendelborn: `undefined` should not be an allowed value.

packages/kotti-checkbox/src/Checkbox.vue
FlorianWendelborn: Yes, vue is unfortunately not 100% strict about this. However, you’re misusing the API and you’re getting a warning in the console that strings aren’t allowed there. Should we also support `MediaStream` values? 🙂

packages/kotti-checkbox/src/Checkbox.vue
FlorianWendelborn: @tomquinonero the `label` seems to be required, judging from the template. Otherwise, add `default: null` and add a `v-if="label !== null"`.

FlorianWendelborn:
#71

packages/kotti-checkbox/src/Checkbox.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+	<label class="form-checkbox">
+		<input
+			class="form-checkbox__input"
+			v-model="model"
+			type="checkbox"
+			v-bind="$attrs"
+		/>
+		<i class="form-icon" />
+		<span class="form-checkbox__label">
```
amrdraz: make label optional

PR #75, opened 2019-04-24
thClass implementation + typo fix

AZeghers:
thClass is now properly implemented on KtTableColumn

### DIFF
--- a/docs/pages/components/tables.vue
+++ b/docs/pages/components/tables.vue
@@ -584,7 +584,7 @@ slot-scope is not required for the `loading` and `empty` slots
 | `hiddenColumns`          | prop for setting hidden columns                                     | `Array`                     | [{ prop, hidden }]    | `[]`    |
 | `filterdColumns`         | prop for setting filterd columns                                    | `Array`                     | [{ prop, filter }]    | `[]`    |
 | `tdClasses`              | classes to apply to all `<td>` elements                             | `Array`, `String`, `Object` | `"responsive"`        | `[]`    |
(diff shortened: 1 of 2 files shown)


docs/pages/components/tables.vue
```diff
@@ -584,7 +584,7 @@ slot-scope is not required for the `loading` and `empty` slots
 | `hiddenColumns`          | prop for setting hidden columns                                     | `Array`                     | [{ prop, hidden }]    | `[]`    |
 | `filterdColumns`         | prop for setting filterd columns                                    | `Array`                     | [{ prop, filter }]    | `[]`    |
 | `tdClasses`              | classes to apply to all `<td>` elements                             | `Array`, `String`, `Object` | `"responsive"`        | `[]`    |
-| `thClass`                | classes to apply to all `<th>` elements                             | `Array`, `String`, `Object` | `"responsive"`        | `[]`    |
```
amrdraz: in addition to thClasses there's also thClass which applies to only the column not all columns
so maybe we should asdd that in the docs

also not it's always possible to target `.kt-table  th[prop="propname"]`

AZeghers: thClass already exists in the column docs, it was just not implemented

This change fixes the fact that thClasses in the KtTable attributes was wrongfully called thClass

amrdraz: oh guess I missed it due to the local scope of the comment ok all clear then

FlorianWendelborn:
fix #75

PR #76, opened 2019-05-02
updated KtTable

### LINKED ISSUE #73: KtTable Always Activates when using Tab
**Describe the bug**
When using an interactive `KtTable`, tabbing through the page will trigger a `activateRow` event (same as a click would). This effectively prevents accessibility entirely instead of helping with accessibility. For example, in AMPI, this means that tabbing through the parts list always opens the first part.

**To Reproduce**
1. Go to any interactive `KtTable`.
2. Tab
3. Notice that the app thinks you’ve clicked

**Expected behavior**
Kotti should only fire such an event if the user actually uses `⏎`.

**Screenshots**
N/A

 - OS: any
 - Browser: any
 - Version: 1.5.1

**Smartphone (please complete the following information):**
 - Device: any
 - OS: any
 - Browser: any
 - Version: 1.5.1

**Additional context**
Here are the lines that are causing the issue: https://github.com/3YOURMIND/kotti/blob/7219b57d84cbb31898e4fc34bc58f134a222febb/packages/kotti-table/src/TableRow.js#L156-L157

amrdraz:
Fix Column order swapping instead of setting ordering in correct spot
Fix Reusable columns appearing our of order due to none unique keys
Fix #73
Fix clicking on actions triggering rowClick
Added KtConsumer to be used with KtProvider instead of Mixin solution for clearer DX
Added KtProvider to the docs

### DIFF
--- a/docs/pages/components/tables.vue
+++ b/docs/pages/components/tables.vue
@@ -378,9 +378,9 @@ You can then use the set sortedColumns prop on the KtTable to update the table u
 ```html
 <KtTable :rows="rows" :columns="columns">
 	<template slot-scope="{ row }" slot="actions">
-		<i class="yoco" @click="showAlert(row.name, 'edited')">edit</i>
-		<i class="yoco" @click="showAlert(row.name, 'deleted')">trash</i>
-	</template>
+	<i class="yoco" @click="showAlert(row.name, 'edited')">edit</i>
+	<i class="yoco" @click="showAlert(row.name, 'deleted')">trash</i>
+</template>
 </KtTable>
 ```
 
@@ -392,17 +392,17 @@ Same as `actions`, `slot-scope` is required and is used to pass data to the expa
 
 <KtTable :rows="rows" :columns="columnsResponsive" isExpandable isScrollable>
 	<template slot-scope="{ row, rowIndex }" slot="expand">
-		<KtBanner :message="row.name" icon="user" isGrey/>
-		<KtBanner :message="row.address.line" icon="global" isGrey/>
-	</template>
+	<KtBanner :message="row.name" icon="user" isGrey />
+	<KtBanner :message="row.address.line" icon="global" isGrey />
+</template>
 </KtTable>
 
 ```html
 <KtTable :rows="rows" :columns="columns" isExpandable isScrollable>
 	<template slot-scope="{ row, rowIndex }" slot="expand">
-		<KtBanner :message="row.name" icon="user" isGrey/>
-		<KtBanner :message="row.address.line" icon="global" isGrey/>
-	</template>
+	<KtBanner :message="row.name" icon="user" isGrey />
+	<KtBanner :message="row.address.line" icon="global" isGrey />
+</template>
 </KtTable>
 ```
 
@@ -557,20 +557,20 @@ slot-scope is not required for the `loading` and `empty` slots
 		prop="name"
 		>
 		<template slot="header" slot-scope="{ value, column, columnIndex }">
-			<div>
-				{{value}}
-			</div>
-		</template>
+	<div>
+		{{ value }}
+	</div>
+</template>
 		<template slot-scope="{ value, row, rowIndex, column, columnIndex }">
-			<KtAvatar
-				name={value}
-				hoverable
-				src="https://picsum.photos/200"
-				showTooltip
-				small
-				class="mr-16px"
-			/>
-		</template>
+	<KtAvatar
+		:name="value"
+		hoverable
+		src="https://picsum.photos/200"
+		showTooltip
+		small
+		class="mr-16px"
+	/>
+</template>
 	</KtTableColumn>
 </KtTable>
 ```
@@ -634,8 +634,8 @@ Otherwise all tables under the same Provider will share the same store
 
 ### Attributes
 
-| Attribute                | Description                                                         | Type                        | Accepted values       | Default |
-|:-------------------------|:--------------------------------------------------------------------|:----------------------------|:----------------------|:--------|
+|        Attribute         |                             Description                             |            Type             |    Accepted values    | Default |
+| :----------------------- | :------------------------------------------------------------------ | :-------------------------- | :-------------------- | :------ |
 | `rows` (required)        | table row data                                                      | `Array`                     | —                     | —       |
 | `id`                     | for when using multiple provider                                    | `String`                    | —                     | null    |
 | `rowKey`                 | the row prop used for the rows key                                  | `String`                    | —                     | —       |
@@ -671,8 +671,8 @@ Otherwise all tables under the same Provider will share the same store
 
 ### Column Attributes
 
-| Attribute          | Description                                                                    | Type                        | Accepted values                            | Default                           |
-|:-------------------|:-------------------------------------------------------------------------------|:----------------------------|:-------------------------------------------|:----------------------------------|
+|     Attribute      |                                  Description                                   |            Type             |              Accepted values               |              Default              |
+| :----------------- | :----------------------------------------------------------------------------- | :-------------------------- | :----------------------------------------- | :-------------------------------- |
 | `prop`  (required) | used to match the value in `rows` can be a dot path                            | `String` `String.String`    | —                                          | —                                 |
 | `align`            | alignment of column text                                                       | `String`                    | `"center"`, `"left"`, `"right"`            | `left`                            |
 | `key`              | deprecated is transalted to prop instead                                       | `String`                    | —                                          | —                                 |
@@ -701,16 +701,16 @@ Otherwise all tables under the same Provider will share the same store
 
 #### TableConsumer Attributes
 
-| Attribute                | Description                      | Type                        | Accepted values       | Default |
-|:-------------------------|:---------------------------------|:----------------------------|:----------------------|:--------|
-| `id`                     | for when using multiple provider | `String`                    | —                     | null    |
+| Attribute |           Description            |   Type   | Accepted values | Default |
+| :-------- | :------------------------------- | :------- | :-------------- | :------ |
+| `id`      | for when using multiple provider | `String` | —               | null    |
 
 ### Events
 
 #### Column
 
-| Event Name           | Arguments                                             | Description                                                         |
-|:---------------------|:------------------------------------------------------|:--------------------------------------------------------------------|
+|      Event Name      |                       Arguments                       |                             Description                             |
+| :------------------- | :---------------------------------------------------- | :------------------------------------------------------------------ |
 | `@activateRow`       | `(row, index)`                                        | Row was clicked or activated via keyboard.                          |
 | `@rowClick`          | `(row, index)`                                        | Row was clicked                                                     |
 | `@rowFocus`          | `(row, index)`                                        | Row was in focus Requires setting `isInteractive` or `@activateRow` |
@@ -724,33 +724,33 @@ Otherwise all tables under the same Provider will share the same store
 
 #### Column
 
-| Event Name   | Arguments                                         | Description        |
-|:-------------|:--------------------------------------------------|:-------------------|
+|  Event Name  |                     Arguments                     |    Description     |
+| :----------- | :------------------------------------------------ | :----------------- |
 | `@cellClick` | `({ value, column, row, columnIndex, rowIndex })` | a cell was clicked |
 
 ### Slots
 
 #### Table
 
-| Slot Name | Description                         | Scope                                         |
-|:----------|:------------------------------------|:----------------------------------------------|
-| `empty`   | what to render when no data         | --                                            |
-| `loading` | what to render when loading is true | --                                            |
-| `actions` | action section of each row          | `{value, row, rowIndex }                      |
-| `expand`  | expand section of each row          | `{value, row, rowIndex }                      |
+| Slot Name |             Description             |          Scope           |
+| :-------- | :---------------------------------- | :----------------------- |
+| `empty`   | what to render when no data         | --                       |
+| `loading` | what to render when loading is true | --                       |
+| `actions` | action section of each row          | `{value, row, rowIndex } |
+| `expand`  | expand section of each row          | `{value, row, rowIndex } |
 
 #### Column
 
-| Slot Name | Description          | Scope                                         |
-|:----------|:---------------------|:----------------------------------------------|
+| Slot Name |     Description      |                     Scope                     |
+| :-------- | :------------------- | :-------------------------------------------- |
 | `header`  | render in table row  | `{value, row, rowIndex }`                     |
 | `default` | render in table cell | `{value, row, rowIndex, column, columnIndex}` |
 
 #### TableConsumer
 
-| Slot Name | Description                               | Scope                                                                                                           |
-|:----------|:------------------------------------------|:----------------------------------------------------------------------------------------------------------------|
-| `default` | provide a table's store and other methods | `{store,columns, hiddenColumns, sortedColumns, filteredColumns, hideColumn, showAllColumns, orderBeforeColumn,}`|
+| Slot Name |                Description                |                                                      Scope                                                       |
+| :-------- | :---------------------------------------- | :--------------------------------------------------------------------------------------------------------------- |
+| `default` | provide a table's store and other methods | `{store,columns, hiddenColumns, sortedColumns, filteredColumns, hideColumn, showAllColumns, orderBeforeColumn,}` |
 
 ### Store
 
--- a/docs/pages/components/tables.vue
+++ b/docs/pages/components/tables.vue
@@ -239,6 +239,34 @@ based on your view and specific row data
 }
 ```
 
+## Row Interaction
+
+`KtTable` has a `@rowClick` event for when a user clicks but perfered event is `@activateRow`,
+which gets fired on row click as well as when the user hits the enter/return key while focused.
+
+You can tab and trigger row active by the hitting enter (not you can't select disabled rows)
+
+
+<KtInput v-model='disableName' />
+<KtTable
+	:rows="rows"
+	:columns="columnsResponsive"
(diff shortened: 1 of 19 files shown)


docs/pages/components/tables.vue
```diff
@@ -559,13 +575,69 @@ slot-scope is not required for the `loading` and `empty` slots
 </KtTable>
 ```
 
+### Provider/Consumer and Mixin
+
+Sometimes you may need to access the table's store and control it from outside,
+while `ref` may work if you're modifications are in the same component, you controling ui may be elsewhere
```
FlorianWendelborn:
```suggestion
while `ref` may work if your modifications are in the same component, your controlling ui may be elsewhere
```

docs/pages/components/tables.vue
```diff
@@ -559,13 +575,69 @@ slot-scope is not required for the `loading` and `empty` slots
...
+For that purpose we introduce `KtTableProvider` `KtTableConsumer` and optionally but discourged we expose `KtTableColumnsStateMixin`
```
FlorianWendelborn:
```suggestion
For that purpose we introduce `KtTableProvider` `KtTableConsumer`. There’s also the deprecated `KtTableColumnsStateMixin`.
```

docs/pages/components/tables.vue
```diff
@@ -559,13 +575,69 @@ slot-scope is not required for the `loading` and `empty` slots
...
+
+<KtTableProvider>
+	<div>
+		<KtTableConsumer #default="{ columns, hideColumn }">
```
FlorianWendelborn: I’m assuming that’s a typo. Also, next time please don’t use a reserved keyword for an API.

```suggestion
		<KtTableConsumer default="{ columns, hideColumn }">
```

docs/pages/components/tables.vue
```diff
@@ -559,13 +575,69 @@ slot-scope is not required for the `loading` and `empty` slots
...
+			<div class="parts-edit-columns-filter__container">
+				<KtButtonGroup>
+						<KtButton
+							v-for="column in columns"
+							:key="column.prop"
+							:label="column.label || column.prop"
+							:type="column.hidden?'text':'primary'"
```
FlorianWendelborn:
```suggestion
							:type="column.hidden ? 'text' : 'primary'"
```

docs/pages/components/tables.vue
```diff
@@ -559,13 +575,69 @@ slot-scope is not required for the `loading` and `empty` slots
...
+	</div>
+	<div>
+		<KtTable :rows="rows" :columns="columnsDefault" />
+	</div>
+</KtTableProvider>
+
+```html
+<KtTableProvider>
+	<div>
+		<KtTableConsumer #default="{ columns, hideColumn }">
```
FlorianWendelborn: Same as above.

```suggestion
		<KtTableConsumer default="{ columns, hideColumn }">
```

docs/pages/components/tables.vue
```diff
@@ -559,13 +575,69 @@ slot-scope is not required for the `loading` and `empty` slots
...
+			<div class="parts-edit-columns-filter__container">
+				<KtButtonGroup>
+						<KtButton
+							v-for="column in columns"
+							:key="column.prop"
+							:label="column.label || column.prop"
+							:type="column.hidden?'text':'primary'"
```
FlorianWendelborn:
```suggestion
							:type="column.hidden ? text' : 'primary'"
```

docs/pages/components/tables.vue
```diff
@@ -626,16 +698,23 @@ slot-scope is not required for the `loading` and `empty` slots
 | `headerCellClass`  | classes to this colum's to `.kt-table__header-cell` elements                   | `Array`, `String`, `Object` | `"responsive"`                             | `[]`                              |
 | `cellClass`        | classes to this colum's to `.kt-table__cell` elements                          | `Array`, `String`, `Object` | `"responsive"`                             | `[]`                              |
 
+
+#### TableConsumer Attributes
+
+| Attribute                | Description                      | Type                        | Accepted values       | Default |
```
FlorianWendelborn: This markdown table is too big. You should apply a formatter.

docs/pages/components/tables.vue
```diff
@@ -653,19 +732,31 @@ slot-scope is not required for the `loading` and `empty` slots
...
 #### Table
 
-| Slot Name | Description                         |
-|:----------|:------------------------------------|
-| `empty`   | what to render when no data         |
-| `loading` | what to render when loading is true |
-| `actions` | action section of each row          |
-| `expand`  | expand section of each row          |
+| Slot Name | Description                         | Scope                                         |
+|:----------|:------------------------------------|:----------------------------------------------|
```
FlorianWendelborn: Apply a formatter to this table. It’s too big.

docs/pages/components/tables.vue
```diff
@@ -653,19 +732,31 @@ slot-scope is not required for the `loading` and `empty` slots
...
+
+#### TableConsumer
+
+| Slot Name | Description                               | Scope                                                                                                           |
+|:----------|:------------------------------------------|:----------------------------------------------------------------------------------------------------------------|
+| `default` | provide a table's store and other methods | `{store,columns, hiddenColumns, sortedColumns, filteredColumns, hideColumn, showAllColumns, orderBeforeColumn,}`|
+
+### Store
+
+The table store is exposed through the KtTableConsumer for in ui update as well as the KtTableColumnStateMixin for in component access.
```
FlorianWendelborn: > for in ui update

I have no idea what this is supposed to tell me. Please clarify.

packages/kotti-table/src/Table.vue
```diff
@@ -46,9 +46,10 @@ export default {
 	components: { TableBody, TableHeader, TableColumn },
 	name: 'KtTable',
 	props: {
+		id: { default: null, type: String },
 		rowKey: { type: String },
-		rows: { required: true, type: Array },
-		columns: { default: () => [], type: Array },
+		rows: { default: () => [], type: Array },
```
FlorianWendelborn: Does a table without `rows` make sense?

packages/kotti-table/src/Table.vue
```diff
@@ -46,9 +46,10 @@ export default {
...
+		columns: { type: Array },
```
FlorianWendelborn: Why remove the `default` here, and further-down create an implicit default? Why is this not a `required` property if it doesn’t have a `default`?

packages/kotti-table/src/Table.vue
```diff
@@ -148,8 +159,8 @@ export default {
 			return Boolean(this.$scopedSlots.actions || this.renderActions)
 		},
 		_renderExpand() {
+			const table = this
```
FlorianWendelborn: Is this `var that = this` actually necessary? Why?

packages/kotti-table/src/TableColumn.js
```diff
@@ -63,16 +63,21 @@ const TableColumn = {
...
 		prop: updateColumnsfor('prop'),
 		hidden: updateColumnsfor('hidden'),
+		order: updateColumnsfor('order'),
 		default: updateColumnsfor('default'),
 	},
 	mounted() {
-		this.columnConfig.index =
-			this.columnConfig.index || [].indexOf.call(this[KT_TABLE].$children, this)
-		this[KT_STORE].commit('insertColumn', this.columnConfig)
+		const columnIndex = [].indexOf.call(this[KT_TABLE].$children, this)
```
FlorianWendelborn: Why is this `.call` necessary?

packages/kotti-table/src/TableColumn.js
```diff
@@ -105,6 +110,7 @@ function createColumn(column = {}) {
 
 	column.id = columnId
 	column.type = COLUMN_TYPE
+	column._deleted = false
```
FlorianWendelborn: Any explanation would be appreciated.

packages/kotti-table/src/TableRow.js
```diff
@@ -23,19 +23,22 @@ export default {
...
 			hasActions,
 			renderActions,
 			isDisabled,
 		} = this
 		return (
 			<tr
 				class={_trClasses}
-				role={isInteractive ? 'button' : false}
-				tabindex={isInteractive ? 0 : false}
+				role={isInteractive && !isDisabled ? 'button' : false}
```
FlorianWendelborn: Why would you tab to the row if the row is disabled? Why would it act as a button if it’s disabled?

packages/kotti-table/src/TableRow.js
```diff
@@ -145,20 +158,39 @@ export default {
 			$event.stopPropagation()
 			this[KT_STORE].commit('expandRow', row)
 		},
+		handleActionsClick($event) {
+			$event.stopPropagation()
```
FlorianWendelborn: This should be  a `@click.stop`.

packages/kotti-table/src/TableRow.js
```diff
@@ -145,20 +158,39 @@ export default {
 			$event.stopPropagation()
```
FlorianWendelborn: This should be a `@select.stop`.

packages/kotti-table/src/TableRow.js
```diff
@@ -145,20 +158,39 @@ export default {
...
 		},
 		handleFocus($event, row, index) {
-			if (this[KT_TABLE].isInteractive && !this.isDisabled) {
-				this[KT_TABLE].$emit('activateRow', row, index)
+			if (this.isInteractive && !this.isDisabled) {
+				this[KT_STORE].commit('focuseRow', row)
 				this[KT_TABLE].$emit('rowFocus', row, index)
 			}
 		},
+		handelActivation($event, row, index) {
```
FlorianWendelborn:
```suggestion
		handleActivation($event, row, index) {
```

packages/kotti-table/src/logic/column.js
```diff
@@ -1,94 +1,101 @@
...
+		if (column) {
+			column._deleted = true
+			store.state._destroyedColumns[column.prop] = 1
+			store.commit('updateColumns', { emitChange: false })
+		}
 	},
-	updateColumns({ table, state }) {
+	setColumns(store, columns = store.state._columnsArray) {
+		const { state } = store
+		for (let col of columns) {
```
FlorianWendelborn:
```suggestion
		const pickedColumns = columns.map(column => pick(column, PUBLIC_COLUMN_PROPS)
		for (const column of pickedColumns) {
```

FlorianWendelborn: Or perhaps even use a `.map().forEach()`.

packages/kotti-table/src/logic/column.js
```diff
@@ -1,94 +1,101 @@
...
-		let oldColumn = _columns[oldColumnIndex]
-		_columns.splice(column.index, 1, { ...oldColumn, ...column })
-	} else if (index !== undefined) {
-		// else place at this exact index
-		_columns.splice(index, 0, column)
-	} else if (column.order !== undefined) {
-		// else place at order location
-		if (_columns.length > column.order) {
-			_columns.splice(column.order, 0, column)
+	let oldColumn = _columns[column.prop]
```
FlorianWendelborn:
```suggestion
	const oldColumn = _columns[column.prop]
```

packages/kotti-table/src/logic/column.js
```diff
@@ -99,41 +106,43 @@ export function setColumn(state, column, index) {
...
+export function setColumnsArray(
+	state,
+	prop,
+	shapeKeys,
+	columns,
+	mergeStrategy = Object.assign,
+) {
 	state[prop] = columns.map(column => {
 		column = pick(column, shapeKeys)
 		let oldColumn = getColumn(state, column) || {}
```
FlorianWendelborn:
```suggestion
		const oldColumn = getColumn(state, column) || {}
```

packages/kotti-table/src/logic/order.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export const getters = {
+	getOrderedColumns(state) {
+		return getColumnsArray(state, 'orderedColumns')
+	},
+}
+
+export function resolveColumnsOrder({ _columns = {}, orderedColumns = [] }) {
+	for (let col of orderedColumns) {
```
FlorianWendelborn:
```suggestion
	for (const col of orderedColumns) {
```

packages/kotti-table/src/logic/order.js
```diff
@@ -0,0 +1,76 @@
...
+		_columns[col.prop] = col.order || _columns[col.prop]
+	}
+	return Object.values(_columns)
+		.filter(({ _deleted }) => !_deleted)
+		.map((col, index) => {
+			return {
```
FlorianWendelborn: This arrow function can be compacted. `() => ({})`.

packages/kotti-table/src/logic/order.js
```diff
@@ -0,0 +1,76 @@
...
+				orderAdvantage: 'order' in col ? 1 : -1,
+				order: col.order || col.index,
+				index: col.index || index,
+				col,
+			}
+		})
+		.sort((a, b) =>
```
FlorianWendelborn: This should be a separate function with some comments.

packages/kotti-table/src/logic/order.js
```diff
@@ -0,0 +1,76 @@
...
+			a.order < b.order
+				? -1
+				: a.order > b.order
+				? 1
+				: a.orderAdvantage + a.index - (b.orderAdvantage + b.index),
+		)
+		.map(({ col }, order) => {
+			col.order = order
+			col.index = order
+			return col
```
FlorianWendelborn:
```suggestion
			return {...col, order, index: order }
```

And delete all lines above. Then compact the arrow function. This has the advantage that it treats its arguments as immutable, preventing weird bugs.

packages/kotti-table/src/logic/row.js
```diff
@@ -2,23 +2,22 @@ import { sortData } from './sort'
...
 
 export const mutations = {
-	setRows(store, rows) {
+	setRows(store, data) {
 		const { state } = store
-		const { _rows } = state
-		const dataInstanceChanged = _rows !== rows
+		const { _data } = state
+		const dataInstanceChanged = _data !== data
 		// eslint-disable-next-line no-underscore-dangle
```
FlorianWendelborn: Is this required?

packages/kotti-table/src/TableRow.js
amrdraz: you shouldn't be able to tab to the row if it's disabled, so if disabled is true you won't be able too

not isDisabled means that if isDisabled is true then tabindex is false 

The button one may be unnecessary to switch from being evaluated as a button to not, but I figured not to confuse screen readers

packages/kotti-table/src/TableRow.js
amrdraz: but this is a js file not a view file so I have to handle events manually

packages/kotti-table/src/TableRow.js
amrdraz: same thing, jsx doesn't have the vue syntax sugare

packages/kotti-table/src/Table.vue
amrdraz: I recall I had a reason when I was fixing one of the bugs I'll reset it to default to null and see what happens

packages/kotti-table/src/Table.vue
amrdraz: no, but I recall needing a row array to always exist for the table internal state, so it was either initialise it here, or initialise it in the store.

packages/kotti-table/src/logic/column.js
amrdraz: is the goal to remove the for loop ? 
cause the execution time is more or less the same

ah I see the intent is to separate the pick step from the set step

amrdraz: column is not picked column, the picked columns are assigned (as in the prop values are set to the column, mutating it)
I recognise the seperation of picking and then mutating

packages/kotti-table/src/logic/order.js
amrdraz: this is an intentional mutation
columns are referenced in the table's observable state, if I create a new one I would need to set it with Vue.set

the map only exists to return the same column after the mutation is applied
I could have forEach if it returned the original array

packages/kotti-table/src/logic/row.js
amrdraz: I had eslint complain when I accessed an underscored field, so yes

FlorianWendelborn: Ah, I see, so it’s about `object._`, not about `_`.

packages/kotti-table/src/logic/order.js
FlorianWendelborn: I hope we won’t run into weird bugs because of this in the future. 😟

packages/kotti-table/src/TableColumn.js
amrdraz: It would if $children isn't a proper array but a DOMNode array

But I don't think $children is I'll have to test it, I had taken that part originally as is form element ui when I was looking at other libraries for their solutions

will review

packages/kotti-table/src/TableColumn.js
amrdraz: whenever columns are hidden and shown or reorderd the are unmounted
this unmounting causes the column to be removed along with all it's state and then later added again, but at the end out of order since it's not longer there

explicitly setting the order isn't practical and I couldn't think of a more practical idea then to simply mark the column as deleted until it is restored again in the next cycle

the column is eventually removed when the entire table is removed, and you can never have 2 columns with the same prop key so no memory leak should occur 

should I add a comment stating needed for maintaining column state across re-renders

PR #79, opened 2019-05-20
improvement: use tilds ~ for dependencies version.

tomquinonero:
In semver, the first number stands for major change, the second for minor, the third for patch.
Caret "^" means only update patches, tild "~" means update also minor version.
Minor version does not break anything so let's use minor version updates.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,14 +31,19 @@
 			"email": "jo@3yourmind.com",
 			"name": "James O'Shea",
 			"url": "https://github.com/jamesoshea"
+		},
+		{
+			"email": "tquinonero@gmail.com",
+			"name": "Tom Quinonero",
+			"url": "https://github.com/tomquinonero"
 		}
 	],
 	"dependencies": {
-		"@3yourmind/yoco": "^1.1.0",
-		"color": "^3.1.0",
-		"dayjs": "^1.8.8",
-		"vue-clickaway": "^2.2.2",
-		"vue-popperjs": "^2.0.3"
+		"@3yourmind/yoco": "~1.1.0",
+		"color": "~3.1.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -31,14 +31,19 @@
...
+		},
+		{
+			"email": "tquinonero@gmail.com",
+			"name": "Tom Quinonero",
+			"url": "https://github.com/tomquinonero"
 		}
 	],
 	"dependencies": {
-		"@3yourmind/yoco": "^1.1.0",
-		"color": "^3.1.0",
```
mes3yd: (1) I don't know how stable these other packages are, but IMHO the ones under our control should be ~. cc @amrdraz

package.json
```diff
@@ -31,14 +31,19 @@
...
-		"dayjs": "^1.8.8",
-		"vue-clickaway": "^2.2.2",
-		"vue-popperjs": "^2.0.3"
+		"@3yourmind/yoco": "~1.1.0",
+		"color": "~3.1.0",
+		"dayjs": "~1.8.8",
+		"vue-clickaway": "~2.2.2",
```
amrdraz:
```suggestion
		"vue-clickaway": "^2.2.2",
```

package.json
```diff
@@ -31,14 +31,19 @@
...
+		"vue-popperjs": "~2.0.3"
```
amrdraz:
```suggestion
		"vue-popperjs": "^2.0.3"
```

package.json
amrdraz: I'm with @mes3yd I don't particularly trust, all packages, I've added a suggestion to change back the ones think may not respect minor version

PR #83, opened 2019-05-24
Kt select improvement

### LINKED ISSUE #41: Improve UI and Require Support in KtSelect

In some design there is icon on the left of the KtSelect, but doesn't support it now.

Currently, KtSelect Ignores `required`, making it impossible for the user to tell if the select needs input or not.

tomquinonero:
This PR contains a refactorement allowing us to put the SCSS into the component.
I talked with @ostoc and the reason the style was into a `scss` file was for legacy reason. We can now use the mixins and variables inside components.

But the big part of this PR is a change in the `KtSelect`. It allows us to display the select options outside of the box it is.

From my commit message:
```
An element inside another element using overflow
cannot get outside of it. For our select to work
the only solution is to put the element elsewhere.
```

I know this PR is kind of controversial, because listening on window and playing with the DOM is not very Vue, but this is the best solution in my opinion.

I also partially resolved #41


packages/kotti-single-select/src/SingleSelect.vue
```diff
@@ -68,8 +74,24 @@ export default {
...
 			visible: false,
+			formOptions: null,
+			formOptionsContainer: null,
+			input: null,
+			formOptionStyle: '',
 		}
 	},
+	mounted() {
+		this.formOptions = this.$refs['formOptions']
+		this.input = this.$refs['input']
```
amrdraz: @tomquinonero  this will result in memory leaks as the ref-ed elements can be changed when doing partial rerendering.

the element you have appended to the body and refer to in the variables would not reflect the elements that are rendered by vue.

I would recommend you watch the $refs.formOptions and $refs.input and make sure to run destroy and recreate logic when they change, seeing that you're using v-show instead of v-if would I guess hide this bug though, so perhaps it is not as critical but is never the less present.

packages/kotti-single-select/src/SingleSelect.vue
```diff
@@ -68,8 +85,24 @@ export default {
 			queryString: '',
 			selectedLabel: '',
 			visible: false,
+			formOptions: null,
+			formOptionsContainer: null,
+			input: null,
+			formOptionStyle: '',
 		}
 	},
+	mounted() {
```
FlorianWendelborn: This doesn’t handle scrolling, adding elements that move the component, etc..

FlorianWendelborn: For example, you can verify this by just deleting something in your debugger:
image

Please use a library for this, as correctly implementing this behavior is insanely complicated.

packages/kotti-single-select/src/SingleSelect.vue
```diff
@@ -68,8 +85,24 @@ export default {
 			queryString: '',
 			selectedLabel: '',
 			visible: false,
+			formOptions: null,
```
FlorianWendelborn: Not sorted.

packages/kotti-single-select/src/SingleSelect.vue
```diff
@@ -109,6 +142,17 @@ export default {
 		},
 	},
 	methods: {
+		computeFormOptionsStyle() {
+			let top =
```
FlorianWendelborn:
```suggestion
			const top =
```

packages/kotti-single-select/src/SingleSelect.vue
```diff
@@ -109,6 +142,17 @@ export default {
...
+				this.input.getBoundingClientRect().top -
+				this.formOptionsContainer.getBoundingClientRect().top +
+				this.input.offsetHeight
+			let left =
```
FlorianWendelborn:
```suggestion
			const left =
```

packages/kotti-single-select/src/SingleSelect.vue
```diff
@@ -109,6 +142,17 @@ export default {
...
+				this.input.getBoundingClientRect().left -
+				this.formOptionsContainer.getBoundingClientRect().left
+			let width = this.input.offsetWidth
```
FlorianWendelborn:
```suggestion
			const width = this.input.offsetWidth
```

packages/kotti-single-select/src/SingleSelect.vue
tomquinonero: It is actually not, and I fixed it.

amrdraz: @FlorianWendelborn has a point, there will definitely be some weird IE edge cases, this is why I was under the impression that popper handles this, even if it doesn't render in the body you can make it go into the body

packages/kotti-single-select/src/SingleSelect.vue
amrdraz: you can 
```
watch() {
  '$res.formOptions'(formOptions) {
    this.formOptions = formOptions
    if (!formOptions) {
       // remove the element from body like in the destroy hook
    } else {
      // get select option DOMNode in body, if it's present replace it with new one, if it's not present create it.
    }
  }
}
```

note that you'll need to add an element to capture the events being bubbled (like click) and funnel it back into the select component, otherwise the click events are leaking outside in the body and not being captured by the select (hence why I recommended using vue-portal)

packages/kotti-single-select/src/SingleSelect.vue
tomquinonero: Please give me real cases because I cannot fix it otherwise. I will improve the popover and use it in the select in the future, but that is an improvement and I cannot do it right now.

PR #84, opened 2019-05-26
Updated KtCheckbox to support color strings

Goldziher:


packages/kotti-checkbox/src/Checkbox.vue
```diff
@@ -17,11 +17,18 @@ export default {
...
 	computed: {
-	        hasLabel() {
-			return !(this.label===null || this.label === undefined)
+		hasLabel() {
+			return !(this.label === null || this.label === undefined)
+		},
+		iconColor() {
+			return {
+				backgroundColor: this.color ? this.color : 'none',
+				borderColor: this.color ? this.color : 'none',
```
amrdraz: should none be the default?

amrdraz: also you need to add an example in the docs

Goldziher: example added. None should be the default.

FlorianWendelborn: Colors should consume the KtTheme by default. Anything else means a shitload of work for any app trying to use consistent colors.

ostoc: Actually there is also a requirement to change the color of the select box. if you knew the design in `AdminPanel` there is colorful checkbox to enable different graph.

PR #85, opened 2019-05-26
Added color attribute to KtCheckbox and updated the docs

Goldziher:
This PR adds a color attribute to KtCheckbox, which receives a string (Hex or RGB). If no color is provided the component renders as usual. I also included update to the docs.

### DIFF
--- a/docs/pages/components/checkbox.vue
+++ b/docs/pages/components/checkbox.vue
@@ -4,7 +4,9 @@
 
 ## Basic usage
 
-The `KtCheckbox` component requires the `v-model` and `label` properties. Other arguments (e.g. `name`) get directly passed to the `<input>` element.
+The `KtCheckbox` component requires the `v-model` and `label` properties.
+The 'color' property is optional, if set it will pass the color code to the '<i>' element.
+Other arguments (e.g. `name`) get directly passed to the `<input>` element.
 
 ```html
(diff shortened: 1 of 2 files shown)


tomquinonero:
So code looks good. 

But philosophically I think CSS is the way to go here. Any reason that pushed you to do this that  way @Goldziher ?

Goldziher:
Description:

This PR adds a color attribute to KtCheckbox and updates the docs to reflect this change.

tomquinonero:
> Well, I'm using this component by tying it to state - and i am using this state in another component simultaneously. So adding the color attribute allows me to use less code.
> 
> Other then this my thinking is along these lines:
> 
> 1. The color attribute allows dynamic color changes. Hence this will extend support for themes - if and when we introduce these.
Not really ok, theming should be done via CSS.
> 2. We might need to use colored checkboxes in multiple places in the future.
I don't see why but again, CSS.
> 3. Overriding deep css is rather cumbersome, especially when you have to do it in multiple components.
Not really accurate, CSS is made for this. 

I kind of disagree with inline CSS, but I guess you need it ? Thoughts @amrdraz @FlorianWendelborn ?

tomquinonero:
I'm closing this PR as @Goldziher decided to go with CSS.

PR #86, opened 2019-05-31
Improve navbar link by using vue-router links

tomquinonero:
The navbar was using hardcoded `a` tags for handling links. That is not good.
We can now handle link in navbar like any other link using `router-link`.
We can then benefit from `active` class and we can use names for routes instead of hardcoded paths.
We can also support external links using `isExternal` props.


FlorianWendelborn:
This was an intentional decision to reduce the assumptions about the environment where Kotti runs in.

packages/kotti-navbar/src/KtNavbarLink.vue
```diff
@@ -0,0 +1,29 @@
...
+		</a>
+		<router-link
+			v-else
+			@click="$emit('menuLinkClick', link)"
+			v-bind="$attrs"
+			active-class="is-active"
+			:to="link"
+		>
+			<slot />
+		</router-link>
```
amrdraz: We prefer to not have a peer dependency if possible, so consider implementing a scooped slot instead or the dynamic component element 

```
<component 
  :is="linkComponent"   // defaults to 'a'
  @click="$emit('menuLinkClick', link)"
  v-bind="$attrs"
  active-class="is-active"
  :to="link"
  :href="link"
/>
```

amrdraz: If you want to allow KtNavBarLink to consume state from the Navbar while building customizable implementations of it then you can use the provider/inject pattern to allow users the ability to create and reuse their own custom NabarLink component that will work with the KtNavbar by wrapping it

amrdraz: the dynamic component suggestion I made is not completely thought through and may need reworking (in case we're passing the a tag too many props

packages/kotti-navbar/src/KtNavbar.vue
```diff
@@ -148,3 +159,16 @@ export default {
...
+<style lang="scss">
+.navbar {
+	a,
+	a:hover,
+	a:focus,
+	a:visited {
+		color: inherit;
+	}
+	a.is-active {
+		color: var(--navbar-text-active-color);
```
FlorianWendelborn: Not compatible with internet explorer: https://caniuse.com/#feat=css-variables

packages/kotti-navbar/src/KtNavbar.vue
```diff
@@ -129,6 +129,17 @@ export default {
...
 		},
 	},
+	mounted() {
+		let parent = document.querySelector('html')
+		parent.style.setProperty(
+			'--navbar-text-active-color',
+			this.KtTheme.navbarTextActiveColor,
+		)
+		parent.style.setProperty(
+			'--navbar-text-color',
```
FlorianWendelborn: Not compatible with Internet Explorer: https://caniuse.com/#feat=css-variables

packages/kotti-navbar/src/KtNavbar.vue
```diff
@@ -129,6 +129,17 @@ export default {
 			}
 		},
 	},
+	mounted() {
+		let parent = document.querySelector('html')
+		parent.style.setProperty(
+			'--navbar-text-active-color',
```
FlorianWendelborn: Not compatible with Internet Explorer: https://caniuse.com/#feat=css-variables

packages/kotti-navbar/src/KtNavbar.vue
```diff
@@ -129,6 +129,17 @@ export default {
 			}
 		},
 	},
+	mounted() {
+		let parent = document.querySelector('html')
```
FlorianWendelborn:
```suggestion
		const parent = document.querySelector('html')
```

docs/pages/patterns/navbar.vue
```diff
@@ -57,7 +60,7 @@ menuData: [
 				isActive: false,
 			},
 			{
-				link: '#',
+				link: {name: 'myRouteName'},
```
FlorianWendelborn:
```suggestion
				link: { name: 'myRouteName' },
```

packages/kotti-navbar/src/KtNavbarLink.vue
```diff
@@ -0,0 +1,29 @@
...
+	<div class="navbar-link">
+		<a
+			v-if="isExternal"
+			@click="$emit('menuLinkClick', link)"
+			v-bind="$attrs"
+			:href="link"
+		>
+			<slot />
+		</a>
+		<router-link
```
FlorianWendelborn: Should be replaced with a slot to avoid third-party dependencies.

packages/kotti-style/_variables.scss
```diff
@@ -1,3 +1,9 @@
+// CSS variables
+html {
+	--navbar-text-active-color: #fff;
```
FlorianWendelborn: Not compatible with Internet Explorer: https://caniuse.com/#feat=css-variables

mrdionjr:
Hi, I hope this PR will be merged. Reloading the page gives a bad UX (loss of state, loss of SPA experience). Concerning the initial assumptions I don't think it's (still?) strong enough to not go in this direction. I expect most of the people using this kit to build web apps.

mrdionjr:
@tomquinonero Maybe you can check [this](https://github.com/chrisvfritz/vue-enterprise-boilerplate/blob/master/src/components/_base-link.vue) to get some ideas for the requested change.

amrdraz:
@tomquinonero so have you considered this?

FlorianWendelborn:
Closed as discussed on Slack.

PR #95, opened 2019-07-26
fix: fix comments behavior on edit

tomquinonero:
Fixed behavior in case of canceling comment edition. 
Also removed references to racist people.


docs/pages/patterns/comments.vue
```diff
@@ -185,7 +185,7 @@ export default {
 					id: 1,
 					userId: 12,
 					userName: 'Margaret Atwood',
-					message: `Radric is out, now free Charles`,
+					message: `We miss you, David`,
```
tomquinonero: 🤫🤫🤫

PR #96, opened 2019-08-01
Add dropdown button

tomquinonero:
This PR introduce a dropdown button feature.
It uses the actual dropdown implementation and uses a CSS enhanced button group.

Low  urgence on this one, we can take time.

### DIFF
--- a/docs/pages/components/dropdowns.vue
+++ b/docs/pages/components/dropdowns.vue
@@ -31,6 +31,16 @@ In `KtDropdownMenus` slot, you can write your own stylied element. However, we s
 </KtDropdown>
 ```
 
+## Dropdown button 
+
+<div class="element-example">
+	<KtDropdownButton size="sm" type="primary" v-t="'Dropdown button'" >
+		<div slot="KtDropdownMenus">
+			<KtDropdownMenu>Request item</KtDropdownMenu>
+			<KtDropdownMenu>Download Item</KtDropdownMenu>
+		</div>
+	</KtDropdownButton>
+</div>
 
 ## Icons and Size
 
--- a/docs/pages/components/dropdowns.vue
+++ b/docs/pages/components/dropdowns.vue
@@ -48,6 +48,17 @@ You can also use any attribute you would pass to `KtPopover` or `KtButton`. They
 	</KtDropdownButton>
 </div>
 
+``` html
+<div class="element-example">
+	<KtDropdownButton size="sm" type="primary" v-t="'Dropdown button'" >
+		<div slot="KtDropdownMenus">
+			<KtDropdownMenu>Request item</KtDropdownMenu>
+			<KtDropdownMenu>Download Item</KtDropdownMenu>
+		</div>
+	</KtDropdownButton>
+</div>
+```
+
 ## Icons and Size
 
 <div class="element-example">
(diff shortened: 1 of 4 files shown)


packages/kotti-dropdown-button/KtDropdownButton.vue
```diff
@@ -0,0 +1,90 @@
...
+				<KtButton
+					class="kt-dropdown-button__icon"
+					v-bind="$attrs"
+					:icon="dropdownIcon"
+				>
+					<i class="yoco" v-text="dropdownIcon" />
+				</KtButton>
+			</KtButtonGroup>
+			<div slot="content">
+				<slot name="KtDropdownMenus">Dropdown Menu</slot>
```
amrdraz: Why name the slot, why not just have it be the default slot

packages/kotti-dropdown-button/KtDropdownButton.vue
```diff
@@ -0,0 +1,90 @@
...
+import KtButton from '../kotti-button'
+import KtButtonGroup from '../kotti-button-group'
+import KtPopover from '../kotti-popover'
+
+export default {
+	name: 'KtDropdownButton',
+	props: {
+		dropdownIcon: {
+			type: String,
+			default: 'chevron_up',
```
amrdraz: I think most dropdown default to chevrin_down not up
also opening the drop down down
it's very strange to see the arrow pointing up
the only exception is if the dropdown is in the bottom of the ui at which point maybe like popover we could have the icon default to poointing at the side the popover would open, unless it was explicitly set

packages/kotti-dropdown-button/KtDropdownButton.vue
tomquinonero: I copied the behavior of the "classic" Dropdown. I don't think we should change it here if we don't change it in the `KtDoropdown` component as well.

packages/kotti-dropdown-button/KtDropdownButton.vue
tomquinonero: Yes I kind of did that thinking of bottom buttons. I don't mind defaulting down.

docs/pages/components/dropdowns.vue
```diff
@@ -31,6 +31,33 @@ In `KtDropdownMenus` slot, you can write your own stylied element. However, we s
 </KtDropdown>
 ```
 
+## Dropdown button 
+
+A dropdown button is a current pattern nowadays.
```
FlorianWendelborn: I think this sentence should be removed.

packages/kotti-dropdown-button/KtDropdownButton.vue
FlorianWendelborn: Slot name is very redundant. Just call it `menu` IMO.

docs/pages/components/dropdowns.vue
```diff
@@ -31,6 +31,33 @@ In `KtDropdownMenus` slot, you can write your own stylied element. However, we s
...
+			<KtDropdownMenu>Download Item</KtDropdownMenu>
+		</div>
+	</KtDropdownButton>
+</div>
+
+``` html
+<div class="element-example">
+	<KtDropdownButton size="sm" type="primary" v-t="'Dropdown button'" >
+		<div slot="KtDropdownMenus">
+			<KtDropdownMenu>Request item</KtDropdownMenu>
```
FlorianWendelborn: Looks more like a `KtDropdownItem` or `KtDropdownMenuItem` to me.

tomquinonero: That is using the existing naming for KtDropdown :/

PR #97, opened 2019-08-02
Use class for buttons

tomquinonero:
This PR helps our buttons being less intrusive. Before, the style of button was done directly on the `button` element. This was quite bad because it is a very low selector specificity. Also I don't like the fact that a `KtButton` has to be a `button`. 

Every button are now stylized by their `kt-btn` class. It also allows us to use the new `element` props that let us chose which html element the button should be.

One thing that could brake could be that we overwritten the `KtButton` style using `button` selector.
We'll have to change it ideally but it will not break anything.


packages/kotti-button-group/src/ButtonGroup.vue
```diff
@@ -20,3 +20,27 @@ export default {
 	},
 }
 </script>
+<style lang="scss">
```
mvangent: shouldn't this be scoped as well?

packages/kotti-button/src/button.vue
```diff
@@ -53,6 +56,7 @@ export default {
 			const classes = [this.type, this.objectClass]
 			if (this.size === 'small') classes.push('sm')
 			if (this.size === 'large') classes.push('lg')
+			classes.push('kt-btn')
```
mvangent: why not add 'kt-btn' to `const classes = [this.type, this.objectClasses, 'kt-btn']` ?

packages/kotti-button/src/button.vue
```diff
@@ -103,3 +107,111 @@ export default {
 	},
 }
 </script>
+<style lang="scss">
+$default-button-height: 32px;
+$large-button-height: 40px;
+$small-button-height: 24px;
```
mvangent: maybe good to semantically create a difference between these local variables and the globals?

mvangent:
@tomquinonero, done reviewing. nice refactor. have noted 2 questions.

packages/kotti-button/src/button.vue
tomquinonero: I'm not sure to understand what you mean there, do you have an example ?

packages/kotti-button/src/button.vue
tomquinonero: Sure we can do it :)

packages/kotti-button/src/button.vue
amrdraz: these variables should probably be derived from existing $unit measurements already used instead of being hard code declared

packages/kotti-button/src/button.vue
amrdraz: but you still didn't

packages/kotti-button-group/src/ButtonGroup.vue
```diff
@@ -20,3 +20,27 @@ export default {
...
+.button-group {
+	/** Fix for inline element space
+	https://css-tricks.com/fighting-the-space-between-inline-block-elements/ **/
+	font-size: 0;
+
+	.kt-btn {
```
FlorianWendelborn: Should be `.kt-button` so that no second-guessing “was this class shortened or not” needs to happen.

docs/pages/components/buttons.vue
```diff
@@ -118,6 +118,22 @@
...
+
+	<div class="element-example white">
+		<KtButton type="primary" element="div">Primary Button</KtButton>
+	</div>
+
+	``` html
+	<KtButton type="primary" element="div">Primary Button</KtButton>
+	```
+
+	This can be useful
```
FlorianWendelborn: Should be removed IMO.

packages/kotti-button/src/button.vue
```diff
@@ -1,15 +1,17 @@
 <template>
-	<button
+	<component
+		:is="element"
 		:class="mainClasses"
 		:style="themeColor"
 		@click="handleClick"
 		@mouseover="isHover = true"
 		@mouseleave="isHover = false"
+		role="button"
```
FlorianWendelborn: Nice

docs/pages/components/buttons.vue
tomquinonero: The `button` has issues. For example it cannot be used as a link.

FlorianWendelborn: I’m talking about the line that I commented on. Otherwise, please add "This can be useful" to every paragraph in the docs.

PR #98, opened 2019-08-05
add disabled prop to KtSelect

mvangent:
The title speaks for itself :)

Nevertheless, A small addition addition of functionality to pass make available and pass on a `disabled` prop to the input tag of KtSelect

### DIFF
--- a/docs/pages/components/selects.vue
+++ b/docs/pages/components/selects.vue
@@ -35,6 +35,19 @@
 	<button @click="value2=null">Reset</button>
 </div>
 
+## Disable
+
+<div class="element-example">
+	<KtSelect label="Words"
+		placeholder="Select Word"
+		disabled="true"
+		v-model="value2"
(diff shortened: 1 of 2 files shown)


docs/pages/components/selects.vue
```diff
@@ -35,6 +35,19 @@
...
+	<KtSelect label="Words"
+		placeholder="Select Word"
+		disabled="true"
+		v-model="value2"
+		allowEmpty
+		:options="words"/>
+	<h5 v-text="`English: ${value2}`" class="mt-16px"/>
+	<button @click="value2=null">Reset</button>
+</div>
+
```
tomquinonero: Can you also add the code example ?

mvangent: Line 43.

mvangent: done

PR #99, opened 2019-08-13
updated breadcrumbs to support query and url params

Goldziher:
I added support for url params and query params. I needed this for the checkout workflow - we are using param based navigation. Please expediate this PR.


tomquinonero:
We actually already support. The `to` object will be given to the router as any `to` object and can contain name, query and params.

PR #100, opened 2019-08-14
revert the clickaway behaviour

Daniel-Beckstein:
revert the clickaway behaviour as it was causing issues with closing popovers


amrdraz:
@tomquinonero you had something to do with this can you comment on why this behavior was added in the first place?

mvangent:
@amrdraz The previous solution helped to avoid autoclosing the Popover on any KtSelect click events. But since the filters are not using popover anymore, we can safely revert back to old implementation.

PR #103, opened 2019-08-29
chore(fix number input not working in IE11)

Goldziher:
The number input component did not function correctly in ie11. The bug was the use of stepUp - an html input method that does not exist in IE11.

### DIFF
--- a/packages/kotti-input-number/src/InputNumber.vue
+++ b/packages/kotti-input-number/src/InputNumber.vue
@@ -1,21 +1,30 @@
 <template>
 	<div :class="formGroupStyle">
-		<div :class="decreaseButtonStyle" @click="decreaseNumber">
+		<div
+			:class="decreaseButtonStyle"
+			@click="setCurrentValue(currentValue + 1)"
+		>
 			<i class="yoco">minus</i>
 		</div>
 		<input
+			ref="inputNumber"
 			:class="inputStyle"
 			type="number"
 			:disabled="disabled"
-			:step="step"
 			:value="currentValue"
 			:max="max"
 			:min="min"
-			ref="inputNumber"
-			@change="handleInput"
+			@input="setCurrentValue($event.target.value)"
 		/>
-		<div v-if="max && showMaxNumber" class="kt-input-number__max" v-text="max"/>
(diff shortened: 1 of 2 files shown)


packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -1,21 +1,30 @@
 <template>
 	<div :class="formGroupStyle">
-		<div :class="decreaseButtonStyle" @click="decreaseNumber">
+		<div
+			:class="decreaseButtonStyle"
+			@click="setCurrentValue(currentValue + 1)"
```
tomquinonero: Small mistake here:

tomquinonero:
```suggestion
			@click="setCurrentValue(currentValue - step)"
```

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -1,21 +1,30 @@
...
+		>
 			<i class="yoco">minus</i>
 		</div>
 		<input
+			ref="inputNumber"
 			:class="inputStyle"
 			type="number"
 			:disabled="disabled"
-			:step="step"
```
tomquinonero: This should be keeped!

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -1,21 +1,30 @@
...
+		<div
+			v-if="max && showMaxNumber"
+			class="kt-input-number__max"
+			v-text="max"
 		/>
-		<div v-if="max && showMaxNumber" class="kt-input-number__max" v-text="max"/>
-		<div :class="increaseButtonStyle" @click="increaseNumber">
+		<div
+			:class="increaseButtonStyle"
+			@click="setCurrentValue(currentValue + 1)"
```
tomquinonero: Use step here as well:

tomquinonero:
```suggestion
			@click="setCurrentValue(currentValue + step)"
```

Goldziher:
tested and verified on both latest chrome and IE11

PR #104, opened 2019-09-10
Fix bad refs in select

amrdraz:
KtSelect renders it's options outside in the body
But the way this was implemented was a bit buggy and only accepted because it wasn't really causing a problem, until recently

It seems the append method used did not account for IE not having append but instead appendChild as well as other issues

I took the opportunity to generally refactor our select component to use the a SimplePortal component 

This portal operates similar to how React.createProtal works
see https://github.com/LinusBorg/vue-simple-portal

Unfortunately the library did not work in IE as well (also had some minor issues with server side rendering)

and since there's an open high priority bug and we will probably use portals across out app for modals and other overlay components I figured we might as well copy this small generic solution into our project and maintain it ourselves (at least until my PR to the original lands and a new version is published)

side note the clickaway behavior was already not working (though it now works in IE)

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,5 +112,5 @@
 		"prepublishOnly": "yarn run build:kotti"
 	},
 	"style": "dist/KottiUI.css",
-	"version": "1.7.5"
+	"version": "1.7.6"
 }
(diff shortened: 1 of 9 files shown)


amrdraz:
ok I figured out why the select box doesn't close sometimes
It's the boolean flag that requires a value is selected  if allowEmpty isn't enabled
honestly it's a bit unexpected behaviour UX wise I kept it when I was updating the clickaway but and it kinda makes sense but I'm not entirely for it

PR #106, opened 2019-09-27
fix(B3-6306: fix-table-show-hide)

mvangent:
@Juli0GT and I are not saying this is the most elegant solution, but it is urgent, and it seems to be working based on our tests.


packages/kotti-table/src/logic/column.js
```diff
@@ -90,7 +105,11 @@ export function setColumn(state, { column, index, deleted }) {
 		if (deleted) {
```
mes3yd: (2) This may be right, but you set `_deleted` to false on both if and else, which seems in conflict the above statement.

packages/kotti-table/src/logic/column.js
```diff
@@ -17,25 +17,40 @@ export const defaultState = {
...
 		const pickedColumns = columns.map(col => pick(col, PUBLIC_COLUMN_PROPS))
-		for (const col of pickedColumns) {
+		for (let i = 0; i < pickedColumns.length; i++) {
+			const col = pickedColumns[i]
 			const column = getColumn(state, col)
-			if (column) {
+
+			if (column && state._columns.length > 0) {
 				Object.assign(column, col)
+			} else {
```
mes3yd: (2) It looks unintentional that the else clause fires on column resolving to falsy

mvangent: it is intentional. If the column does not exist in the old state, it will set a new column from pickedColumns

packages/kotti-table/src/logic/column.js
mvangent: ok. nice one. I will set it in the setColumn call.

mvangent: actually, what is wrong with it?

packages/kotti-table/src/logic/column.js
mes3yd: Discussed in person: the extra restriction is to prevent corrupted state to modify the table. This is correct.

packages/kotti-table/src/logic/column.js
mes3yd: Discussed in person: It is trying to correct an initialization error happening elsewhere, but the culprit cannot be found at this moment and this needs to be revisited.

packages/kotti-table/src/TableHeader.vue
```diff
@@ -80,7 +80,14 @@ export default {
...
 		tableColumns() {
-			return this[KT_STORE].state.columns
+			return this[KT_STORE].state.columns.map(column => ({
+				...column,
+				renderHeader: column.renderHeader
+					? column.renderHeader
+					: function DEFAULT_RENDER_HEADER(h, { column }) {
+							return [column.label === undefined ? column.prop : column.label]
+					  },
+			}))
```
amrdraz: this should be covered in the tableColumn creation function where we define renderHeader
why was it moved up here

packages/kotti-table/src/logic/column.js
```diff
@@ -17,25 +17,40 @@ export const defaultState = {
...
 	setColumns(store, columns = store.state._columnsArray) {
 		const { state } = store
 		const pickedColumns = columns.map(col => pick(col, PUBLIC_COLUMN_PROPS))
-		for (const col of pickedColumns) {
+		for (let i = 0; i < pickedColumns.length; i++) {
+			const col = pickedColumns[i]
 			const column = getColumn(state, col)
-			if (column) {
+
+			if (column && state._columns.length > 0) {
```
amrdraz: _columns is an object dictionary it has no length
_columnsArray is the array form of _columns

also there's no need to do that check as getColumn will return undefined if a column is not in _columns

and there for you can just return the condition to `if (colum) { ... } else { ... }`

PR #107, opened 2019-10-02
Closable block

tomquinonero:
I have this component in part of the sales transactions and I want to make it part of kotti.

Pretty simple component except maybe the opening animation that uses CSS vars. It works with IE11, it just doesn't animate there.

I used also an interval that compute the height every 500ms. It is because the content was changing in my case (thumbnails were loading). It will also handle cases like resize in a mildly elegant way.

I got a concern about the naming, I don't really like that it's call `ClosableBlock` but could come up with something better and didn't find any generic name for that UI pattern. Thoughts?

As this is a simple PR I will merge with 2 approvals, but please suggest a better naming even if you cannot review the code itself!


ostoc:
There is one good naming system is from IBM: https://www.carbondesignsystem.com/components/accordion/code 

Screen Shot 2019-10-02 at 14 55 25

the bottom line is doubled.

tomquinonero:
> There is one good naming system is from IBM: https://www.carbondesignsystem.com/components/accordion/code

Thanks a lot! I think we'll use that!

> Screen Shot 2019-10-02 at 14 55 25
> 
> the bottom line is doubled.

Soooo this wasn't on the original mockup, BUT it looks cool to me, give that subtle shadow and I find it actually better looking with than without. Thoughts?

tomquinonero:
Just renamed it to "accordion", check it out there: https://5d9b132f46e55100081d8615--kotti.netlify.com/components/accordion/

amrdraz:
I feel the icon should be smaller compared to then the text

Screenshot 2019-11-13 at 11 48 22

amrdraz:
@tomquinonero  You didn't add the accordion link to the components navigation, I had to navigate to the page via url

amrdraz:
Last comment can we have an accordion group? for toggling the opening of accordions

tomquinonero:
> @tomquinonero You didn't add the accordion link to the components navigation, I had to navigate to the page via url

Didn't manage to show it for some reason :/ can you check ?



> I feel the icon should be smaller compared to then the text, or the padding should be smaller
> 
> Screenshot 2019-11-13 at 11 48 22

Smaller now



> Last comment can we have an accordion group? for toggling the opening of accordions

That'll be cool, but might be part of another PR

PR #110, opened 2019-10-24
Improve dropdown button

tomquinonero:
Two improvements for `KtDropdownButton`:

- Handles `disabled` attribute
- Handles `size` attribute (passed to KtPopover`)

Also added an example to the documentation for the disabled state.

### DIFF
--- a/packages/kotti-dropdown-button/KtDropdownButton.vue
+++ b/packages/kotti-dropdown-button/KtDropdownButton.vue
@@ -1,6 +1,6 @@
 <template>
 	<div class="kt-dropdown-button">
-		<KtPopover v-bind="$attrs" :placement="placement">
+		<KtPopover :size="size" v-bind="$attrs" :placement="placement">
 			<KtButtonGroup>
 				<KtButton
 					:disabled="disabled"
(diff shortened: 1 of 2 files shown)


amrdraz:
@tomquinonero  why does the disabled dropdown have a different color?

Screenshot 2019-11-13 at 11 58 16

amrdraz:
@tomquinonero I approved but I saw no change regarding the different coloured dropdown

packages/kotti-popover/src/Popover.vue
```diff
@@ -49,6 +50,7 @@ export default {
 	},
 	methods: {
 		handleClick() {
+			if (this.disabled) return null
```
FlorianWendelborn:
```suggestion
			if (this.disabled) return
```

Alternatively, use `if (!this.disabled) this.showPopper = !this.showPopper`.

packages/kotti-popover/src/Popover.vue
```diff
@@ -16,6 +16,7 @@ export default {
 		placement: { type: String, default: 'bottom' },
 		size: { type: String, default: null },
 		content: { type: String, default: '' },
+		disabled: { type: Boolean, default: false },
```
FlorianWendelborn: Bogo sort isn’t a good sorting algorithm with 1 iteration.

packages/kotti-dropdown-button/KtDropdownButton.vue
```diff
@@ -39,6 +49,10 @@ export default {
 			type: String,
 			default: null,
 		},
+		disabled: {
```
FlorianWendelborn: Same as below.

FlorianWendelborn:
@tomquinonero @carsoli Can this be merged?

PR #112, opened 2019-11-04
Feature/add required prop to kotti inputs

carsoli:
[ KtRadioGroup - KtSelect - KtCheckbox - KtDateInput ] are all missing 1) a required prop that invalidates the form if the field is missing, and 2) corresponding visual indication. 
The former two now have an indicator if the field is required but won't raise an invalid flag, therefore, something needs to be done so that form-validators can check if these fields are invalid (open for discussion)
The latter have an invalid flag.
Note: the visual indicator now is a red border/red icons that are red upon render, however if we only need to make them red when the fields are dirty - this can be easily done.

### DIFF
--- a/docs/pages/components/selects.vue
+++ b/docs/pages/components/selects.vue
@@ -113,6 +113,29 @@
 	<h5 v-text="`English: ${value3}`" class="mt-16px"/>
 </div>
 
+## Required 
+
+<div class="element-example">
+	<KtSelect label="Words"
+		placeholder="Select Word"
+		v-model="value5"
+		required
+		icon="user"
+		:options="words"/>
+	<h5 v-text="`English: ${value5}`" class="mt-16px"/>
+</div>
+
+if you allowEmpty, you can't make the field required. allowEmpty will make field valid in that case
+
+```html
+	<KtSelect label="Words"
+		placeholder="Select Word"
+		v-model="value5"
+		required
+		icon="user"
+		:options="words"/>
+```
+
 ## Async Search
 
 <div class="element-example">
@@ -168,14 +191,15 @@ export default {
 	data() {
 		return {
 			value1: null,
-			value2: 'Avocados',
+			value2: null,
 			value3: 'Avocados',
 			value4: null,
+			value5: null,
 			words: [
-				{
-					label: 'Empty',
-					value: null,
-				},
+				// {
+				// 	label: 'Empty',
+				// 	value: null,
+				// },
 				{
 					label: '西瓜',
 					value: 'Watermelon',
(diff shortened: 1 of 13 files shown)


packages/kotti-checkbox/src/Checkbox.vue
```diff
@@ -17,11 +18,16 @@ export default {
...
 	props: {
 		label: { default: null, type: String },
+		required: { default: false, type: Boolean },
 		value: { default: false, type: Boolean },
 	},
 	computed: {
-	        hasLabel() {
-			return !(this.label===null || this.label === undefined)
+		hasLabel() {
+			return !(this.label === null || this.label === undefined)
```
Goldziher: ```return Boolean(this.label)```, otherwise if you want to be more precise - ```return this.label && this.label.length```

packages/kotti-datepicker/src/DateInput.vue
```diff
@@ -54,6 +56,13 @@ export default {
 			if (this.currentDate) return this.currentDate.toLocaleDateString()
 			return null
 		},
+		hasLabel() {
+			return !(this.label === null || this.label === undefined)
```
Goldziher: same as prev

packages/kotti-datepicker/src/DateInput.vue
```diff
@@ -54,6 +56,13 @@ export default {
...
+		},
+		labelRep() {
+			if (!this.hasLabel) return
```
Goldziher: you should return an empty string instead of returning undefined. Otherwise, add a v-if condition to the span that check ```hadLabel``` instead of this.

packages/kotti-radio-group/src/RadioGroup.vue
```diff
@@ -26,20 +25,48 @@ export default {
...
 				'form-group--list-options': this.listStyle,
+				'form-group--invalid': this.invalidInput,
 			}
 		},
+		hasLabel() {
+			return !(this.label === null || this.label === undefined)
+		},
+		labelRep() {
+			if (!this.hasLabel) return
+			return this.required ? `${this.label} *` : this.label
```
Goldziher: same as prev. comments

docs/pages/components/selects.vue
```diff
@@ -168,14 +191,15 @@ export default {
...
+			value2: null,
 			value3: 'Avocados',
 			value4: null,
+			value5: null,
 			words: [
-				{
-					label: 'Empty',
-					value: null,
-				},
+				// {
```
mvangent: very minor. but these commented out lines can be deleted I assume?

PR #113, opened 2019-11-12
fix(required prop works natively now and not just using css)

carsoli:
Radio and Select didn't pass the required prop right away but rather worked around it
radiogroup was a wrapper with no actual input html on the parent component >
(now it has a hidden input that becomes invalid and uses the native required attribute)
singleselect had a readonly attribute so it couldn't be invalidated when it was readonly >
now it mimics the readonly attribute but can be invalidated

-fixed singleselect to show all options (not just filtered) when the input field is out of focus then selected again
-fixed a null invocation error

issue#111

### DIFF
--- a/docs/pages/components/selects.vue
+++ b/docs/pages/components/selects.vue
@@ -85,6 +85,10 @@
 			value: 'Watermelon',
 		},
 		{
+			label: '西蕉',
+			value: 'Waternana',
+		},
+		{
 			label: '香蕉',
 			value: 'Banana',
 		},
@@ -203,6 +207,10 @@ export default {
 				{
 					label: '西瓜',
 					value: 'Watermelon',
+				},
+				{
+					label: '西蕉',
+					value: 'Waternana',
 				},
 				{
 					label: '香蕉',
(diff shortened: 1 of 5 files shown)


docs/pages/components/selects.vue
```diff
@@ -204,6 +208,10 @@ export default {
 					label: '西瓜',
 					value: 'Watermelon',
 				},
+				{
+					label: '西蕉',
+					value: 'Waternana',
+				},
```
amrdraz: we're keeping this example?

carsoli: just as an example of how a filter on 西 can fit multiple options :-)

packages/kotti-single-select/src/SingleSelect.vue
```diff
@@ -236,15 +232,21 @@ export default {
 			if (!this.isOptionAllowed(option)) return
 
 			this.selectedLabel = option.label
-
-			const selectedItem = this.options.find(
-				({ label }) => label === this.selectedLabel,
-			)
+			// const selectedItem = this.options.find(
+			// 	({ label }) => label === this.selectedLabel,
+			// )
```
tomquinonero: Is that comment something you want to keep in the future?

carsoli: well it's now done by directly assigning the option to the selectedItem.. i checked that they are always the same object, but i wasn't sure if there was maybe a use-case in particular that would require it to be this way - so i left it just in case

PR #116, opened 2019-11-29
Fonts quick update

tomquinonero:
A quick PR that changes how we use fonts.
We needed to clarify which fonts are used and we decided on Roboto (for now :stuck_out_tongue: )

I simplified how fonts are called and updated the documentation regarding it. 

NB: This is a breaking change (as you need to implements Roboto on your end), this is why I require the merge to `kotti-2.0` branch.

### DIFF
--- a/docs/assets/css/main.scss
+++ b/docs/assets/css/main.scss
@@ -2,7 +2,7 @@
 
 body {
 	margin: 0;
-	font-family: Roboto, sans-serif;
+	font-family: Roboto, Arial, sans-serif;
 	font-size: 15px;
 }
 
--- a/docs/assets/css/main.scss
+++ b/docs/assets/css/main.scss
@@ -2,8 +2,7 @@
 
 body {
 	margin: 0;
-	font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen,
-		Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
+	font-family: Roboto, sans-serif;
 	font-size: 15px;
 }
 
--- a/docs/pages/foundations/typography.vue
+++ b/docs/pages/foundations/typography.vue
@@ -3,9 +3,10 @@
 
 ## Typeface
 
-Kotti uses Open Sans as default font for all text.
-
-For macOS, we use Apple system font San Francisco, you can download from [Fonts - Apple Developer](https://developer.apple.com/fonts/)
+Kotti uses Roboto as default font for all text.
+You are responsible for loading it into your app. For this documentation we use the
+google font version online, but you might want to load it locally if your app doesn't
+have access to the network.
 
 ## Headers
 
@@ -143,5 +144,7 @@ For macOS, we use Apple system font San Francisco, you can download from [Fonts 
 </template>
 
 <script>
-export default { name: 'Typography' }
+export default {
+	name: 'Typography',
+}
 </script>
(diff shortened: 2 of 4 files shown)


docs/assets/css/main.scss
```diff
@@ -1,6 +1,5 @@
 body {
-	font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen,
-		Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
+	font-family: Roboto, sans-serif;
```
tomquinonero: This is just for the documentation of kotti!

docs/pages/foundations/typography.vue
```diff
@@ -133,15 +134,15 @@ For macOS, we use Apple system font San Francisco, you can download from [Fonts
...
-|:--------------------------------------------------------------------------------------------------------------------------|:-----------------------------------------------------------|
-| *Italics text*                                                                                                            | Using `<em>`                                               |
-| **This is bold text**                                                                                                     | Using `<strong>`                                           |
-| <ol><li>This is an ordered list;</li><li>Second item of ordered list;</li><li>Third item of ordered list.</li></ol>       | `<li>` element has margin `8px` with default `line-height` |
-| <ul><li>This is an unordered list;</li><li>Second item of unordered list;</li><li>Third item of unordered list.</li></ul> | same as ordered list                                       |
+| Style                                                                                                                     | Notes                                                    |
```
tomquinonero: This was because smart ass ESlint consider `<li>` as an opened tag so I had to change it to `li`.

nuxt.config.js
```diff
@@ -15,7 +15,13 @@ const config = {
 design decisions. The Kotti design system has two main parts: foundation and components.`,
 			},
 		],
-		link: [{ rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' }],
+		link: [
+			{ rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' },
+			{
+				href: 'https://fonts.googleapis.com/css?family=Roboto&display=swap',
```
tomquinonero: Call it from google for the doc. Needs to be implemented into yoda after.

docs/assets/css/main.scss
amrdraz: I don't approve of this, if we're going to implement a font then we either require it or we allow fallbacks

the way to solve font in consistency is to include our own font not to compound the problem worse by making none linux platforms render the wrong thing

Finally Ruboto? really I get Open Sans, might as well pick Ariel

tomquinonero: > Also a very similar font to Roboto that's on every platform is Arial so you can add that in the fallback list

Solved !

tomquinonero:
> Still not a fan of Roboto.

This was the UI designer choice. At least it can be displayed consistency and is quite similar to SF pro actually. We can change that in the future! We have now a cleaner way of setting fonts at least.

mvangent:
@tomquinonero is this ready for merging as far as you are concerned?

FlorianWendelborn:
@tomquinonero is this still relevant? Should this be merged?

tomquinonero:
> @tomquinonero is this still relevant? Should this be merged?

yes please :)

FlorianWendelborn:
I think the new fonts decrease readability and contrast. They also seem more blurry to me. The major differences are in the "GUIDES" headline and in the blurriness of the "Avatars" menu entry.

## New Fonts

image

## Old Fonts

image

PR #118, opened 2019-11-29
Fix number inputs firefox

tomquinonero:
The big thing is just the style being moved.
The key point here is: ` 	-moz-appearance: textfield; ` 

Voila.

One review necessary


FlorianWendelborn:
What does this "fix"?

PR #125, opened 2020-01-15
124 fix kt table drag and drop reorder behavior

carsoli:


docs/pages/components/tables.vue
```diff
@@ -404,33 +420,33 @@ You can then use the set sortedColumns prop on the KtTable to update the table u
 </KtTable>
 
 ```html
-<KtTable :rows="rows" :columns="columns">
```
FlorianWendelborn: Does this example really make sense without `columns`?

packages/kotti-banner/src/Banner.vue
```diff
@@ -65,7 +65,7 @@ export default {
 			}
 		},
 		expandable() {
-			return this.$slots.expand
+			return !!this.$slots.expand
```
FlorianWendelborn:
```suggestion
			return Boolean(this.$slots.expand)
```

packages/kotti-table/src/Table.vue
```diff
@@ -70,10 +70,10 @@ export default {
 
 		loading: Boolean,
 
-		tdClasses: { default: () => [], types: [Array, String, Object] },
```
FlorianWendelborn: Good catch, I fixed the rest of these on `master` now.

packages/kotti-table/src/TableColumn.js
```diff
@@ -47,6 +47,7 @@ const TableColumn = {
 		renderHeader: Function,
 		renderCell: Function,
 		renderContext: Object,
+		isPropDefined: { default: false, type: Boolean },
```
FlorianWendelborn: random order is best order 😉

packages/kotti-banner/src/Banner.vue
mvangent: agreed. @carsoli can you commit this suggestion?

mvangent:
@carsoli code is looking good to me. can you add a test instruction to the description?

packages/kotti-banner/src/Banner.vue
carsoli: i don't even know why this commit is here, it should be on #126 :D but i was going to add it to 126...

packages/kotti-table/src/TableColumn.js
carsoli: _FEATURE_

docs/pages/components/tables.vue
carsoli: no, u r right

carsoli:
> @carsoli code is looking good to me. can you add a test instruction to the description?
If you follow the issue description, and try to see the expected behavior, that should be it. In the docs it's just `Ordering Columns` subheader that has draggable columns and you should be able to drag as many times as possible without the rendering breaking. (so, in the docs, the formatting of the date should always or the rendering of the avatar as well)

mvangent:
> @carsoli code is looking good to me. can you add a test instruction to the description?
> If you follow the issue description, and try to see the expected behavior, that should be it. In the docs it's just `Ordering Columns` subheader that has draggable columns and you should be able to drag as many times as possible without the rendering breaking. (so, in the docs, the formatting of the date should always work and the rendering of the avatar as well)

hi @carsoli I ran the test. I can verify dragging and dropping is working smoothly. no console.errors. minor: couldn't manage to drag a column to the last row, however from the last row also works.

carsoli:
> > @carsoli code is looking good to me. can you add a test instruction to the description?
> > If you follow the issue description, and try to see the expected behavior, that should be it. In the docs it's just `Ordering Columns` subheader that has draggable columns and you should be able to drag as many times as possible without the rendering breaking. (so, in the docs, the formatting of the date should always work and the rendering of the avatar as well)
> 
> hi @carsoli I ran the test. I can verify dragging and dropping is working smoothly. no console.errors. minor: couldn't manage to drag a column to the last row, however from the last row also works.

I know about the last column thing. it's because of the way the dragging is implemented so in theory it is not a bug, it really is a feature. if you want me to look into allowing for it, i can create a separate feature issue. and yes, the only way to get around this is to drag whatever column to the before-last column THEN drag the last column to the left!

mvangent:
> > > @carsoli code is looking good to me. can you add a test instruction to the description?
> > > If you follow the issue description, and try to see the expected behavior, that should be it. In the docs it's just `Ordering Columns` subheader that has draggable columns and you should be able to drag as many times as possible without the rendering breaking. (so, in the docs, the formatting of the date should always work and the rendering of the avatar as well)
> > 
> > 
> > hi @carsoli I ran the test. I can verify dragging and dropping is working smoothly. no console.errors. minor: couldn't manage to drag a column to the last row, however from the last row also works.
> 
> @mvangent I know about the last column thing. it's because of the way the dragging is implemented so in theory it is not a bug, it really is a feature. if you want me to look into allowing for it, i can create a separate feature issue. and yes, the only way to get around this is to drag whatever column to the before-last column THEN drag the last column to the left!

@carsoli I think its fine to leave it, it was more a heads up, and if you say it is a feature, well... features nowadays..  I approved.

Issue #126, opened 2020-01-16
KtBanner shows empty action button when it has no expand slots


carsoli:
**Describe the bug**
KtBanner that doesn't have an expand slot shows an empty button by default

**To Reproduce**
1. Go to KtTable docs (tables.vue)
2. Go to Expandable subheader
3. You'll see an empty button 

**Expected behavior**
no button shows at all

Issue #124, opened 2020-01-15
fix-ktTable-drag-and-drop-reorder-behavior


carsoli:
**Description**
A) reordering the columns doesn't render the columns correctly after the first drag event (works only with the *first* drag&drop event)

B) dragging a column and dropping it to the same position makes it move to the before-last column  

**Steps To Reproduce**
1. Go to a draggable kotti table (`useColumnDragToOrder` flag is set to true)
A) Drag a column around more than once.
B) Drag a column to its same position or maybe n pixels to the right where n << the width the column occupies

**Expected behavior**
A) table shows columns in the new order with every cell rendered correctly
B)  column doesn't change order at all and table does not re-render

**Actual behavior**
1) columns disappear 
2) columns are not rendered correctly (e.g.: translation breaks, any function that's bound to "formatter"/"renderCell")

**Desktop (please complete the following information):**
 - OS: makes no difference (reproducible on a linux/macOS)
 - Browser: reproducible on Chrome, Chromium, Safari, Firefox
 - Version (multiple)

**Smartphone (please complete the following information):**
 - did not try 
- but generally drag and/or drop event is not detected on iOS Safari (v. 13.3)

PR #127, opened 2020-01-16
fix(issue#126): removed empty default actionButton in KtBanner

carsoli:
made `expandable` return boolean when no slot="expand" exists
if no actionText is provided and since it defaults to empty string, button doesn't show at all (v-if)
#126

### DIFF
--- a/packages/kotti-banner/src/Banner.vue
+++ b/packages/kotti-banner/src/Banner.vue
@@ -65,7 +65,7 @@ export default {
 			}
 		},
 		expandable() {
-			return !!this.$slots.expand
+			return Boolean(this.$slots.expand)
 		},
 	},
 	methods: {
--- a/packages/kotti-banner/src/Banner.vue
+++ b/packages/kotti-banner/src/Banner.vue
@@ -13,9 +13,9 @@
 			</div>
 			<div class="message" v-text="message" />
 			<div class="action" v-if="!expandable" @click="handleClick">
-				<button class="text" v-text="actionText" />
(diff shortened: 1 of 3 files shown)


packages/kotti-banner/src/Banner.vue
```diff
@@ -65,7 +65,7 @@ export default {
 			}
 		},
 		expandable() {
-			return this.$slots.expand
+			return !!this.$slots.expand
```
AZeghers: henlo pls Boolean(thing) ty

packages/kotti-banner/src/Banner.vue
```diff
@@ -13,9 +13,9 @@
 			</div>
 			<div class="message" v-text="message" />
 			<div class="action" v-if="!expandable" @click="handleClick">
-				<button class="text" v-text="actionText" />
+				<button class="text" v-if="actionText.length" v-text="actionText" />
 			</div>
-			<div class="action" v-else @click="isExpand=!isExpand">
+			<div class="action" v-else @click="isExpand = !isExpand">
```
AZeghers: outstanding linting
I wholeheartedly approve this change
Thanks for the catch

packages/kotti-banner/src/Banner.vue
FlorianWendelborn: Why is this code also in #125?

carsoli: #124 u mean. idk what is #125 icet
and yes i just told u i reset head and made a separate issue (126)

will commit this here now.

Issue #128, opened 2020-01-21
fix-selected-property-on-ktTable/fix-multipleExpand-property-on-KtTable


carsoli:
**Describe the bug**
A. `selected` is the prop that replaces the deprecated v-model usage to bind selected rows on KtTable
passing a selected array of rows doesn't bind correctly b/c of strict comparison (check the additional context section)

B. `expandMultiple` allows for expansion of multiple rows at the same point in time. It didn't work. So a table with expandMultiple flag set was the same as one with the flag = false. 

**To Reproduce**
A]
1) bind selected to an array of an object defined in the column.
2) it is not bound correctly; the row that matches is not selected

B]
1) pass the expandMultiple to <KtTable /> 
2) try expanding more than one row at one point (only one row is allowed to be expanded) 

**Expected behavior**
A] the row to be correctly selected
B] multiple rows expand

**Additional context**
it all happened b/c we use array.prototype.indexOf() in checks, which internally does strict comparison of the objects within the array.
once the state.selected has been mapped to a JSON string, non-strict comparison (==) works fine
couldn't find a function that does non-strict comparison of indexOf so i mapped on each call for a check and now it works. 
Functional testing for it can be done after merging #124 in this PR since i'm updating the docs to support proper linting for the table page and adding a thorough example for selected

Same problem happens with expandMultiple. fixed same way. and the flag wasn't even initialized on the default state, so I passed this. 
few lines of changes :) 

UPDATE:: introduced the usage of deep-eql thanks to @FlorianWendelborn's eye. JSON.stringify doesn't behave as expected but the bug might have not presented itself in the near future :) 
It's non deterministic in the sense that Object.keys are not arranged the same way so the comparison fails for Object that have the same ({k,v}) pairs but sorted differently.

PR #129, opened 2020-01-21
fix(KtTable-selected-expandMultiple): fixed selected binding

carsoli:
#128

### DIFF
--- a/packages/kotti-table/src/logic/select.js
+++ b/packages/kotti-table/src/logic/select.js
@@ -26,7 +26,8 @@ export const mutations = {
 		if (rows.length === 0 || isAllRowsDisabled) return
 
 		const oldSelection = state.selection
-		const shouldSelectAll = !(state.isAllSelected || oldSelection.length)
+		const shouldSelectAll = !state.isAllSelected //more intuitive
+		// const shouldSelectAll = !(state.isAllSelected || oldSelection.length)
 
 		state.selection = shouldSelectAll ? [...enabledRows] : []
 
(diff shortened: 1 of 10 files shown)


packages/kotti-table/src/logic/select.js
```diff
@@ -61,14 +61,22 @@ export const getters = {
 			: row[state.rowKey]
 	},
 	isSelected(state, row) {
-		return (state.selection || []).indexOf(row) > -1
+		return (
+			(state.selection.map(e => JSON.stringify(e)) || []).indexOf(
```
FlorianWendelborn: https://www.npmjs.com/package/deep-eql or https://www.npmjs.com/package/shallow-equal

amrdraz:
@carsoli testing on netlify the binding preview stop working after I select all, which doesn't happen with the v-model mode 
also there's a merge conflict

carsoli:
> @carsoli testing on netlify the binding preview stop working after I select all, which doesn't happen with the v-model mode
Great Catch! Will see to it in a separate to commit.

> also there's a merge conflict
Resolved, now. 
@amrdraz

packages/kotti-table/src/logic/select.js
amrdraz: @carsoli did you address this?

@FlorianWendelborn would deep equal windup interacting with Vue's proxies?
I think stringily does the job of deep equal as far as the kind of props used in the columns are concerned as far as we need it too, if we're talking performance we may be ok with shallow equal

PR #135, opened 2020-02-03
feat: Add showPopperProp in KtPopover so it can be open from parent

Juli0GT:


packages/kotti-popover/src/Popover.vue
```diff
@@ -20,6 +20,16 @@ export default {
 		placement: { type: String, default: 'bottom' },
 		size: { type: String, default: null },
 		content: { type: String, default: '' },
+		showPopperProp: {
```
mvangent: Hi @Juli0GT, great addition and nice work. Regarding the naming, I would consider to rename the prop to `forceShowPopover`. since force is a bit clearer that it indicates an override of normal behaviour and popover is more descriptive than the library name popper.

Juli0GT: Yeah good point @mvangent I'll rename the prop.

PR #139, opened 2020-02-05
feat(popover-disabled-hover): Add forceDisable prop to KtPopover

Juli0GT:


packages/kotti-popover/src/Popover.vue
```diff
@@ -38,22 +43,25 @@ export default {
...
+			if (!this.forceDisable && !!this.showPopper) {
 				this.$nextTick(() => {
 					this.initPopper()
 				})
 			}
 		},
-		forceShowPopover(val, oldVal) {
-			if (val !== oldval && val !== null) {
+		forceShowPopover(val) {
+			console.log(val)
```
mvangent: console.log can be removed

PR #142, opened 2020-02-12
resolved min and max not being factored in factored in

Goldziher:
This PR fixes some buggy behaviour in the kt-number-input component. The bug was that "min" was not being factored in - for either the css or for the purposes of disallowing decrementing below the min threshold.

### DIFF
--- a/packages/kotti-input-number/src/InputNumber.vue
+++ b/packages/kotti-input-number/src/InputNumber.vue
@@ -1,17 +1,16 @@
 <template>
 	<div :class="formGroupStyle">
 		<div :class="decreaseButtonStyle" @click="decrementValue">
-			<i class="yoco">minus</i>
+			<i :class="yocoClassDecrement">minus</i>
 		</div>
 		<input
-			ref="inputNumber"
 			type="number"
 			:min="min"
 			:max="max"
 			:class="inputStyle"
 			:disabled="disabled"
 			:value="currentValue"
-			@input="setCurrentValue($event.target.value)"
+			@input="handleInput($event.target.value)"
 		/>
 		<div
 			v-if="max && showMaxNumber"
@@ -19,7 +18,7 @@
 			v-text="max"
 		/>
 		<div :class="increaseButtonStyle" @click="incrementValue">
-			<i class="yoco">plus</i>
+			<i :class="yocoClassIncrement">plus</i>
 		</div>
 	</div>
 </template>
@@ -54,7 +53,7 @@ export default {
 		formError() {
 			if (this.max !== null && this.currentValue > this.max) return true
 			if (this.min !== null && this.currentValue < this.min) return true
-			if (typeof this.currentValue === 'string' || isNaN(this.currentValue))
+			if (Number.isNaN(this.currentValue) || typeof this.currentValue !== 'number')
 				return true
 			return false
 		},
@@ -76,6 +75,18 @@ export default {
 			return {
 				'kt-input-number__button': true,
 				'kt-input-number__button--disabled': this.decrementDisabled,
+			}
+		},
+		yocoClassIncrement() {
+			return {
+				'yoco': true,
+				'yoco--disabled': this.incrementDisabled || this.disabled,
+			}
+		},
+		yocoClassDecrement() {
+			return {
+				'yoco': true,
+				'yoco--disabled': this.decrementDisabled || this.disabled,
 			}
 		},
 	},
@@ -102,15 +113,18 @@ export default {
 				}
 			}
 		},
-		disabled(newVal) {
+		disabled(newVal, oldVal) {
 			if (newVal) {
 				this.incrementDisabled = newVal
 				this.decrementDisabled = newVal
+			} else if (!newVal && oldVal) {
+				this.incrementDisabled = this.currentValue + this.step > this.max
+				this.decrementDisabled = this.currentValue - this.step < this.min
 			}
 		},
 	},
 	created() {
-		this.currentValue = typeof this.value === 'number' ? this.value : 0
+		this.currentValue = typeof this.value === 'number' && !Number.isNaN(this.value) ? this.value : 0
 	},
 	methods: {
 		incrementValue() {
@@ -123,7 +137,7 @@ export default {
 		},
 		/* Do not use input.stepUp - this method does not work in internet explorer and is not transpiled by Babel  */
 
-		setCurrentValue(value) {
+		handleInput(value) {
 			if (value === this.currentValue) return
 			this.currentValue =
 				typeof value === 'number' && !Number.isNaN(value)
@@ -133,9 +147,7 @@ export default {
 	},
 }
 </script>
-<style lang="scss">
-@import '../../kotti-style/_variables.scss';
-
+<style lang="scss" scoped>
 .kt-input-number__input {
 	width: auto;
 	border: 0;
@@ -152,7 +164,7 @@ export default {
 		text-align: right;
 	}
 	&--disabled {
-		background: $lightgray-300;
+		color: $lightgray-300;
 		&:hover {
 			cursor: not-allowed;
 		}
@@ -212,6 +224,9 @@ export default {
 	&--disabled:hover {
 		cursor: not-allowed;
 		background: $lightgray-300;
+	}
+	.yoco--disabled {
+		color: $lightgray-400;
 	}
 }
 </style>
--- a/packages/kotti-input-number/src/InputNumber.vue
+++ b/packages/kotti-input-number/src/InputNumber.vue
@@ -134,7 +134,7 @@ export default {
 }
 </script>
 <style lang="scss">
-// @import '../../kotti-style/_variables.scss';
+@import '../../kotti-style/_variables.scss';
 
 .kt-input-number__input {
 	width: auto;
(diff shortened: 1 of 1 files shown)


packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -72,35 +66,72 @@ export default {
...
+				typeof value === 'number' && !Number.isNaN(value)
+					? value
+					: parseFloat(value)
 		},
 	},
 }
 </script>
 <style lang="scss">
-@import '../../kotti-style/_variables.scss';
+// @import '../../kotti-style/_variables.scss';
```
carsoli: i think ur fix from the configs PR will eliminate the need for this line

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -1,30 +1,21 @@
...
 	<div :class="formGroupStyle">
-		<div
-			:class="decreaseButtonStyle"
-			@click="setCurrentValue(currentValue - step)"
-		>
+		<div :class="decreaseButtonStyle" @click="decrementValue">
 			<i class="yoco">minus</i>
 		</div>
 		<input
 			ref="inputNumber"
```
carsoli: i think u can remove this ref (not used)

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -1,30 +1,21 @@
...
 			:class="inputStyle"
-			type="number"
 			:disabled="disabled"
 			:value="currentValue"
-			:max="max"
-			:min="min"
 			@input="setCurrentValue($event.target.value)"
```
carsoli: hmm so i was wondering why we don't have this instead of this custom component then i remembered we wanted the yoco custom plus and minus (works consistently across web clients since this was a bug before) but to see that we had both the type=number, min/max logic AND the divs with the whole decrement/increment control :D wow

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -1,30 +1,24 @@
...
-			@click="setCurrentValue(currentValue - step)"
-		>
+		<div :class="decreaseButtonStyle" @click="decrementValue">
 			<i class="yoco">minus</i>
 		</div>
 		<input
 			ref="inputNumber"
-			:class="inputStyle"
 			type="number"
+			:min="min"
```
carsoli: adding the type and min/max would introduce two buttons if i'm not mistaken

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -72,29 +69,66 @@ export default {
...
 				'kt-input-number__button': true,
-				'kt-input-number__button--disabled':
-					this.currentValue == this.min || this.disabled,
+				'kt-input-number__button--disabled': this.decrementDisabled,
 			}
 		},
 	},
 	watch: {
 		value(val) {
 			this.currentValue = parseFloat(val)
```
carsoli: use the setter please

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -72,29 +69,66 @@ export default {
...
+			if (typeof this.max === 'number') {
+				if (newVal + this.step > this.max) {
+					this.incrementDisabled = true
+				} else if (!this.disabled) {
+					this.incrementDisabled = false
+				}
+			}
+		},
+		disabled(newVal) {
+			if (newVal) {
```
carsoli: i see why you did this but consider the following
if for some reason, disabled was externally set to `true` WHILE incrementDisabled and/or decrementDisabled were false (b/c min and max were not exceeded) 
then later disabled is set to false - which would make us skip this if-condition; the only way incrementDisabled would be then set to their correct values according to the value of min or max would be when currentValue changes (check scope of watcher on currentValue above)

So, have an else body in this watcher for this very rare scenario
OR ternaries: my point being the following
this.incrementDisabled = newVal?newVal:(newVal+this.step>this.max)
this.decremetnDisabled= newVal?newVal:(newVal-this.step<this.min)

having said this: the step prop is not reset to null with every increment/decrement action so this, too, is not right...

carsoli: i.e. to guard against using the step when no action is triggered now to change the value, make sure you reset it to null

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -72,35 +66,72 @@ export default {
...
+			if (this.decrementDisabled) return
+			this.currentValue -= this.step
+		},
 		/* Do not use input.stepUp - this method does not work in internet explorer and is not transpiled by Babel  */
+
 		setCurrentValue(value) {
 			if (value === this.currentValue) return
-			this.currentValue = typeof value === Number ? value : parseFloat(value)
-			this.$emit('input', this.currentValue)
+			this.currentValue =
```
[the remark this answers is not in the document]
Goldziher: yes, i intentionally did this. I used small modular components because i had to add guard. And the reason I had to use guards is that computed is not functioning as expected.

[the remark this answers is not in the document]
Goldziher: also regarding the watcher - i have to validate that the user did not input a false value before emitting - the component react by showing its invalid, but state should not be updated externally.

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -1,30 +1,24 @@
 <template>
 	<div :class="formGroupStyle">
-		<div
-			:class="decreaseButtonStyle"
-			@click="setCurrentValue(currentValue - step)"
-		>
+		<div :class="decreaseButtonStyle" @click="decrementValue">
```
carsoli: ik it's not on you
but changing the background color of a div when it has an icon shouldn't really do anything
does it?
so the kt-input_number--disabled would change the background color of the div but since it contains the icon, it won't do anything

soo 
move the class one step below and since the yoco icons are svgs, change the color property instead to light-gray

i'm looking to see if we have this implemented somewhere already

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -56,9 +49,10 @@ export default {
 			}
 		},
 		formError() {
-			if (this.currentValue > this.max && this.max !== null) return true
-			if (this.currentValue < this.min && this.min !== null) return true
-			if (this.currentValue === '' || isNaN(this.currentValue)) return true
+			if (this.max !== null && this.currentValue > this.max) return true
+			if (this.min !== null && this.currentValue < this.min) return true
+			if (typeof this.currentValue === 'string' || isNaN(this.currentValue))
```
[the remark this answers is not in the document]
Goldziher: I don't follow - 
max and min are both default null, so i have to check for this in case one of them is defined as 0., and because i want to regard them as numbers.  

the typeof current value string is indeed redundant - so i'll remove it. 

type is number for value - in the vue props will be a false positive if NaN is actually passed from without, or the user types in a value that will be parsed as NaN. So I have to check this - no?

I could actually move most of the logic to validators in the vue props - but we need this as a computed property to show the red border on the input from the html.  

min and max are both default null.

packages/kotti-input-number/src/InputNumber.vue
Goldziher: it should - but i had to comment this out for use in yoda when i copied the component, and then forgot this :)

packages/kotti-input-number/src/InputNumber.vue
Goldziher: kk. although we might keep it - its useful for testing because u can select the component by ref name.

packages/kotti-input-number/src/InputNumber.vue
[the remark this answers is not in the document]
carsoli: i personally prefer watchers to computed bc i've run into a case when computed didn't seem like it behaved correctly as well with a reactive property

don't remove the watcher, i edited my comment, just move the logic of currentValue watcher to the setter and make the watcher call the setter

this way you will always set after running all the needed checks 
and always use the setter when u assign a new value to the currentValue

packages/kotti-input-number/src/InputNumber.vue
Goldziher: i returned this back - i thought, but if not i should - 
this gives us the standard html invlaid if the user types in a value that is too high or low.

packages/kotti-input-number/src/InputNumber.vue
Goldziher: ha so i have return it :).

it didn't - i thought this was the bug i was seeing - but apparently not. The bug was that for some reason computing the decrease limit was breaking. I tried more elegant solutions first. I'll tell you in our meet because its long.

packages/kotti-input-number/src/InputNumber.vue
Goldziher: if value is changed from without we shouldn't emit, should we?

packages/kotti-input-number/src/InputNumber.vue
Goldziher: disabled() { this.forceUpdate() }?

packages/kotti-input-number/src/InputNumber.vue
Goldziher: kk

PR #143, opened 2020-02-13
fixed disabled color and added disabled to create hook

Goldziher:
This PR fixes wrong color in the input when disabled (shouldn't be there at all), and makes it so that if the component is created disabled, increment and decrement are disabled as well.

### DIFF
--- a/packages/kotti-input-number/src/InputNumber.vue
+++ b/packages/kotti-input-number/src/InputNumber.vue
@@ -53,7 +53,10 @@ export default {
 		formError() {
 			if (this.max !== null && this.currentValue > this.max) return true
 			if (this.min !== null && this.currentValue < this.min) return true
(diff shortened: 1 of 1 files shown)


packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -124,7 +127,14 @@ export default {
 		},
 	},
 	created() {
-		this.currentValue = typeof this.value === 'number' && !Number.isNaN(this.value) ? this.value : 0
+		this.currentValue =
+			typeof this.value === 'number' && !Number.isNaN(this.value)
+				? this.value
+				: 0
+		if (this.disabled) {
```
carsoli: IIRC you had a watcher on this.disabled, so using `immediate` flag on the watcher can make sure this happens so u don't add the same code to the created @Goldziher

PR #144, opened 2020-02-14
Unify popover dropdown

tomquinonero:
Popovers and dropdowns were already a bit confusing to me. 
They have the same purpose and `KtDropdown` is actually just calling `KtPopover`.
We can deprecate `KtDropdown`, but there is also `KtDropdownMenu` that does not exist anywhere else.

I then created `KtPopoverItem` that does the same as `KtDropdownMenu` was doing. And now `KtDropdownMenu` just call `KtPopoverItem` and is deprecated.

In kotti 2.0 we will totally remove `KtDropdown` and `KtDropdownMenu`.

There was also `KtDropdownButton` that'll stay and is now documented under the `buttons` section. 

Also `popper.js` was updated, and `KtPopover` handle overflowing issues, and you can use `popper.js` API via a `KtPopover` props called `options` for more advanced use.


Goldziher:
Great stuff. Will code review asap

Goldziher:
reviewed in depth, and all is great.

Issue #151, opened 2020-02-28
KtInput: not showing total number and red border on valid state


mvangent:
**Describe the bug**
KtInput is not showing total number and red border on valid state

**Additional context**
Bug happens in MES

FlorianWendelborn:
@carsoli @mvangent I think this can be closed? Can you confirm?

FlorianWendelborn:
Can be closed according to @carsoli.

PR #152, opened 2020-02-28
fix(#151): KtInputNumber Increment Issues & Refactor

carsoli:
Bug fixes:
1) we can input numbers that are invalid (>max or <min) --> they show an invalid style and store the new currentValue locally (in the component) but don't emit the faulty input (which is ok)
However, this non-emission does not allow for the faulty value to be recognized by a parent component so we allow submission if we use this field in a form (with the correct value since it's not emitted with an @input event) BUT it's confusing for the user to see an invalid styling, and be able to submit, then see that the actual saved value is not the FAULTY one they "typed" in 
2) initially, we didn't disable increment and decrement buttons and so if initially the component had a value == max or value == min, (2/2, for example) they could go up by one increment, before the buttons are really disabled --> (so they can have 3/2)

### DIFF
--- a/packages/kotti-input-number/src/InputNumber.vue
+++ b/packages/kotti-input-number/src/InputNumber.vue
@@ -29,7 +29,6 @@ export default {
 	name: 'KtInputNumber',
 	props: {
 		disabled: { default: false, type: Boolean },
-		fullWidth: { default: false, type: Boolean },
 		max: { default: null, type: Number },
 		min: { default: null, type: Number },
 		pattern: { required: false, type: String },
@@ -38,15 +37,12 @@ export default {
 		value: { default: 0, type: [Number, null] },
 	},
 	data() {
-		return {
-			currentValue: this.value ?? 0,
-		}
+		return { currentValue: this.value }
 	},
 	computed: {
 		formGroupStyle() {
 			return {
 				'kt-input-number form-group': true,
-				'form-group--100': this.fullWidth,
 				'form-group--error': this.formError,
 			}
 		},
@@ -64,7 +60,6 @@ export default {
 		inputClasses() {
 			return {
 				'kt-input-number__input': true,
-				'kt-input-number__input--100': this.fullWidth,
 				'kt-input-number__input--max': this.showMaxNumber,
 			}
 		},
@@ -176,9 +171,6 @@ export default {
 		background-color: unset;
 	}
 
-	&--100 {
-		width: 100%;
-	}
 	&--max {
 		text-align: right;
 	}
@@ -230,9 +222,6 @@ export default {
 		border: 1px solid $lightgray-400;
 		border-radius: $border-radius;
 
-		&--100 {
-			width: 100%;
-		}
 		&--error {
 			border-color: $red-500;
 		}
--- a/packages/kotti-input-number/src/InputNumber.vue
+++ b/packages/kotti-input-number/src/InputNumber.vue
@@ -113,14 +113,16 @@ export default {
 			immediate: true,
 			handler(newVal, oldVal) {
 				if (newVal === oldVal) return
+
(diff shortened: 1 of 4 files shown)


packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -183,12 +185,17 @@ export default {
...
-					? Number(value)
-					: 0
+			this.currentValue = this.isValidNumber(value)
+				? value
+				: this.isValidNumber(Number(value))
+				? Number(value)
+				: 0
+		},
+		hasFractionalValue(num) {
+			return num % 1 !== 0 ? true : false
```
mvangent: return num % 1 !== 0 would be enough

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -126,49 +121,56 @@ export default {
...
-			}
-		},
-		disabled(newVal, oldVal) {
-			if (newVal) {
-				this.incrementDisabled = newVal
-				this.decrementDisabled = newVal
-			} else if (!newVal && oldVal) {
 				if (typeof this.min === 'number') {
-					this.decrementDisabled = this.currentValue - this.step < this.min
+					if (newVal - this.step < this.min) {
```
mvangent: shouldn't this be? 

```
if (this.disabled) 
{
    this.decrementDisabled = true
 } 
else if (newVal - this.step <= this.min)
 {
    this.decrementDisabled = true
 }
else 
{
    this.decrementDisabled =  false
}
```

mvangent: also I think this could be wrapped in a separate function

```
updateDecrementDisabled(newVal){ // implementation }
```

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -126,49 +121,56 @@ export default {
...
+						this.decrementDisabled = true
+					} else if (!this.disabled) {
+						this.decrementDisabled = false
+					}
 				}
 				if (typeof this.max === 'number') {
-					this.incrementDisabled = this.currentValue + this.step > this.max
+					if (newVal + this.step > this.max) {
```
mvangent: same comments for max

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -1,220 +1,203 @@
...
+		</div>
+		<div :class="incrementButtonClasses" @click="incrementValue">
 			<i :class="yocoClassIncrement">plus</i>
 		</div>
 	</div>
 </template>
 
 <script>
+const DECIMAL_PLACES = 3
+const DECIMAL_SEPARATOR = (1.1).toLocaleString().substring(1, 2)
```
mvangent: fancy

mvangent:
@carsoli @FlorianWendelborn 💯 great improvement

packages/kotti-input-number/src/InputNumber.vue
FlorianWendelborn: @carsoli actually, I think we should probably change this to `(1.1).toLocaleString().replace(/\d/g, '')`, or what do you think? Some languages have really odd decimal separators.

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -227,39 +210,60 @@ export default {
 }
 
 .kt-input-number {
+	display: flex;
+	align-items: center;
+
+	width: inherit;
```
FlorianWendelborn: @carsoli I really think this should be `100%` instead. `inherit` just copies the parent value, which e.g. means that if the parent has `width: 50%`, the `KtInputNumber` would have another `width: 50%`, making it `25%` wide in total.

See here: https://stackoverflow.com/a/9374612/2857873

PR #156, opened 2020-03-03
Updating docs with ktcheckbox color

FlorianWendelborn:
Found this branch with unmerged changes. Freshly rebased on top of the latest master.

### DIFF
--- a/docs/pages/components/checkbox.vue
+++ b/docs/pages/components/checkbox.vue
@@ -4,7 +4,9 @@
 
 ## Basic usage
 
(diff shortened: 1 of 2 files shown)


tomquinonero:
Not quite ok with that. We should not allow things that should be CSS to live in the props or in JS in general!
If we want to stylize the icon we should just use CSS!

Goldziher:
tom and draz opposed this change the last time. I don't mind either way - i think allowing passing of props is a nice feature and a good api, but i don't have that a strong of an opinion on the subject.

PR #157, opened 2020-03-03
Update breadcrumbs to support query and url params

FlorianWendelborn:
Found this old branch & rebased it on the latest master. No idea what it’s about as the commit messages are pretty cryptic, but @Goldziher does.


Goldziher:
its as the title says. I eventually bypassed the issue in yoda externally, but would still be a nice feature.

FlorianWendelborn:
Closing because this is already supported. Just set `to` to `{ path: ..., query: ..., params: ... }`

Issue #167, opened 2020-03-09
different handling of nulls in KtInputNumber


carsoli:
**Is your feature request related to a problem? Please describe.**
KtInputNumber had disabled Increment/decrement buttons when the input is an empty string/bound value is null

**Describe the solution you'd like**
enable increment, decrement buttons, not just manually inputting a value, then setting the value to zero when either button is clicked so the user can begin to work with the input as intended

**Additional context**
better user experience than defaulting to zero at first, because it enables validation from outside the component and draws the user's attention to the null value
if it defaults to zero, it's a valid input right away, and the user can submit a form that contains this component with a value equal to zero without realizing that you should have input an actual value.

PR #168, opened 2020-03-09
enhance(#167): different handling of null values in KtInputNumber

carsoli:
if value is null, increment and decrement buttons are still enabled
if either buttons are pressed, we set internal currentValue to "0"
thus setting internal currentValueNumber to 0
then component behaves as normal
Notes: 
1) incentive for this is in issue description
2) FT can be done in docs (there is a null example) but also test other examples
issue #167

### DIFF
--- a/packages/kotti-input-number/src/InputNumber.vue
+++ b/packages/kotti-input-number/src/InputNumber.vue
@@ -96,8 +96,8 @@ export default {
 		},
 		isDecrementEnabled() {
 			return (
-				this.currentValueNumber === null ||
-				(!this.disabled &&
+				!this.disabled &&
+				(this.currentValueNumber === null ||
 					isInRange({
 						max: null,
 						min: this.min,
@@ -107,8 +107,8 @@ export default {
 		},
 		isIncrementEnabled() {
 			return (
-				this.currentValueNumber === null ||
-				(!this.disabled &&
+				!this.disabled &&
+				(this.currentValueNumber === null ||
 					isInRange({
 						max: this.max,
 						min: null,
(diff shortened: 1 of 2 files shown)


packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -96,24 +96,24 @@ export default {
...
 		isDecrementEnabled() {
 			return (
-				this.currentValueNumber !== null &&
-				!this.disabled &&
-				isInRange({
-					max: null,
-					min: this.min,
-					value: this.currentValueNumber - this.step,
-				})
+				this.currentValueNumber === null ||
```
mvangent: Looking good @carsoli , one suggestion:

I think we need to guard against this.disabled always and then add the range and null check logic, currently a disabled component has active increment/decrement buttons.
`
isDecrementEnabled() {
    if (this.disabled) return false

    this.currentValueNumber === null || isInRange({
 						max: null,
 						min: this.min,
 						value: this.currentValueNumber - this.step,
 					}))

}
`

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -96,24 +96,24 @@ export default {
...
 		isIncrementEnabled() {
 			return (
-				this.currentValueNumber !== null &&
-				!this.disabled &&
-				isInRange({
-					max: this.max,
-					min: null,
-					value: this.currentValueNumber + this.step,
-				})
+				this.currentValueNumber === null ||
```
mvangent: same

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -148,7 +148,8 @@ export default {
 	},
 	methods: {
 		decrementValue() {
-			if (!this.isDecrementEnabled || this.currentValueNumber === null) return
+			if (!this.isDecrementEnabled) return
+			if (this.currentValueNumber === null) return this.setValue('0')
```
mvangent: nice

PR #174, opened 2020-03-10
enhance(173): prohibit KtInputNumber from emitting input with null payload

carsoli:
prohibit emission of input with null values


FlorianWendelborn:
I don’t approve of this behavior. Either we support `null` or we don’t. This will cause unpredictable bugs.

Issue #176, opened 2020-03-10
KtInputNumber Does Not Emit Properly


FlorianWendelborn:
**Describe the bug**
`KtInputNumber` doesn’t behave as expected when used as a controlled input field. It sometimes does not emit the appropriate `onInput` event.

**To Reproduce**
```vue
<template lang="md">

<div>
	<KtInputNumber :value="value" @input="onInput"/>
</div>

</template>

<script>
export default {
	name: 'KtInputNumberDoc',
	data() {
		return {
			value: null,
		}
	},
	methods: {
		onInput(value) {
			console.error(value)
			this.value = value
		},
	},
}
</script>
```

1. Input `1`
2. Clear Input
3. Notice that the state is still `1`.

**Expected behavior**
The state should now be `null`.

**Additional context**

This bug was introduced by PR #174 (issue #173).

PR #178, opened 2020-03-16
#176 make KtInputNumber controlled

carsoli:
1. made `this.value` the source of truth
2. rename: `currentValue` was the `internalStringValue` --> now it's clearer
3. treat DECIMAL_SEPARATOR (e.g. `"."`) as null
4. added `isStepMultiple` check as validity check, which mirrors the behavior of a native number-typed input in html 
5. updated `isInRange` to include offset, to differentiate a `null` value, and a value that became `0` when the offset was passed as part of the value
6. currentValueNumber is now redundant because this.value is the source of truth
7. only update internal state in the watcher for value, thus, relying on `@input` and only emit 'input' when we have VALID user-input or empty strings (treated as null)

### DIFF
--- a/packages/kotti-input-number/src/InputNumber.vue
+++ b/packages/kotti-input-number/src/InputNumber.vue
@@ -12,7 +12,7 @@
 				:max="max"
 				:min="min"
 				:step="step"
-				:value="currentValue"
+				:value="internalStringValue"
 				@blur="handleBlur"
 				@input="handleInput($event.target.value)"
 			/>
@@ -28,7 +28,8 @@
 <script>
 const DECIMAL_PLACES = 3
 const DECIMAL_SEPARATOR = (1.1).toLocaleString().replace(/\d/g, '')
-const STRINGS_THAT_ARE_TREATED_AS_NULL = ['', '-', '+']
+const SIGN_STRINGS = ['-', '+']
+const STRINGS_THAT_ARE_TREATED_AS_NULL = [...SIGN_STRINGS, '']
 const LEADING_ZEROES_REGEX = new RegExp(`^0+([1-9]|0\\${DECIMAL_SEPARATOR}?)`)
 const TRAILING_ZEROES_REGEX = new RegExp(
 	`\\${DECIMAL_SEPARATOR}0*$|(\\${DECIMAL_SEPARATOR}\\d*[1-9])0+$`,
@@ -38,7 +39,8 @@ const VALID_REGEX = new RegExp(
 )
 
 const isInRange = ({ max, min, value }) =>
-	(max === null || value <= max) && (min === null || value >= min)
+	value === null ||
+	((max === null || value <= max) && (min === null || value >= min))
 
 const toNumber = (string) =>
 	STRINGS_THAT_ARE_TREATED_AS_NULL.includes(string)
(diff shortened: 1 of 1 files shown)


packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -97,23 +108,23 @@ export default {
 		isDecrementEnabled() {
 			return (
 				!this.disabled &&
-				(this.currentValueNumber === null ||
-					isInRange({
-						max: null,
-						min: this.min,
-						value: this.currentValueNumber - this.step,
-					}))
+				isInRange({
```
carsoli: no need for an internal currentValueNumber anymore --> since this.value is now the source of truth, and immediate watcher exists 
the null check is not necessary b/c we have a null check to the value property passed to the isInRange func (see above comment on func)

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -137,70 +148,91 @@ export default {
 	},
 	watch: {
 		value: {
-			handler(newNumber) {
+			handler(newNumber, oldNumber) {
 				const newString = toString(newNumber)
-				const shouldUpdate = this.currentValueNumber !== toNumber(newString)
+				const truncatedNumber = toNumber(newString)
```
carsoli: this is just refactored code with different naming, 
and two error handlers, obvs

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -137,70 +148,91 @@ export default {
...
 	},
 	methods: {
 		decrementValue() {
 			if (!this.isDecrementEnabled) return
-			if (this.currentValueNumber === null) {
-				const defaultString = this.min === null ? '0' : this.min + ''
-				this.setValue(defaultString)
+			if (this.value === null) {
+				const defaultNumber = this.min || 0
+				this.emitInput(defaultNumber)
```
carsoli: now we only emit, and then the value will pass the new number (only source of truth is this.value)

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -137,70 +148,91 @@ export default {
...
+					max,
+					min,
+					offset: 0,
+					value: nextNumber,
+				})
+
+			if (isTypedNumberValid) {
+				const shouldEmit = nextNumber !== this.value
+				if (shouldEmit) this.emitInput(nextNumber)
+				else this.setInternalStringValue(valueWithoutLeadingZeroes)
```
carsoli: `toNumber` has a `parseFloat` call which would treat `12.` as `12` for example, so if we would emit, we would prevent the user from ever typing `"12."` , because we will emit the parsed `12` instead which will trigger the watcher, then we set the `internalStringValue` to `"12"` again. 
nevertheless, we want to set the `internalStringValue` to bind it to the `value` of the `<input>` (check template), to show the `12.`

packages/kotti-input-number/src/InputNumber.vue
```diff
@@ -137,70 +148,91 @@ export default {
...
+				this.emitInput(defaultNumber)
 				return
 			}
 
-			this.setValue(toString(this.currentValueNumber + this.step))
+			this.emitInput(this.value + this.step)
 		},
-		setValue(newValue) {
-			const oldNumber = this.currentValueNumber
+		setInternalStringValue(value) {
```
carsoli: internal setter, that shouldn't emit to avoid the above-mentioned issue with the `12.`

mvangent:
nvm testing @carsoli I read in Slack you both did, and so did I just now. Is looking rock solid in the docs. So it has my full approval.

PR #179, opened 2020-03-18
fix(KtButton): Set Type Button Explicilty to Prevent Implicit Type Submit

FlorianWendelborn:
According to the HTML specs, every `<button>` element is implicitly `type="submit"`. This can cause forms to submit by clicking arbitrary buttons. This might cause it to click on a delete button when the user hits enter.

### DIFF
--- a/packages/kotti-button/src/button.vue
+++ b/packages/kotti-button/src/button.vue
@@ -4,7 +4,7 @@
 		:class="mainClasses"
 		:style="themeColor"
 		role="button"
-		type="button"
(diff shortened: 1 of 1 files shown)


packages/kotti-button/src/button.vue
```diff
@@ -4,6 +4,7 @@
 		:class="mainClasses"
 		:style="themeColor"
 		role="button"
+		type="button"
```
carsoli: :type="element" 
boop

PR #180, opened 2020-03-18
fix(#176): added guard against null value in isStepMultiple

carsoli:
apparently my upstream revived itself
isStepMultiple didn't guard against null values now that we allow them, and that results is arbitrary


carsoli:
@mvangent ding dong - very minor :) awaiting your review

PR #186, opened 2020-04-04
fix wrong datepicker import

mvangent:
Can be tested with https://gitlab.3yourmind.com/dev/yoda/-/merge_requests/97


carsoli:
For reference:
This is specific to our use-cases in Yoda
we don't need the KtDateInput because we have our own YodaDateInput which is a replica of DateInput.vue 
we, thus, only need KtDateInput

Issue #187, opened 2020-04-06
Value=0 results doesn't render row cell


carsoli:
**Describe the bug**
KtTableColumn doesn't render a number cell if it has a value = `0`

**To Reproduce**
copy the following snippet into the table docs
```

```
**Expected behavior**
row with value 0 renders

**Actual behavior** 
row with value 0 doesn't even render

**Screenshots**

PR #188, opened 2020-04-06
fix(#187): deprecated default function on KtTableColumn

carsoli:
it's always used as a string in our use-cases and the function doesn't do anything
updated docs accordingly


packages/kotti-table/src/constants.js
```diff
@@ -20,10 +20,8 @@ export const DEFAULT_RENDER_CELL = function DEFAULT_RENDER_CELL(
 	{ row, rowIndex, column, columnIndex, value },
 ) {
 	return [
-		column.formatter(value, row, column, columnIndex, rowIndex) ||
-			(typeof column.default === 'function'
-				? column.default(value, row, column, columnIndex, rowIndex)
-				: column.default),
+		column.formatter(value, row, column, columnIndex, rowIndex) ??
```
FlorianWendelborn, line 23: There’s one behavior in this line that we might want to re-consider actually. While I personally prefer only treating `null` as "empty", I think with the current table there’s a point to be made for potentially allowing `''` to count as empty as well, which this operator doesn’t do. So, maybe we should change this to `[null, undefined, ''].includes(formattedValue) ? column.default : formattedValue`.

FlorianWendelborn:
@mvangent Can you review this so this long-standing bug can be resolved?

PR #191, opened 2020-04-07
fix(KtButton): Re-apply Fix for Button Being Implicitly Type Submit

FlorianWendelborn:


packages/kotti-button/src/button.vue
```diff
@@ -4,6 +4,7 @@
 		:class="mainClasses"
 		:style="themeColor"
 		role="button"
+		v-bind="element === 'button' ? { type: 'button' } : {}"
```
tomquinonero, line 7: Do you mean `type: submit` ??

FlorianWendelborn: In the `true`-case? No, as this is the thing we’re fixing. Buttons in HTML implicitly default to `type="submit"`, which can cause a lot of unintended bugs like this one that I experienced with a different client:

1. There was a list-based view where rows could be added and removed. Every row has an `X` button to delete itself.
2. The row happened to be implemented as a form because it was convenient
3. BUG: When the user hit <kbd>return</kbd>, the form tried to click the "submit" button, which happened to just delete the row

Since this behavior can easily lead to strange bugs like this, we decided to make buttons stay buttons. The existing behavior can still be used by just settings `onSubmit`/`onClick` on the button, instead of praying that the form will magically find the right button to "click".

FlorianWendelborn:
@tomquinonero @carsoli can we merge this bugfix?

FlorianWendelborn:
Closing in favor of #392

PR #195, opened 2020-04-11
fixed emitting mouse event instead of value

Goldziher:
this is a really small PR - the kt-inline component is emitting the mouse event instead of its value on `@confirm` - I discovered this when i tried to convert from the YodaInLineEdit to this component, assuming the two were identical. I'd recommend refactoring this component in the improvement epic, since it could do with some love. For the time being I am using the `@change` event emitted by the component, which works correctly (I don't think we need so many events emitted from one component TBH).

### DIFF
--- a/packages/kotti-inline-edit/src/InlineEdit.vue
+++ b/packages/kotti-inline-edit/src/InlineEdit.vue
@@ -86,10 +86,10 @@ export default {
 		},
 	},
 	methods: {
-		handleConfirm(event) {
+		handleConfirm() {
 			this.editMode = false
 			this.preValue = this.currentValue
-			this.$emit('confirm', event)
+			this.$emit('confirm', this.currentValue)
 		},
 		handleDismiss(event) {
 			this.editMode = false


packages/kotti-inline-edit/src/InlineEdit.vue
```diff
@@ -86,10 +86,10 @@ export default {
 		},
 	},
 	methods: {
-		handleConfirm(event) {
+		handleConfirm() {
```
carsoli, line 89: don't approve of not emitting $event - better give the user the entire control

packages/kotti-inline-edit/src/InlineEdit.vue
```diff
@@ -86,10 +86,10 @@ export default {
...
 			this.editMode = false
 			this.preValue = this.currentValue
-			this.$emit('confirm', event)
+			this.$emit('confirm', this.currentValue)
```
carsoli, line 92: also it seems that we sync currentValue with value so it shouldn't be problematic
but if we change confirm then should we consider also changing the dismiss for consistency ?

mvangent: this.$emit('confirm', {...event, currentValue: this.currentValue }) would keep the event data being available and have the current value available.

FlorianWendelborn:
Closing as this obviously isn’t urgent and as new components will land eventually.

PR #201, opened 2020-05-04
fix(KtDatePicker): export KtDateInput as used in the docs - default e…

carsoli:
this with relevance to 41f4c8e4516e8bc97fb972a5ec1d6916c82c7550 ; to undo it, because it broke the docs and it broke things in yoda bc the api for KtDatePicker was stll assuming it is a KtateInput - the fix in yoda will be to use the api for KtDatePicker's api tho

### DIFF
--- a/packages/kotti-datepicker/index.js
+++ b/packages/kotti-datepicker/index.js
@@ -6,5 +6,5 @@ KtDatePicker.install = function(Vue) {
 	Vue.component(KtDatePicker.name, KtDatePicker)
 }
 
-export { KtDateInput }
+export { KtDateInput, KtDatePicker }
 export default KtDatePicker
--- a/packages/kotti-datepicker/index.js
+++ b/packages/kotti-datepicker/index.js
@@ -1,7 +1,10 @@
 import KtDatePicker from './src/DatePicker'
+import KtDateInput from './src/DateInput'
 
 KtDatePicker.install = function(Vue) {
+	Vue.component(KtDateInput.name, KtDateInput)
 	Vue.component(KtDatePicker.name, KtDatePicker)
 }
 
+export { KtDateInput }
 export default KtDatePicker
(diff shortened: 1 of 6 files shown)


carsoli:
@mvangent and @FlorianWendelborn updated the pr 
two liner change 
and the test can be done in the docs
it now is finally fully functional :everything-is-fine:

packages/kotti-datepicker/src/DatePicker.vue
```diff
@@ -109,9 +109,11 @@ export default {
 	methods: {
 		monthPrevious() {
 			this.date = this.date.subtract(1, 'month')
+			this.$emit('KtDateSelected', this.date.toDate())
```
FlorianWendelborn, line 112: Not a fan of the event name, but I’m assuming you chose this one for compatibility reasons

packages/kotti-datepicker/index.js
```diff
@@ -1,7 +1,10 @@
 import KtDatePicker from './src/DatePicker'
+import KtDateInput from './src/DateInput'
 
 KtDatePicker.install = function(Vue) {
+	Vue.component(KtDateInput.name, KtDateInput)
 	Vue.component(KtDatePicker.name, KtDatePicker)
 }
 
+export { KtDateInput }
```
FlorianWendelborn, line 9: Would prefer it if the `KtDatePicker` would also be exported like this

packages/kotti-datepicker/src/DatePicker.vue
carsoli: ye exactly - don't  see anyone relying on v-model anyways and yoda has a wrapper around this that emits input allowing for the vue magical v-model

packages/kotti-datepicker/index.js
carsoli: ait

PR #205, opened 2020-05-12
fix(defaultLayout): provided logourl to KtNavBar

carsoli:
logo now doesn't fall to alt in default layout in docs


FlorianWendelborn:
@carsoli @tomquinonero What’s the state of this?

PR #207, opened 2020-05-18
Include packages/ into npm bundle

Tradias:
Include packages/ into npm bundle to allow consumption of individual kotti components without the need of redistributing all of them.

This is especially useful when creating web-components based on kotti components.

Compare the difference of using:

`import { KtButton, KtSwitch } from '@3yourmind/kotti-ui'`

kotti-full

to this PR:

```
import KtSwitch from '@3yourmind/kotti-ui/packages/kotti-switch'
import KtButton from '@3yourmind/kotti-ui/packages/kotti-button'
```

kotti-individual-components

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
 	},
 	"files": [
 		"dist",
-		"packages/kotti-style/_variables.scss"
+		"packages"
 	],
 	"husky": {
 		"hooks": {


FlorianWendelborn:
I think this should rather be solved by using tree shaking and a better transpilation chain.

FlorianWendelborn:
@Tradias we have the improved build setup now. However, kotti is still not side-effect free, which means that tree-shaking will be very limited. It should be possible to fix that though.

I’ll close this PR as it’ll conflict with the new build setup entirely anyway. So, not much we can do right now. I’ll open an issue for removing the side-effects as well.

PR #210, opened 2020-06-12
feature(#215): Add KtForm & KtFields

carsoli:
This PR implements #215.

### Features

- [x] implement `useField` hook used in all KottiField components for common logic
- [x] implement `KtForm` which provides values for child `KtField`s via a `KtFormContext`
- [x] implement `KtFormControllerList` & `KtFormControllerObject` that are sub-contexts of the `KtFormContext` to make using array items and deep objects easier
- [x] KtFields:
  - [x] KtFieldText
  - [x] KtFieldTextArea
  - [x] KtFieldSelect (folder containing the logic for both selects) - each of which is a wrapper for el-ui selects
  - [x] KtFieldSingleSelect 
  - [x] KtFieldMultiSelect
  - [x] KtFieldCheckbox
  - [x] KtFieldCheckboxGroup (don't confuse with a radiogroup - this allows for more than one check, it's stylistic however and accepts the list of selected options - has ~ similar api to radiogroup but not strict to one checked input at a time)
  - [x] KtFieldRadiogroup
  - [ ] KtFieldDate :construction: (considering the min/max date/time props and a reiteration over disabledDate)
    - [x] KtFieldDate 
    - [x] KtFieldDateRange  
    - [x] KtFieldDateTime
    - [x]  KtFieldDateTimeRange 
  - [x] KtFieldToggle(Group) [for Checkbox(Group) and Switch(Group)]
  - [x] KtFieldInputNumber (remaining style fixes)
- [ ] documentation for components & form
---

Combinatory effort of @FlorianWendelborn, @tomquinonero, and @carsoli.


packages/next/kotti-field-radio-group/KtFieldRadioGroup.vue
```diff
@@ -0,0 +1,182 @@
...
+			type: Array,
+			validator: (options: KottiFieldRadioGroup.Props['options']) =>
+				[...new Set(options.map(({ value }) => value))].length ===
+				options.length,
+		},
+	},
+	setup(props: KottiFieldRadioGroup.Props, { emit }) {
+		const field = useField<KottiFieldRadioGroup.Value | null>({
+			emit,
+			isCorrectDataType: (value): value is KottiFieldRadioGroup.Value | null =>
```
mvangent, line 63: same as prev

packages/next/kotti-field-radio-group/KtFieldRadioGroup.vue
```diff
@@ -0,0 +1,182 @@
...
+				['number', 'string', 'boolean'].includes(typeof value) ||
+				value === null,
+			isEmpty: (value) => value === null,
```
mvangent, line 66: same as prev

packages/next/kotti-field-radio-group/KtFieldRadioGroup.vue
```diff
@@ -0,0 +1,182 @@
...
+			props,
+			supports: KOTTI_FIELD_RADIO_GROUP_SUPPORTS,
+		})
+
+		const { forceUpdate, forceUpdateKey } = useForceUpdate()
+		const name = ref(`autoGeneratedRadioName${nameIndex++}`)
+
+		return {
+			inputProps: computed(() => ({
```
mvangent, line 75: same with regards to defining functions in the return object

packages/next/kotti-field-radio-group/KtFieldRadioGroup.vue
```diff
@@ -0,0 +1,182 @@
...
+
+				forceUpdate()
+			},
+		}
+	},
+})
+</script>
+
+<style lang="scss">
+@import '../kotti-field/mixins';
```
mvangent, line 94: perhaps define these styles in a separate file as well?

packages/next/kotti-field-select/KtFieldMultiSelect.vue
```diff
@@ -0,0 +1,216 @@
...
+
+		const isDropdownOpen = ref(false)
+
+		const scheduleFocusAfterFieldClick = ref(false)
+		const elSelectClasses = computed(() => ({
+			'el-select--disabled': field.isDisabled,
+		}))
+
+		return {
+			elMultipleSelectProps: computed(
```
mvangent, line 124: would consider to separate definitions from the return object

packages/next/kotti-field-select/KtFieldSingleSelect.vue
```diff
@@ -0,0 +1,153 @@
...
+			field,
+			inputSelectors: ['.el-input__inner'],
+			ktFieldRef,
+		})
+
+		const isDropdownOpen = ref(false)
+		const scheduleFocusAfterFieldClick = ref(false)
+
+		return {
+			elSelectRef,
```
mvangent, line 94: same

packages/next/kotti-field-text-area/KtFieldTextArea.vue
```diff
@@ -0,0 +1,133 @@
...
+				typeof value === 'string' || value === null,
+			isEmpty: (value) => value === null,
+			props,
+			supports: KOTTI_FIELD_TEXT_AREA_SUPPORTS,
+		})
+
+		const { forceUpdate, forceUpdateKey } = useForceUpdate()
+
+		return {
+			field,
```
mvangent, line 38: same

packages/next/kotti-field-text/KtFieldText.vue
```diff
@@ -0,0 +1,69 @@
...
+			isCorrectDataType: (value): value is string | null =>
+				typeof value === 'string' || value === null,
+			isEmpty: (value) => value === null,
+			props,
+			supports: KOTTI_FIELD_TEXT_SUPPORTS,
+		})
+
+		const { forceUpdate, forceUpdateKey } = useForceUpdate()
+
+		return {
```
mvangent, line 36: same

packages/next/kotti-field-toggle/components/ToggleBox.vue
```diff
@@ -0,0 +1,159 @@
...
+
+<script lang="ts">
+import { defineComponent } from '@vue/composition-api'
+
+export default defineComponent({
+	name: 'ToggleBox',
+})
+</script>
+
+<style lang="scss">
```
mvangent, line 30: styles in a separate file?

packages/next/kotti-field/KtField.vue
```diff
@@ -0,0 +1,394 @@
...
+		{ emit }: SetupContext,
+	) {
+		const validationType = computed(() => props.field.validation.type)
+		const showValidation = computed(
+			() => !(props.field.hideValidation || validationType.value === null),
+		)
+
+		const translations = useTranslationNamespace('KtFields')
+
+		return {
```
mvangent, line 126: same with regards to defining in the return obj

packages/next/kotti-field/KtField.vue
```diff
@@ -0,0 +1,394 @@
...
+				padding: 0.25em;
+				color: var(--support-info);
+				cursor: none;
+				background-color: var(--ui-01);
+				border: 1px solid var(--ui-02);
+				border-radius: 3px;
+				transform: translateX(calc(-50% + 5px));
+			}
+
+			// &__icon {
```
mvangent, line 308: commented out. code can go?

packages/next/kotti-field/hooks.ts
```diff
@@ -0,0 +1,282 @@
...
+			: context === null
+			? false
+			: context.hideValidation.value,
+	)
+
+	const isMissingRequiredField = computed(
+		(): boolean => !props.isOptional && isEmpty(currentValue.value),
+	)
+
+	const validation = computed(
```
mvangent, line 146: both the property on the reactive field object this computed value are named `validation`, could you add a differentiation?

.eslintrc.js
```diff
@@ -36,19 +37,51 @@ module.exports = {
...
+				'sonarjs/no-duplicate-string': 'off',
+				'no-console': 'off',
+				'import/order': [
+					'warn',
+					{
+						alphabetize: {
+							order: 'asc',
+							caseInsensitive: true,
+						},
+						'newlines-between': 'always',
```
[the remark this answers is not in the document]
FlorianWendelborn: Having this set to sometimes (or whatever that value is) would be inconsistent, so the only alternative would be never. I think the groups are a useful feature and thus shouldn’t be hidden entirely.

docs/components/YocoPreview.vue
```diff
@@ -33,7 +35,7 @@ export default {
 					this.copySuccess = true
 					setTimeout(() => {
 						this.copySuccess = false
-					}, 2000)
+					}, 2 * TimeConversion.MILLISECONDS_PER_SECOND)
```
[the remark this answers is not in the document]
FlorianWendelborn: This is from a library that I made called `@metatypes/units`. It contains type definitions for all sorts of time constants. So, we’d have to fork it to make that change happen. Personally, I think this is good as-is

packages/next/kotti-field-toggle/components/ToggleBox.vue
carsoli: This is the styles-only component of the `checkbox__box` (we only stylize the "box" itself here) 

---

1. `ToggleInner` includes `v-if` / `v-else` on `ToggleBox` or `ToggleSwitch`. It is the logic wrapper for both the switch and checkbox because they basically internally are both `<input type='checkbox' .../>`
2. `ToggleBox.vue` &  `ToggleSwitch.vue` are the style components, one works on `checkbox__box`, other works for the stadium part of the switch
3. `KtFieldToggle` is the field itself, it wraps `<ToggleInner />` 
4. to extend 3, there is also `KtFieldToggleGroup` which basically wraps multiple `ToggleInner` (`v-for` depending on how many options)

sorry to be writing all this when u may have possibly seen it but just making sure we r on the same page

---

1. no point (in our opinion though) unless u see smth specific ofc to move out styles from `ToggleBox` anywhere else since that's its sole purpose as a file.
2. if u mean in general have separate style sheets for folders with multiple things, i think we only do it gets hugee (example `KtFieldSelect`s and `KtFieldDate`s) -- for obvious reasons - the styles are at least 200+ lines there 
but other than that - this component specially is realistically JUST to style 
3. If u meant smth entirely different pls say

packages/kotti-toaster/KtToaster.vue
```diff
@@ -38,7 +42,13 @@ export default {
 		this.$root.$off('vue-yodify', this.addNotification)
 	},
 	methods: {
-		addNotification({ id = generateId(), text, type, duration = 3000 }) {
+		addNotification({
+			id = generateId(),
+			text,
+			type,
+			duration = DEFAULT_DURATION_IN_SECONDS *
```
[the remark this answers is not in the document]
carsoli: we dont mind renaming that, it really makes no difference but is there a general reason that we should in the future look out for as to why amount is more preferable than duration here

docs/components/YocoPreview.vue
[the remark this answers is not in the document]
mvangent: not super important indeed, agreed

.eslintrc.js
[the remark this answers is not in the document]
mvangent: ok

packages/kotti-toaster/KtToaster.vue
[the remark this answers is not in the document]
mvangent: its more semantics, v minor, would be fine with it being ignored

packages/next/kotti-field-toggle/components/ToggleBox.vue
mvangent: yeah makes sense, I was reasoning from it getting towards huge, but script is small, so this is basically about style.  Also, it was more to raise the point that styles are sometimes separated and wondered about the logic, so thanks for explaining it.  All of this is not a blocker either for me.

packages/next/kotti-field/hooks.ts
carsoli: @mvangent 
> The main suggestion I have left to reconsider is to define functions outside of return object of the component script. This would allow for stronger grouping of logic that belongs together. Defining the functions in the return objects feel a bit Vue 2-y, and from a personal standpoint I think the api should only contain properties that are already defined (more readable).

we are currently discussing this - we r going to look into that refacor :+1:

packages/next/kotti-field-date/hooks.ts
```diff
@@ -0,0 +1,237 @@
...
+}
+
+export type ElDateWithInternalAPI = ElDate & {
+	blur(): void
+	picker: Vue & { width: number; $el: HTMLElement }
+	pickerVisible: boolean
+	referenceElm: Element
+}
+
+const getDateComponent = <DATA_TYPE extends Values>({
```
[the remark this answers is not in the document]
carsoli: they r only relevant to the el-ui fixing hooks though, i'm not in favor of either but if u think that it's better to maintain the heirarchy we keep, i will do  it

PR #236, opened 2020-07-12
build(#233): Add Rollup & Lerna

FlorianWendelborn:
Re-structures the entire build system. Very experimental right now.

---

## Initial Setup

```bash
yarn config set workspaces-experimental true
yarn install
yarn run lerna bootstrap
```

## Test-Out Composition API Plugin (Yoda use-case)

```bash
yarn --cwd packages/kotti-ui build
yarn --cwd packages/documentation serve
```


packages/kotti-ui/source/kotti-style/index.scss
```diff
@@ -6,7 +6,7 @@
 @import 'transitions';
 
 // Reset and dependencies
-@import '~normalize.css';
+// @import '~normalize.css';
```
mvangent, line 9: @FlorianWendelborn don't we need normalize here?

packages/kotti-ui/source/next/kotti-field/mixins.scss
```diff
@@ -395,9 +395,13 @@
 
 @mixin sizes {
 	$fontSizes: (
-		'small': var(--font-size-small),
-		'medium': var(--font-size-medium),
-		'large': var(--font-size-large),
+		// FIXME: reeeeee
```
mvangent, line 398: they're all (sm, md, lg) 12px?

carsoli:
https://github.com/3YOURMIND/kotti/pull/236#discussion_r454612665
@mvangent can't answer u on the thread
no they r not all 12px, but there were errors thrown bc of those vars

@FlorianWendelborn i do think this was before adding the sass loader that resolved it but i can be mixing things up bc everything was happening at the same time

packages/kotti-ui/source/kotti-style/index.scss
FlorianWendelborn: @mvangent `normalize.css` is now imported in `index.ts`

PR #256, opened 2020-08-17
fix(scripts): stylelint scripts moved to each package separately

carsoli:
the files to check differ according to each package, and directories like node_modules and dist should not be checked
otherwise, output dist contains css that's checked

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,10 +120,9 @@
 	"scripts": {
 		"check": "yarn run check:script && yarn run check:style",
 		"check:script": "eslint --ignore-path .gitignore '**/*.{js,ts,vue}'",
-		"check:style": "stylelint '**/*.{css,scss,vue}'",
 		"lint": "yarn run lint:script && yarn run lint:style",
 		"lint:script": "yarn run check:script --fix",
-		"lint:style": "yarn run check:style --fix",
+		"lint:style": "lerna run check:style --fix",
 		"postpublish": "yarn run postpublish:deploy",
 		"postpublish:deploy": "yarn --cwd packages/documentation run deploy",
 		"prepublishOnly": "yarn run lint && lerna run build",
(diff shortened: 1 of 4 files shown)


carsoli:
@krysopath ran scripts along - 
merging for now and open for changes later

krysopath:
womm

package.json
```diff
@@ -120,10 +120,9 @@
 	"scripts": {
 		"check": "yarn run check:script && yarn run check:style",
 		"check:script": "eslint --ignore-path .gitignore '**/*.{js,ts,vue}'",
-		"check:style": "stylelint '**/*.{css,scss,vue}'",
 		"lint": "yarn run lint:script && yarn run lint:style",
 		"lint:script": "yarn run check:script --fix",
-		"lint:style": "yarn run check:style --fix",
+		"lint:style": "lerna exec --ignore @3yourmind/storybook 'yarn run check:style --fix'",
```
FlorianWendelborn, line 125: Why did you use lerna exec/ignore here? I think lerna run does the same

Probably something I'm missing tbh

carsoli: lerna treats the `fix` option like it's own if i don't use the exec

since `lint:style` was always the same for all packages that need it, i just put the `check:style` there
and if u run `lerna run check:style --fix` u'll get an error about the fix
so i have to treat the yarn script like a command and pass the args as part of it

as for the ignore
storybook's structure is unclear to me, and i don't see anything yet since it's a PoC so i decided to ignore it for now
alternatively could have used `scope` but same same and in the future the ignore should be removed anyways

hope that answers it

packages/documentation/package.json
```diff
@@ -30,6 +30,7 @@
 	"private": true,
 	"scripts": {
 		"build": "DEPLOY_ENV=GH_PAGES nuxt generate",
+		"check:style": "stylelint ./{assets,components,layouts,pages}{/**,}/*.{css,scss,vue}",
```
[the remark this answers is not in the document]
carsoli: was playing around with glob all morning, if i specificy **/* then i ignore any css that exists directly under assets, components, etc

u'll see i used the same `{/**,}` for all the packages

PR #257, opened 2020-08-18
fix/refact(KtCol): refactoring broke KtCol

carsoli:
introduced KT_ROW_CONTEXT for proper provide/inject
(added relevant types and constants)
injected CONTEXT in KtCol

bug-fix: most props defaulted to null, which matched the `typeof prop === 'object'` which threw an error

cleanup: restricted all media-query props to be of type number only and we double-checked all usages

simplified the classes computation accordingly

TLDR; fixed bug and simplified KtCol, and refactored types, and used proper context provision

Co-authored-by: Florian Wendelborn <1133858+FlorianWendelborn@users.noreply.github.com>


FlorianWendelborn (approved):
LGTM

PR #258, opened 2020-08-19
feature(sass-node-modules-importer): First Implementation

FlorianWendelborn:
This commit adds a custom sass importer that’s able to properly resolve node_modules, package.json style property, etc.

### DIFF
--- a/packages/sass-node-modules-importer/source/logger.ts
+++ b/packages/sass-node-modules-importer/source/logger.ts
@@ -1,0 +1,10 @@
+import { Options } from './types'
+
+export const log = (
+	options: Options,
+	message: string,
+	...more: unknown[]
+): void => {
+	// eslint-disable-next-line no-console
+	if (options.debug) console.debug(`importer: ${message}`, ...more)
+}
--- a/packages/sass-node-modules-importer/source/resolvers/directory.ts
+++ b/packages/sass-node-modules-importer/source/resolvers/directory.ts
@@ -4,6 +4,8 @@ import * as path from 'path'
 import { log } from '../logger'
 import { Resolver } from '../types'
 
+import { resolveFromExtension } from './file'
+
 /**
  * Directory resolver that attempts to look for a package.json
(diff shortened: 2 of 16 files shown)


packages/sass-node-modules-importer/source/logger.ts
```diff
@@ -0,0 +1,10 @@
+import { Options } from './types'
+
+export const log = (
+	options: Options,
+	message: string,
+	...more: unknown[]
```
carsoli, line 6: don't know if it's safe to allow the log usage to accept any other args 
destructing the more in the console.debug call would be faulty in the case of a boolean for example

packages/sass-node-modules-importer/source/resolvers/directory.ts
```diff
@@ -0,0 +1,37 @@
...
+	log(options, 'found package.json')
+	const packageJson = JSON.parse(fs.readFileSync(packageJsonPath).toString())
+
+	if (!Object.prototype.hasOwnProperty.call(packageJson, 'style')) {
+		log(options, 'no style specified in package.json, aborting')
+		return null
+	}
+
+	log(options, 'found style')
+	return path.resolve(directory, packageJson.style)
```
carsoli, line 23: unless this returns null by default

carsoli: this will be faulty

packages/sass-node-modules-importer/source/logger.ts
FlorianWendelborn: This is handled by `console.debug` accepting `any[]`

packages/sass-node-modules-importer/source/resolvers/directory.ts
FlorianWendelborn: This is intentional. I added a JSDoc comment to explain why above.

carsoli (approved):

Issue #255, opened 2020-08-14
Move Yoco to Kotti


carsoli:
now that we have lerna and packaging and whatso:
- move yoco to kotti
- properly maintain it
- TODO: more specifics later

---

See <https://3yourmind.atlassian.net/browse/B3-8200>

carsoli:
FYI @mvangent

PR #259, opened 2020-08-22
fix(#255): move yoco to kotti

FlorianWendelborn:
Also managed to remove the dependency on Icomoon. There’s now a very simple local build system instead. The icons are just defined as `.svg` files inside the `svg` folder and need to by added to the `types.ts` `enum` to be picked-up by the builder.

This also ensures that we’ll always have an up-to-date yoco enum, as icons won’t be able to exist in the fonts without being added to it.

---

## Open Problems

- [ ] Someone who knows Sketch should adjust the sketch to export proper `.svg` names (`_` instead of `-` and `3d_box` is now `box_3d`)
- [x] Make sure **all** icons have `.svg`s and properly work. There are some missing from the demo page now, as they were never generated as `.svg` files for some reason
- [x] Make sure the new workflow works for designers as well, this is only tested from a programmer perspective but appears to work very well
- [x] Make sure the icons still work in everred browsers (looking at you, Internet Exploder)

## Notable Changes Requiring a Major Bump

1. `3d_box` -> `box_3d` (variable names shouldn’t start with numbers, making this one pretty inconvenient)
2. `/dist/yoco-icon.css` -> `/style.css`
3. `/dist/index.js` introduced with `Icons` enum
4. Dropped support for `.svg` fonts due to the font generator not supporting them and them (presumably) not being too important
5. `/dist/svg` -> `/svg`
6. Renamed raw `.svg` files to remove `-` (makes it consistent with the ligatures)

---

See #255


ostoc:
> @ostoc Would be nice if you could also check if anything seems wrong about this. One of the few things we lost due to this is support for `.svg` fonts. I think that tradeoff is worth it, but I don’t know the reasons why they were initially supported.

The `.svg` font was added to support old browser, and automatically generated by `icomoon`. I don't think there is a huge impact if we drop the svg font support.

https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/SVG_fonts

packages/yoco/.gitignore
```diff
@@ -0,0 +1,5 @@
+dist/
+gh-pages/
+docs/
```
carsoli, line 3: out of curiosity really, can we combine lines 3-5

FlorianWendelborn: I'm not aware of a way to do that.

In case you're asking because you're unaware of what it does, this ignores the folder besides the two files I mentioned

carsoli: i guess you mean except the two files, but no, i assumed and checked to verify 
was just curious if there was a shorter one-liner for it, but apparently not, that's the recommended method

packages/yoco/source/generate.ts
```diff
@@ -0,0 +1,26 @@
+import * as path from 'path'
+
+import IconFontBuildr from 'icon-font-buildr'
+
+import { Icon } from './types'
```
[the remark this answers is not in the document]
carsoli: and that is should be Icon*s* actually but that's pretty minor

packages/yoco/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+**With module bundler**
+
+```scss
+@import '~@3yourmind/yoco'; // might need a sass importer like @3yourmind/sass-node-modules-importer
+```
+
+```typescript
+// enum of icons and the associated ligature / svg name
+import { Icons } "@3yourmind/yoco";
```
carsoli, line 31: according to what i saw on the index and the types, it's `Icon` not `Icons`

packages/yoco/source/types.ts
```diff
@@ -0,0 +1,115 @@
+export enum Icon {
```
carsoli, line 1: did you find the full list or how did you figure?

packages/yoco/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
carsoli, line 1: did lerna add this? :D

yarn.lock
```diff
@@ -2,11 +2,6 @@
 # yarn lockfile v1
 
 
-"@3yourmind/yoco@*":
```
carsoli, line 5: YEY

yarn.lock
```diff
@@ -3218,6 +3213,11 @@
   resolved "https://registry.yarnpkg.com/@types/less/-/less-3.0.1.tgz#625694093c72f8356c4042754e222407e50d6b08"
   integrity sha512-dBp05MtWN/w1fGVjj5LVrDw6VrdYllpWczbUkCsrzBj08IHsSyRLOFvUrCFqZFVR+nsqkrRLNg6oOlvqMLPaSA==
 
+"@types/lodash@^4.14.117":
```
carsoli, line 3216: why do we suddenly have `@types/lodash` bc of this? does yoco even really on lodash or did you do an update as well or what :thinking:

carsoli (changes requested):
only for `Icon` vs `Icons` enum clarification
I was asking if `Icon` should be on a `constants.ts` bc I didn't know enum's can be treated as Objects at runtime
there was a line with `Object.value(enumThat'sCaleldIcon)`; that's my only source of confusion.

only blocker is either an update to readme or to enum name then i will approve

packages/yoco/source/types.ts
carsoli, line 1: yey

packages/yoco/yarn.lock
FlorianWendelborn: I thought it’s a fuck-up by me, but actually it turns out to be [from the original source code](https://github.com/3YOURMIND/yoco/blob/5a40fbf4d3f9915e66b174fbe0d8f6ca082524a0/yarn.lock). This can safely be deleted though.

packages/yoco/source/generate.ts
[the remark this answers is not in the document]
FlorianWendelborn: I actually used to use plural for `enum`s as well, but this was intentionally made to be singular as that’s apparently the industry standard:

* <https://stackoverflow.com/questions/1335959/singular-or-plural-for-enumerations>
* <https://stackoverflow.com/questions/1405851/enum-naming-convention-plural>

packages/yoco/README.md
FlorianWendelborn: Good catch, this is because I renamed `Icons` to `Icon` at some point IIRC.

yarn.lock
FlorianWendelborn: Those most likely come from `yoco`s `devDependencies` for the new build process:
<https://github.com/3YOURMIND/kotti/pull/259/files#diff-85807792aef76b0bddd8749ae410dd0aR19-R20>

packages/yoco/README.md
carsoli: will resolve and wait for commit then :+1:

carsoli (approved):

FlorianWendelborn:
@ostoc I rebase-fixed your commit authorship information so you’ll get the contributions from your `yoco` commits correctly attributed to your GitHub profile

Issue #242, opened 2020-07-23
KtFieldNumber should have the plus and minus buttons hideable 


tomquinonero:
We need to be able to hide these <kbd>+</kbd> and <kbd>-</kbd> buttons to have the field looking more like any other field.

If there is no maximum or the maximum is hidden, we should also align the field content left (instead of center).

PR #260, opened 2020-08-22
feature(#242): Add Show Change Buttons

FlorianWendelborn:
~~TBD: Should we make this a `hideChangeButtons` flag instead?~~

Answer: Yes


tomquinonero (approved):
Reviewed, looks good

PR #266, opened 2020-09-04
fix(#272): Add Style for Date Ranges

tomquinonero:
This highlight the range when selecting a date range. It uses some trickery and `!important` because I'm overwriting element-ui's styles there.

Screenshot_2020-09-04 Kotti Design System


packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -7,6 +7,7 @@
 	>
 		<i v-if="loading" class="kt-circle-loading" />
 		<i v-else-if="icon" class="yoco" v-text="icon" />
+		HEELO
```
FlorianWendelborn, line 10: Hi

FlorianWendelborn:
This interacts very weirdly with `isDisabled`:

image

packages/kotti-ui/source/next/kotti-field-date/mixins.scss
```diff
@@ -226,10 +226,14 @@
 								}
 
 								&:not(.disabled) {
-									@supports (
-										clip-path: circle(calc(#{$td-diameter/2} + 2px) at center)
-									) {
-										clip-path: circle(calc(#{$td-diameter/2} + 2px) at center);
+									&:not(.in-range) {
```
carsoli, line 229: but this was the style for the selected dates (end and start) which have an `.in-range` class 
if we exclude in-range then i exclude the start and end date from having a circle clip path

packages/kotti-ui/source/next/kotti-field-date/styles.scss
```diff
@@ -159,6 +159,38 @@ $year-font-size: 1rem;
...
+			&.end-date {
+				background: linear-gradient(
+					90deg,
+					var(--interactive-04) 0%,
+					var(--interactive-04) 50%,
+					rgba(0, 0, 0, 0) 50%
+				);
+			}
+			&.start-date.end-date {
+				background: transparent !important;
```
carsoli, line 187: please no use of !important
elaborate on what was happening that led u to do that?

packages/kotti-ui/source/next/kotti-field-date/styles.scss
```diff
@@ -159,6 +159,38 @@ $year-font-size: 1rem;
...
+			}
+		}
+		.start-date:hover,
+		.end-date:hover {
+			background-color: transparent !important;
```
carsoli, line 192: please find the part where i add a hover to td:hover and exclude in-range 

please refrain from the !important pattern

packages/kotti-ui/source/next/kotti-field-date/styles.scss
carsoli: so the fix for this, is suggested below (exclude .in-range) from td:hover

and it will automatically include start-date and end-date

if you look at the added classes by element-ui, you'll see why

packages/kotti-ui/source/next/kotti-field-date/styles.scss
```diff
@@ -159,6 +159,38 @@ $year-font-size: 1rem;
 				}
 			}
 		}
+
+		.in-range {
+			background-color: var(--interactive-04);
```
carsoli, line 164: line 164 to 185 should be moved to their concerned part, under td: then add .in-range
the rest should be removed as explained below

thank you

carsoli (changes requested):

Daniel-Beckstein:
@tomquinonero @carsoli @FlorianWendelborn Can we get this merged today until 14:00, if possible, even if some CSS is not perfect, as long as the component works like it should, because this is quite urgent. (We can later improve it in a Hotfix/next release)

tomquinonero:
> This interacts very weirdly with `isDisabled`:
> 
> image

This is fixed by my last commit, this is how it looks now:
Screenshot_2020-09-07 Kotti Design System

packages/kotti-ui/source/next/kotti-field-date/styles.scss
tomquinonero: Not how I want structure my code

Daniel-Beckstein (commented):
I would say lets merge it like it is for now.

packages/kotti-ui/source/next/kotti-field-date/mixins.scss
FlorianWendelborn: Tested this and they’re still round, so I’m resolving.

packages/kotti-ui/source/next/kotti-field-date/styles.scss
FlorianWendelborn: @carsoli moved it, still appears to work.

packages/kotti-ui/source/next/kotti-field-date/styles.scss
FlorianWendelborn: Done, appears to allow removing all `!important`s.

FlorianWendelborn: c5711aec7d881cceccac92a53db555db7d098edd

packages/kotti-ui/source/next/kotti-field-date/styles.scss
FlorianWendelborn: c5711aec7d881cceccac92a53db555db7d098edd

FlorianWendelborn:
Assuming implicit approval from @carsoli as I’ve implemented the PR suggestions. Merging.

PR #276, opened 2020-09-08
Multiline and block buttons

tomquinonero:
This PR tackles two things:
 - A `isBlock` modifier for the button to act like a block element and use all the width available
 - A `isMultiline` modifier so it can handle long text on buttons

The reason I used a modifier instead of having the button just handle it is that using `line-height` is less reliable than `height` for button sizing.
Also the handling of icons needs to be different and we rather "control" which are meant to multiline or not.


FlorianWendelborn:
Boolean flags should be prefixed with `is*`, `hide*`, `show*`, etc.

PR #279, opened 2020-09-09
documentation(installation): Add Hints about @vue/composition-api

FlorianWendelborn:


README.md
```diff
@@ -44,12 +44,16 @@ or
 **Documentation**: [current release](https://3yourmind.github.io/kotti/) or [current master](https://kotti.netlify.com/)
 
 ```typescript
-// in main.js / entrypoint
+// in main.ts / entrypoint
```
carsoli, line 47: arguably we should just say in your entry point, no specification of a ts must or a js must
and maybe no specification of `main` altho it's ofc the default if we spin up a vue project

or what do u think

README.md
```diff
@@ -44,12 +44,16 @@ or
...
 import Vue from 'vue'
 import KottiUI from '@3yourmind/kotti-ui'
 import '@3yourmind/kotti-ui/dist/styles.css'
 
-// register all KtComponents globally
+// also make sure to set-up @vue/composition-api as it’s required for newer Kotti features
+import VueCompositionAPI from '@vue/composition-api' // right now, 0.5.0 is recommended
+Vue.use(VueCompositionAPI)
```
carsoli, line 54: should we stress on plugin order as a comment? 
also is it verified to be an issue?

README.md
FlorianWendelborn: But it says `/ entrypoint`?

README.md
FlorianWendelborn: What do you mean "verified to be an issue"?

carsoli: if we use the composition plugin after using kotti-ui

carsoli: does it actually prove to be a problem

FlorianWendelborn: Ah, that’s what you mean. It’s at least conceptually a problem and I don’t want to rely on the order not mattering in the global magic land that vue is. I can add a comment for that then I guess.

README.md
carsoli: yes im retarded :)

README.md
```diff
@@ -44,12 +44,16 @@ or
...
 ```typescript
-// in main.js / entrypoint
+// in main.ts / entrypoint
 import Vue from 'vue'
 import KottiUI from '@3yourmind/kotti-ui'
 import '@3yourmind/kotti-ui/dist/styles.css'
 
-// register all KtComponents globally
+// also make sure to set-up @vue/composition-api as it’s required for newer Kotti features
+import VueCompositionAPI from '@vue/composition-api' // right now, 0.5.0 is recommended
```
carsoli, line 53: we specify it to be 0.5.x on our peer deps
maybe we shouldn't be that specific and the best we can add is 'pls no use of beta'

carsoli: so even higher patches are supported and i think we already have that in yoda if i'm not mistaken

just saying maybe let's not overkill with specificity but also leave room for a beta-fuckup

carsoli: not an essential change just a suggestion and the last i have 

so i will approve already

FlorianWendelborn: https://github.com/3YOURMIND/kotti/blob/cc115881976692d3d14504ab2430fb1a3a63f8ab/packages/kotti-ui/package.json#L50

PR #282, opened 2020-09-10
Multiline buttons v2

tomquinonero:
This PR just increase specificity of selector to avoid bug due to CSS being loaded in a random order


tomquinonero:
New PR with clean git history: https://github.com/3YOURMIND/kotti/pull/284

PR #284, opened 2020-09-11
fix: up the selector specificity for multiline and block buttons

tomquinonero:
This highen up the specificity of the selector to fix random order CSS.

### DIFF
--- a/packages/kotti-ui/source/kotti-button/KtButton.vue
+++ b/packages/kotti-ui/source/kotti-button/KtButton.vue
@@ -93,12 +93,14 @@ export default {
 		text-transform: none;
 	}
 
-	&--is-block {
+	// increase selector specificity by using .kt-button twice
+	&--is-block.kt-button {
 		display: flex;
 		width: 100%;
 	}
 
-	&--is-multiline {
+	// increase selector specificity by using .kt-button twice
+	&--is-multiline.kt-button {
 		height: auto;
 		padding-top: var(--unit-1);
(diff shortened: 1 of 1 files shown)


packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -218,4 +198,25 @@ export default {
 	min-width: 10rem;
 	margin-top: var(--unit-8);
 }
+
+// block modifier
+.kt-button.kt-button--is-block {
```
FlorianWendelborn, line 203: Can you use `&.&--is-block` and put it back to its original location? AFAIK this should work and is more in-line with the rest of SCSS in this repo.

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -218,4 +198,25 @@ export default {
...
+	display: flex;
+	width: 100%;
+}
+// multiline modifier
+.kt-button.kt-button--is-multiline {
```
FlorianWendelborn, line 208: Same here

FlorianWendelborn (changes requested):

packages/kotti-ui/source/kotti-button/KtButton.vue
tomquinonero: I don;t think there is any benefit to that. It render the same code and I don't find it any more readable or understandable.

PR #299, opened 2020-09-23
documentation(meta): Add Kotti.Meta & Display In Documentation

FlorianWendelborn:


packages/documentation/data/menu.ts
```diff
@@ -39,69 +46,73 @@ export const menu: Array<Section> = [
 				title: 'Foundations',
 				path: 'foundations',
 				pages: [
-					{ label: 'Layout', tags: [], path: 'layout' },
-					{ label: 'Grids', tags: ['ts'], path: 'grids' },
-					{ label: 'Icons', tags: ['css', 'ts'], path: 'icons' },
-					{ label: 'Typography', tags: ['css'], path: 'typography' },
-					{ label: 'Writing Style', tags: [], path: 'writing' },
+					{ label: 'Layout', path: 'layout', tags: [Tag.GUIDE] },
+					{ label: 'Grids', path: 'grids', tags: [Tag.TS] },
```
carsoli, line 50: don't think it needs a ts tag, is not just styles..

carsoli: or what does this lead to

carsoli (approved):

packages/documentation/data/menu.ts
FlorianWendelborn: I’m right now using it as an indicator that the component is documented in the `Kotti` typescript namespace. Also, `KtCol`/`KtRow` are entirely implemented in TS already since we refactored them a couple weeks ago.

Do you think we should somehow clarify this better in the docs? If yes, any ideas how?

Issue #252, opened 2020-08-04
Tree-Shaking: Remove Side-Effects & Adjust Build


FlorianWendelborn:
See #207

---

The CSS part will be handled via #317

PR #301, opened 2020-09-23
build(#252): Support ESM Tree Shaking

FlorianWendelborn:
The difference can be tested via the test app:

```bash
yarn workspace @3yourmind/test-app run build --report
open packages/test-app/dist/report.html
```

This should enable ESM tree shaking and hopefully make importing parts of Kotti feasible.

E.g. re-enabling the `KtFieldDate` in the test-app should cause element-ui to be re-added to the final bundle as dependencies are now resolved as-needed.


Tradias:
Tested with a vue-cli 4 based JS project. Works as intended.

Tradias (approved):

packages/test-app/src/App.vue
```diff
@@ -1,20 +1,26 @@
...
-			/></KtTranslationContext>
-		</KtForm>
-	</div>
+	<KtTranslationContext locale="en-US">
+		<div style="padding: 50px;">
+			<KtAccordion title="Hello Tree Shaking">
+				<KtForm v-model="values">
+					<KtHeading text="Test App" />
+					<KtFieldText formKey="text" label="Text" />
+					<!-- <KtFieldDate formKey="date" label="Date" /> -->
```
carsoli, line 8: @FlorianWendelborn do we remove this comment

packages/kotti-ui/source/index.ts
```diff
@@ -120,76 +120,74 @@ import { KtTranslationContext } from './next/kotti-translation'
...
-	KtToaster,
-	KtTranslationContext,
-	KtUserMenu,
-}
-
 export default {
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
 	install(Vue: any) {
-		for (const component of Object.values(components)) Vue.use(component)
+		for (const component of [
```
carsoli, line 126: noice

FlorianWendelborn: Was worried that this would be considered a side-effect so I inlined it, just to be safe

packages/test-app/src/App.vue
carsoli: no, i think they make sense to keep. in the future even to still test that the tree shaking still works (assuming config changes), uncommenting those lines should be the easiest workflow, you were right about that.

PR #303, opened 2020-09-25
feat(297): use old tooltip style colors; adjust widtht with some 'max…

HansAuger:
Nicify kt field tooltip popover
I took the color from the tooltip style introduced by Junyu (to my knowledge only in use at KtAvatar. see [here](https://3yourmind.github.io/kotti/components/avatars/)

Also added some basic java script magic to determine if popup is shown below icon or above it

### DIFF
--- a/packages/kotti-ui/source/next/kotti-field/KtField.vue
+++ b/packages/kotti-ui/source/next/kotti-field/KtField.vue
@@ -298,6 +298,7 @@ export default defineComponent({
 
 			&__tooltip {
 				position: absolute;
+				z-index: 1; // fix weird layout issues with el-select
 				width: max-content;
 				max-width: 12rem;
 				max-height: 7em;
--- a/packages/kotti-ui/source/next/kotti-field/KtField.vue
+++ b/packages/kotti-ui/source/next/kotti-field/KtField.vue
@@ -21,11 +21,18 @@
 					class="kt-field__header__help-text"
 					:class="iconClasses(['interactive'])"
 				>
-					<div class="kt-field__header__help-text__icon">
+					<div
+						ref="helpTextIcon"
+						class="kt-field__header__help-text__icon"
+						@mouseenter="() => (isTooltipHovered = true)"
+						@mouseleave="() => (isTooltipHovered = false)"
+					>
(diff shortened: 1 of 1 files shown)


packages/kotti-ui/source/next/kotti-field/KtField.vue
```diff
@@ -282,20 +297,27 @@ export default defineComponent({
...
+				z-index: 1; // fix weird layout issues with el-select
+				width: max-content;
+				max-width: 12rem;
+				max-height: 7em;
+				padding: 0.75em 1em;
+				overflow: scroll;
+				color: var(--gray-10);
 				cursor: none;
-				background-color: var(--ui-01);
+				background-color: var(--gray-80);
```
carsoli, line 308: try ui-04, it's the closest to that, and thn we stick to the tokens for the background
check this out if you didn't know about it already
https://3yourmind.github.io/kotti/tokens/colors

carsoli (approved):
totally approve, the change i requested is trivial and up for discussion anyways
 just think it's the convention now to stick to bg tokens vs text tokens etc

and if we dont support gray80 on backgrounds, then design knows this and the darkest they want now is gray60 

hope that makes sense

packages/kotti-ui/source/next/kotti-field/KtField.vue
```diff
@@ -282,20 +297,27 @@ export default defineComponent({
...
-				display: none;
-				padding: 0.25em;
-				color: var(--support-info);
+				z-index: 1; // fix weird layout issues with el-select
+				width: max-content;
+				max-width: 12rem;
+				max-height: 7em;
+				padding: 0.75em 1em;
+				overflow: scroll;
+				color: var(--gray-10);
```
tomquinonero, line 306: Can't we use a token here?

FlorianWendelborn: @tomquinonero which token are you suggesting? Personally, I never know which token belongs to what token as the names seem rather arbitrary to me.

Personally, I’d prefer it if we had something like: `ui-01-background` + `ui-01-foreground` so that it’s clear that they should be used together. With the current tokens, it feels like `gray-10` with extra steps to me but perhaps I just don’t get it

PR #311, opened 2020-10-14
fix(300): counter-act chrome autocomplete attribute disregard

carsoli:
when using the selects with a field that has keywords like "address" in its label/parent wrapper, chrome recognizes the input field as autofillable and shows autofill suggestions as per saved autofill data in chrome://settings
to counter-act this, since `autocomplete=off` doesn't work in recent versions of chrome to disable autofill suggestions, we change the input type to search instead of text, thus excluding the field from the fillable criteria
https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-expectation-mantle

### DIFF
--- a/packages/kotti-ui/source/next/kotti-field-select/hooks.ts
+++ b/packages/kotti-ui/source/next/kotti-field-select/hooks.ts
@@ -131,6 +131,42 @@ const useSelectInputSizeFix = <DATA_TYPE extends Values>({
 	})
 }
 
+/**
+ * @summary HACK: Prevent Google Chrome from autocompleting the search field
+ *
+ * @description
+ * chrome 66+ shows suggestions for saved autofill data,
+ * through intelligent detection, recognizing "address" fields, for instance,
+ * and suggesting autofills.
+ *
+ * using `autocomplete="off" on `input` of `type="text"` would not work to disable autofill
+ *
+ * @see {@link https://gist.github.com/niksumeiko/360164708c3b326bd1c8#gistcomment-2032962} for suggested fixes
+ * @see {@link https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-expectation-mantle} for more information on the autocomplete attribute and accepted values
+ * @see {@link https://bugs.chromium.org/p/chromium/issues/detail?id=370363#c7} for more information on the autocomplete tokens for password fields
+ */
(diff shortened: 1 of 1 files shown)


packages/kotti-ui/source/next/kotti-field-select/hooks.ts
```diff
@@ -131,6 +131,44 @@ const useSelectInputSizeFix = <DATA_TYPE extends Values>({
...
+ * and suggesting autofills.
+ *
+ * using `autocomplete="off" on `input` of `type="search"` would not work to disable autofill
+ *
+ * HACK to exclude the input field from being auto-fillable
+ * @see {@link https://gist.github.com/niksumeiko/360164708c3b326bd1c8#gistcomment-2032962 }
+ *
+ * could potentially break in even newer versions of chrome.
+ * could be solved by accepting an autocomplete prop
+ * and using tokens for commonly saved autofill fields (e.g. address-line-1)
```
FlorianWendelborn, line 146: Would still argue that this is wrong, as we DO have a `search` field. No matter what the browser autocompletes there, it will be wrong.

The only thing `autocomplete="address-line-1"` does is tell the browser to "please auto complete addresses in this field" which we do not want in a single select search field.

packages/kotti-ui/source/next/kotti-field-select/hooks.ts
```diff
@@ -131,6 +131,44 @@ const useSelectInputSizeFix = <DATA_TYPE extends Values>({
...
+ * @see {@link https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-expectation-mantle}
```
FlorianWendelborn, line 147: Love the quote

> The "off" keyword indicates either that the control's input data is particularly sensitive (for example the activation code for a nuclear weapon); or that it is a value that will never be reused (for example a one-time-key for a bank login) and the user will therefore have to explicitly enter the data each time, instead of being able to rely on the UA to prefill the value for them; or that the document provides its own autocomplete mechanism and does not want the user agent to provide autocompletion values.

So Google risks accidentally triggering nukes with ignoring `autocomplete="off"`

carsoli: tagging south korea for ideas about accidental oopsies

PR #313, opened 2020-10-14
Use refs as recommended on kt field

carsoli:
### Issue description
`refs` shouldn't be extracted from SetupContext (it still works, though)
there was a type error that `refs` is undefined on `SetupContext`
recommended method in docs https://composition-api.vuejs.org/api.html#template-refs

### Steps to test
serve docs with `yarn run --cwd packages/documentation serve`
on the `form-fields` page under `Components`
test that the tool-tip classes goes above and below according to the space available above it on `form-fields` page

### DIFF
--- a/packages/kotti-ui/source/next/kotti-field/KtField.vue
+++ b/packages/kotti-ui/source/next/kotti-field/KtField.vue
@@ -94,12 +94,7 @@
 
 <script lang="ts">
 import { Yoco } from '@3yourmind/yoco'
-import {
-	defineComponent,
-	computed,
-	ref,
-	SetupContext,
-} from '@vue/composition-api'
+import { defineComponent, computed, ref } from '@vue/composition-api'
 
 import { useTranslationNamespace } from '../kotti-translation/hooks'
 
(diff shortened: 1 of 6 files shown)


packages/kotti-ui/source/next/kotti-field/KtField.vue
```diff
@@ -126,9 +126,11 @@ export default defineComponent({
 			isGroup: boolean
 			getEmptyValue: () => DATA_TYPE
 		},
-		{ emit, refs }: SetupContext,
+		{ emit },
 	) {
+		const helpTextIconRef = ref<Element>(null)
```
FlorianWendelborn, line 131: Does this actually get correctly detected as `Ref<Element | null>` or is this weird again in Vue?

FlorianWendelborn (changes requested):
LGTM from a CR perspective besides the minor comment. Will run it locally once to verify it still works

FlorianWendelborn (changes requested):
LGTM from a CR perspective besides the minor comment. Will run it locally once to verify it still works

FlorianWendelborn:
WIMM

packages/kotti-ui/source/next/kotti-field/KtField.vue
carsoli: no it doesn't, we established that on initialization of ref in Vue, whatever the type is Vue ignores it xD

I think it's not just for the `null` haha

FlorianWendelborn (approved):

PR #319, opened 2020-10-20
312 change required ui for fields

carsoli:
### Changes:
- removed `requiredLabel` from translation context (breaking change)
- removed requiredLabel from UI (i.e. no longer show `(Required)` next to label of required fields
- show `*` instead ^
- make the asterix go red with validation when a field is required and empty
- add validation icon next to the error msg (Yoco.Icon.CIRCLE_CROSS for error messages, and Yoco.Icon.CIRCLE_CHECK for success messages and Yoco.Icon.CIRCLE_ATTENTION for warning messages)
---
### Functional Testing
1. `git pull --rebase`
2. `yarn run watch` should build and serve the docs
3. go to components/form-fields, change the field switch `isOptional` to be false (i.e. make field required)
4. then if the field is empty, the asterix should be red and there should be a custom error state (not the msg that the field is required)
5. type in the field, the asterix should be grey again and the validation msg gone
6. if you hide validation with the relevant switch, the asterix should still show in grey
7. try to change the validation state to `warning` and `success` as well and see how the icon to the left of the msg shows

### DIFF
--- a/packages/kotti-ui/source/next/kotti-field/KtField.vue
+++ b/packages/kotti-ui/source/next/kotti-field/KtField.vue
@@ -14,7 +14,7 @@
 					class="kt-field__header__label"
 				>
 					<span class="kt-field__header__label__text" v-text="field.label" />
-					<span class="kt-field__header__label__suffix" v-text="labelSuffix" />
+					<span :class="labelSuffixClasses" v-text="labelSuffix" />
 				</component>
 				<div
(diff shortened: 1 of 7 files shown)


HansAuger (approved):

carsoli:
one reviewer was needed (tiny PR)
multiple assigned so whoever could make time would review.

PR #320, opened 2020-10-23
More theming for navbar

tomquinonero:
This PR add:
 - A way of theming the navbar differently, by overwriting the `primary` colors on `.kt-navbar`.
 - Two navbar themes: `light` and `dark`
 - A way of changing navbar theme in the navbar documentation

This PR removes:
 - The `grayscale` theme, that is not used anymore (was used for SP)

To test: 
Run server and go to <http://localhost:3000/examples/layouts#>, or go to <http://localhost:3000/patterns/navbar> and click on "Open Example".
There is a select that allow us to select theme. 
Please ignore the notification area as it wasn't really supported on default theme.

### DIFF
--- a/packages/documentation/pages/examples/layouts.vue
+++ b/packages/documentation/pages/examples/layouts.vue
@@ -26,6 +26,7 @@
 		<div slot="workspace">
 			<KtFieldSingleSelect
 				v-model="theme"
+				hideClear
 				isOptional
 				label="Navbar Theme"
 				:options="[
--- a/packages/documentation/pages/examples/layouts.vue
+++ b/packages/documentation/pages/examples/layouts.vue
@@ -24,7 +24,6 @@
 			:menu="actionbarMenu"
 		/>
 		<div slot="workspace">
-			<div>Hello {{ theme }}</div>
 			<KtForm>
 				<KtFieldSingleSelect
 					v-model="theme"
--- a/packages/documentation/pages/examples/layouts.vue
+++ b/packages/documentation/pages/examples/layouts.vue
@@ -7,6 +7,7 @@
 			:notification="navbarNotification"
 			:quickLinks="quickLinksData"
 			:sections="navbarMenu"
+			:theme="theme"
 		>
 			<div slot="navbar-footer">
 				<KtUserMenu
@@ -22,7 +23,23 @@
 			headerTitle="LayoutContainer Example"
 			:menu="actionbarMenu"
 		/>
-		<div slot="workspace"></div>
+		<div slot="workspace">
+			<div>Hello {{ theme }}</div>
+			<KtForm>
+				<KtFieldSingleSelect
+					v-model="theme"
+					formKey="NONE"
+					isOptional
+					label="Navbar Theme"
+					:options="[
+						{ label: 'Reversed theme', value: 'reverse' },
+						{ label: 'Light theme', value: 'light' },
+						{ label: 'Dark theme', value: 'dark' },
+						{ label: 'None', value: null },
+					]"
+				/>
+			</KtForm>
+		</div>
 	</LayoutContainer>
 </template>
 
@@ -37,6 +54,7 @@ export default {
 	},
 	data() {
 		return {
+			theme: null,
 			quickLinksData: {
 				links: [
 					{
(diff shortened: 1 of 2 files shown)


packages/documentation/pages/examples/layouts.vue
```diff
@@ -22,7 +23,22 @@
 			headerTitle="LayoutContainer Example"
 			:menu="actionbarMenu"
 		/>
-		<div slot="workspace"></div>
+		<div slot="workspace">
+			<KtForm>
```
carsoli, line 27: you can choose to use a field without a form if the formKey is NONE anyways

so no need for `KtForm` and thus `formKey`

packages/documentation/pages/examples/layouts.vue
```diff
@@ -22,7 +23,22 @@
...
+				<KtFieldSingleSelect
+					v-model="theme"
+					formKey="NONE"
+					isOptional
+					label="Navbar Theme"
+					:options="[
+						{ label: 'Reversed theme', value: 'reverse' },
+						{ label: 'Light theme', value: 'light' },
+						{ label: 'Dark theme', value: 'dark' },
+						{ label: 'None', value: null },
```
carsoli, line 37: also if you want a  null option explicity, pass `hideClear` so u don't have a null option AND the option to clear which will also have a vaule null but a label `''`

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
```diff
@@ -252,7 +273,6 @@ $narrow-navbar-width: 3.4rem;
 	.kt-navbar-quick-links {
```
carsoli, line 273: do u think following BEM since we now have `kt-navbar` this can become `kt-navbar__quick-links`
previously we didn't consider probably bc we didn't have it

assuming we don't need to overwrite it in Yoda this should be fine but this will be a breaking change and we can adjust release notes accordingly to handle this on any user app's side 

and then also move within the above class as `&__quicklinks`

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
```diff
@@ -146,17 +149,35 @@ $narrow-navbar-width: 3.4rem;
 		--user-menu-color: var(--primary-80);
 	}
 
-	&--theme-grayscale {
+	&--theme-light {
```
carsoli, line 152: in general, i have a question

do we customize the primary colors to our customers
so if we want to extend themes to fit our customers requirements, do we only need to change what primary-* values are

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
carsoli: i went through the file and it seems the entire BEM convention is absolutely weird for this file, subcomponents seem to be separated with `-` in many cases

you can skip this comment, and i'll make time for a separate PR when u merge yours

sorry for the hassle

packages/documentation/pages/examples/layouts.vue
tomquinonero: So no `formKey` on field and no `Form` ? Cool!

packages/documentation/pages/examples/layouts.vue
tomquinonero: I can't name it "None" then right?? I think empty label is a bit confusing?

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
tomquinonero: yes we do

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
tomquinonero: Yes this is old and not really using BEM. 
Ok for separate PR :+1:

packages/documentation/pages/examples/layouts.vue
carsoli: No you can keep the name, I'm suggesting you add the `hideClear` flag on the component because you have a null option but also you can clear so you can two possible ways of having a null value and i don't think you need that given that you have explicitly a null option

FlorianWendelborn (approved):
LGTM

carsoli (approved):

PR #326, opened 2020-11-10
Navbar adjustment

tomquinonero:
Small adjustments for theming :)


carsoli (approved):

FlorianWendelborn (approved):

carsoli:
@tomquinonero we both approve but please don't merge if you will need to publish this ASAP

we would rather publish from a separate branch cause master has merged changes that are un-publishable

PR #327, opened 2020-11-12
Fork Vue Tippy

FlorianWendelborn:


packages/vue-use-tippy/README.md
```diff
@@ -0,0 +1,29 @@
...
+```typescript
+import { useTippy } from '@3yourmind/vue-use-tippy'
+import { defineComponent, ref } from '@vue/composition-api'
+
+export default defineComponent({
+	setup() {
+		const contentRef = ref(null)
+
+		useTippy(helpTextTriggerRef, {
+			appendTo: () => document.body,
```
carsoli, line 16: do u think we should just link the props documentation of tippy

packages/vue-use-tippy/source/index.ts
```diff
@@ -0,0 +1,89 @@
...
+) => {
+	const instance = ref<Instance<Props>[] | null>(null)
+
+	const onMountCallbacks: Callback[] = []
+	onMounted(() => {
+		if (isRef(options.content)) {
+			watch(
+				options.content,
+				(newValue) => {
+					options.content = newValue // NOOP?
```
carsoli, line 39: @KABBOUCHI  hey :) so we forked `vue-tippy/composition/useTippy.js` and we are wondering about the purpose of that particular line (https://github.com/KABBOUCHI/vue-tippy/blob/4655fc1a6162cc0338c35f2e16251b10d5b7070f/composition/useTippy.js#L52)

is it intentionally there?

carsoli (approved):

packages/vue-use-tippy/source/index.ts
KABBOUCHI: incase `content` is `ref` or `computed`, `content` will be updated automatically

`computed` example:
https://github.com/KABBOUCHI/vue-tippy/blob/4c8d3ef930a0078493446583394edca6b5bbe909/playground/App.vue#L181-L195

carsoli: @KABBOUCHI but that answers for why the watcher exists - I'm confused about the particular line? like do you intentionally want to retrigger the watcher?

carsoli: @KABBOUCHI but that answers for why the watcher exists. 

I'm confused about why you're updating `options.content` in line `39`, even though it's been already _updated_  (..thus, us being in its watcher) 

are you intentionally triggering the *deep* watcher on `options`  [lines 72-80](https://github.com/3YOURMIND/kotti/pull/327/files/f84716164497c0630c06cbd00edffe7221ad44b8#diff-3fdf5100ebb81a1de86a715eb0f35a1277fbe8ebd4a2119f4ed2cbb46a82859dR72-R80)?

KABBOUCHI: > I'm confused about why you're updating options.content in line 39, even though it's been already updated (..thus, us being in its watcher)

This only work if `options.content` is `ref` the other watcher enabled if `options` is `ref` or `reactive`, 

```js
// only content is reactive

const counter = ref(0)

let options = {
  content : counter,
  ...
}
```
vs

```js
// whole object is reactive (props)
let options = reactive({
  content : 0,
  ...
})
```

FlorianWendelborn: @KABBOUCHI Thanks for your input, we think we’re fine with removing the line in question

Issue #334, opened 2020-12-22
KtFieldDate* popper initially misplaced


carsoli:
**Describe the bug**
the date-picker was misplaced in contexts where the date *field's* direct parent width was < width of the popper

**To Reproduce**
1) Open a date range picker in a window resolution where the parent of the field has a width < `700px`
2) Initially, the popper is placed BUT 
a) if you scroll, it automatically properly gets placed
b) if you close it and open it, it properly gets placed

**Expected behavior**
At any point in time, opening the picker should trigger the right placement

**Screenshots**
![Screenshot from 2020-12-22 20-41-15](https://user-images.githubusercontent.com/17387510/102922218-59f5db80-4496-11eb-8220-7fb2f88f2c38.png)

**Desktop (please complete the following information):**
 - OS: irrelevant
 - Browser: irrelevant
 - 

**Additional context**
- it was only reproducible on initial render bc we deeply mutated the popper width directly, which on initial render waas done AFTER the position of the popper was computed (taking into account its width)

- on following cycles, the popper's width had already been changed so the computation resolves correctly

PR #335, opened 2020-12-22
hack(#334): force trigger popperjs.update to recompute boundaries

carsoli:


packages/kotti-ui/source/next/kotti-field-date/hooks.ts
```diff
@@ -161,6 +163,20 @@ const usePickerDimensionsFix = <DATA_TYPE extends Values>({
...
+				 * which calls `updatePopper` from the mixin vue-popper
+				 * https://github.com/ElemeFE/element/blob/649670c55a45c7343eb7148565e2d873bc3d52dd/src/utils/vue-popper.js#L120
+				 * which calls the prototype-defined function `update` from popper.js
+				 * https://github.com/ElemeFE/element/blob/3ceec7aa6a7ab46e61dbafa4c68b77ba09384b40/src/utils/popper.js#L224
+				 *
+				 * HACK: to force re-compute the boundaries of the popper element (picker) (in `update` function)
+				 * after we set its width and height, since this is computed in the constructor of the popper, and
+				 * at the instance of initialization, we don't have the new dimensions set, thus resulting in
+				 * a placement issue.
+				 */
```
FlorianWendelborn, line 178: I'd say it'd be good to add a link to the JIRA ticket or (even better, maybe?) a description of how to reproduce the placement issue

Otherwise, future us might find this and have no clue if the bug was already fixed somewhere else

FlorianWendelborn (approved):
LGTM, didn't test it but it makes sense that this could fix it

I'm assuming you did the FT. In that case, the only suggestion I have is describing the placement issue you mentioned

packages/kotti-ui/source/next/kotti-field-date/hooks.ts
carsoli: it's reproducible on the docs though, regardless of the JIRA issue or do you want that to see a video? bc I can also upload that on this issue

FlorianWendelborn: No, my point is that right now we know what issue this is, but in the future we or someone else might want to refactor this and they'd want to make sure they're not accidentally re-introducing the bug

So, a comoment that basically contains the requirements & symptoms of the issue (or, alternatively — but harder — a unit test) would be amazing for someone in that situation

Right now, I sorta know what this is about, but that won't be the case in the future

FlorianWendelborn:
FT on Safari and Firefox Developer Edition worked as expected. It’s a lot less hacky than I’d have expected this solution to be. Good job

Will merge it now. Feel free to publish a new alpha

Issue #337, opened 2021-01-13
KtFieldNumber doesn't maintain cursor position


carsoli:
**Describe the bug**
KtFieldNumber parses the input with each key press, thus moving the cursor to the end, always

**To Reproduce**
On any `KtFieldNumber`, try typing any number (e.g. `12`) then let's say you want to make that `1462` so you move your cursor between the 1 & 2 and you type `4` - the cursor automatically moves to the end so by the time you type `6` it becomes `1426` instead of `1462`

**Expected behavior**
you can type multiple characters to where you left your cursor, last

carsoli:
Reported by @Daniel-Beckstein 
would hopefully be tackled ASAP by him and myself

PR #339, opened 2021-01-14
fix(#337): fix cursor position on KtFieldNumber

carsoli:
track cursor position before feeding the value back to the input field
via native DOM attributes/functions
setting the cursor position back once we settle on what the value is

Co-authored-by: Danie Beckstein <53575247+Daniel-Beckstein@users.noreply.github.com>


FlorianWendelborn (approved):
LGTM with our changes.

PR #340, opened 2021-01-20
feature(KtFieldToggleGroup): Support tooltip on options

FlorianWendelborn:
image

Mostly used to improve the tooltips in `/service-panel/materials/1/price`:

image

### DIFF
--- a/packages/kotti-ui/source/next/kotti-field-toggle/components/ToggleInner.vue
+++ b/packages/kotti-ui/source/next/kotti-field-toggle/components/ToggleInner.vue
@@ -66,5 +66,9 @@ export default defineComponent({
 		color: var(--text-05);
 		cursor: not-allowed;
 	}
+
+	> *:not(:first-child) {
+		margin-left: 0.3rem;
+	}
 }
 </style>
--- a/packages/kotti-ui/source/next/kotti-field-toggle/types.ts
+++ b/packages/kotti-ui/source/next/kotti-field-toggle/types.ts
@@ -10,9 +10,10 @@ export namespace KottiFieldToggle {
 
 export namespace KottiFieldToggleGroup {
 	export type Entry = {
-		disabled?: boolean
+		isDisabled?: boolean
 		key: keyof Value
 		label: string
+		tooltip?: string
 	}
 
 	export type Props = KottiField.Props<Value> & {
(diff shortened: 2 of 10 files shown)


packages/kotti-ui/source/next/kotti-field-toggle/types.ts
```diff
@@ -10,9 +10,10 @@ export namespace KottiFieldToggle {
 
 export namespace KottiFieldToggleGroup {
 	export type Entry = {
-		disabled?: boolean
+		isDisabled?: boolean
```
FlorianWendelborn, line 13: Breaking change, but still unused in yoda.

carsoli: @FlorianWendelborn please add to release notes regardless of course

packages/kotti-ui/source/next/kotti-field-toggle/components/ToggleInner.vue
```diff
@@ -66,5 +66,9 @@ export default defineComponent({
 		color: var(--text-05);
 		cursor: not-allowed;
 	}
+
+	> *:not(:first-child) {
```
carsoli, line 70: what's that about

packages/kotti-ui/source/next/kotti-field-toggle/types.ts
carsoli: can u extend this to selects as well

we unify all options, supposedly

so the options on `KtFieldMultiSelect` and `KtFieldSingleSelect`

FlorianWendelborn: https://github.com/3YOURMIND/kotti/releases/edit/untagged-1a4388e0038e68afaefc

packages/kotti-ui/source/next/kotti-field-toggle/components/ToggleInner.vue
FlorianWendelborn: I moved it from the `ToggleBox.vue` and `ToggleSwitch.vue` components and made it 25% smaller

Thought it’s better that way so that we can also style the tooltip with the same rule and avoid the duplication of the two files both having the same margin-right

PR #341, opened 2021-01-20
feature(KtFieldRadioGroup): Add tooltip option on options

FlorianWendelborn:
Second part of #340


packages/kotti-ui/source/next/kotti-field-radio-group/KtFieldRadioGroup.vue
```diff
@@ -19,13 +19,6 @@
 						field.isDisabled || Boolean(option.isDisabled),
 				}"
 			>
-				<input
```
FlorianWendelborn, line 22: Re-ordered to fix issues with `:not(:first-child)` adding margin to the first visible child

carsoli (approved):

PR #363, opened 2021-03-01
feature(#366): Implement Options in KtPopover

RueRivoli:
Created component KtPopoverIconTextItem to handle buttons displayed as icons + text and permits disabled class on them.
Created fallback case in KtPopover that does a loop on prop "options" to display KtPopoverIconTextItems.
Added validator function in KtPopover to control prop options.

### DIFF
--- a/packages/kotti-ui/source/index.ts
+++ b/packages/kotti-ui/source/index.ts
@@ -76,7 +76,11 @@ import { KtNavbar } from './kotti-navbar'
 export * from './kotti-navbar'
 import { KtPagination } from './kotti-pagination'
 export * from './kotti-pagination'
-import { KtPopover, KtPopoverItem } from './kotti-popover'
+import {
+	KtPopover,
+	KtPopoverItem,
+	KtPopoverIconTextItem,
+} from './kotti-popover'
 export * from './kotti-popover'
 import { KtRadio } from './kotti-radio'
 export * from './kotti-radio'
@@ -153,6 +157,7 @@ export default {
 			KtPagination,
 			KtPopover,
 			KtPopoverItem,
+			KtPopoverIconTextItem,
 			KtRadio,
 			KtRadioGroup,
 			KtRow,
--- a/packages/kotti-ui/source/index.ts
+++ b/packages/kotti-ui/source/index.ts
@@ -76,11 +76,7 @@ import { KtNavbar } from './kotti-navbar'
 export * from './kotti-navbar'
 import { KtPagination } from './kotti-pagination'
 export * from './kotti-pagination'
-import {
-	KtPopover,
-	KtPopoverItem,
-	KtPopoverIconTextItem,
-} from './kotti-popover'
+import { KtPopover, KtPopoverItem } from './kotti-popover'
 export * from './kotti-popover'
 import { KtRadio } from './kotti-radio'
 export * from './kotti-radio'
@@ -157,7 +153,6 @@ export default {
 			KtPagination,
 			KtPopover,
 			KtPopoverItem,
-			KtPopoverIconTextItem,
 			KtRadio,
 			KtRadioGroup,
 			KtRow,
--- a/packages/kotti-ui/source/kotti-popover/KtPopover.vue
+++ b/packages/kotti-ui/source/kotti-popover/KtPopover.vue
@@ -26,6 +26,8 @@
 import { createPopper } from '@popperjs/core'
 import { mixin as clickaway } from 'vue-clickaway'
 
+import KtPopoverIconTextItem from './KtPopoverIconTextItem.vue'
+
 const optionIsValid = (option) =>
 	(!option.isDisabled || typeof option.isDisabled === 'boolean') &&
 	option.onClick &&
@@ -35,6 +37,7 @@ const optionIsValid = (option) =>
 
 export default {
 	name: 'KtPopover',
+	components: { KtPopoverIconTextItem },
 	mixins: [clickaway],
 	props: {
 		content: { default: '', type: String },
--- a/packages/kotti-ui/source/kotti-popover/KtPopover.vue
+++ b/packages/kotti-ui/source/kotti-popover/KtPopover.vue
@@ -13,8 +13,8 @@
 					v-for="(option, index) in options"
 					:key="index"
 					:icon="option.icon"
-					:isDisabled="option.disabled"
-					:text="option.text"
+					:isDisabled="option.isDisabled"
+					:label="option.label"
 					@click="clickHandler(option)"
 				/>
 			</slot>
@@ -27,11 +27,11 @@ import { createPopper } from '@popperjs/core'
 import { mixin as clickaway } from 'vue-clickaway'
 
 const optionIsValid = (option) =>
-	(!option.disabled || typeof option.disabled === 'boolean') &&
+	(!option.isDisabled || typeof option.isDisabled === 'boolean') &&
 	option.onClick &&
 	typeof option.onClick === 'function' &&
-	option.text &&
-	typeof option.text === 'string'
+	option.label &&
+	typeof option.label === 'string'
 
 export default {
 	name: 'KtPopover',
@@ -95,7 +95,7 @@ export default {
 	methods: {
 		clickHandler(option) {
 			if (
-				!option.disabled &&
+				!option.isDisabled &&
 				option.onClick &&
 				typeof option.onClick === 'function'
 			)
--- a/packages/kotti-ui/source/kotti-popover/KtPopover.vue
+++ b/packages/kotti-ui/source/kotti-popover/KtPopover.vue
@@ -4,7 +4,7 @@
 		class="kt-popover"
 		:class="{ showPopper }"
 	>
-		<div ref="anchor" @click="handleClick">
+		<div ref="anchor" @click="handleAnchorClick">
 			<slot>Anchor</slot>
 		</div>
 		<div v-if="showPopper" ref="content" :class="popperClass">
@@ -15,7 +15,7 @@
 					:icon="option.icon"
 					:isDisabled="option.isDisabled"
 					:label="option.label"
-					@click="clickHandler(option)"
+					@click="handleItemClick(option)"
 				/>
 			</slot>
 		</div>
@@ -26,14 +26,17 @@
 import { createPopper } from '@popperjs/core'
 import { mixin as clickaway } from 'vue-clickaway'
 
+import { isYocoIcon } from '../validators'
+
 import KtPopoverIconTextItem from './KtPopoverIconTextItem.vue'
 
 const optionIsValid = (option) =>
-	(!option.isDisabled || typeof option.isDisabled === 'boolean') &&
-	option.onClick &&
-	typeof option.onClick === 'function' &&
-	option.label &&
-	typeof option.label === 'string'
+	typeof option === 'object' &&
+	option !== null &&
+	(typeof option.icon === 'undefined' || isYocoIcon(option.icon)) &&
+	['undefined', 'boolean'].includes(typeof option.isDisabled) &&
+	['undefined', 'function'].includes(typeof option.onClick) &&
+	['undefined', 'string'].includes(typeof option.label)
 
 export default {
 	name: 'KtPopover',
@@ -96,15 +99,10 @@ export default {
 		this.destroyPopper()
 	},
 	methods: {
-		clickHandler(option) {
-			if (
-				!option.isDisabled &&
-				option.onClick &&
-				typeof option.onClick === 'function'
-			)
-				option.onClick()
+		handleItemClick(option) {
+			if (!option.isDisabled && option.onClick) option.onClick()
 		},
-		handleClick() {
+		handleAnchorClick() {
 			if (!this.forceShowPopoverIsNull) return
 			this.showPopper = !this.showPopper
 		},
@@ -169,7 +167,7 @@ export default {
 	z-index: $zindex-4;
 	width: auto;
 	padding: 0.8rem;
-	background: #fff;
+	background: var(--white);
 	border-radius: $border-radius;
 	box-shadow: $box-shadow;
 
--- a/packages/kotti-ui/source/kotti-popover/KtPopover.vue
+++ b/packages/kotti-ui/source/kotti-popover/KtPopover.vue
@@ -9,7 +9,14 @@
 		</div>
 		<div v-if="showPopper" ref="content" :class="popperClass">
 			<slot :close="handleClickaway" name="content">
-				{{ content }}
+				<KtPopoverIconTextItem
+					v-for="(option, index) in options"
+					:key="index"
+					:icon="option.icon"
+					:isDisabled="option.disabled"
+					:text="option.text"
+					@click="clickHandler(option)"
+				/>
 			</slot>
 		</div>
 	</div>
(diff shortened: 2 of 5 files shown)


packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+	<div
+		v-bind="$attrs"
+		class="kt-popover-item"
+		:class="{
+			'kt-popover-item--disabled': isDisabled,
```
FlorianWendelborn, line 6: Should be `--is-disabled`

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+	<div
+		v-bind="$attrs"
```
FlorianWendelborn, line 3: Is this actually necessary? If yes, why? This should be the default behavior

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
```diff
@@ -0,0 +1,55 @@
...
+		class="kt-popover-item"
+		:class="{
+			'kt-popover-item--disabled': isDisabled,
+		}"
+		v-on="$listeners"
```
FlorianWendelborn, line 8: Same question

packages/kotti-ui/source/kotti-popover/KtPopover.vue
```diff
@@ -75,6 +93,14 @@ export default {
 		this.destroyPopper()
 	},
 	methods: {
+		clickHandler(option) {
+			if (
+				!option.disabled &&
```
FlorianWendelborn, line 98: Should be `.isDisabled` if this is a new API

packages/kotti-ui/source/kotti-popover/KtPopover.vue
```diff
@@ -19,13 +26,24 @@
 import { createPopper } from '@popperjs/core'
 import { mixin as clickaway } from 'vue-clickaway'
 
+const optionIsValid = (option) =>
+	(!option.disabled || typeof option.disabled === 'boolean') &&
```
FlorianWendelborn, line 30: `isDisabled`

packages/kotti-ui/source/kotti-popover/KtPopover.vue
```diff
@@ -19,13 +26,24 @@
...
+	option.onClick &&
+	typeof option.onClick === 'function' &&
+	option.text &&
```
FlorianWendelborn, line 33: If this is a new API, this should be `label` to align with `KtField*Select`

FlorianWendelborn (changes requested):

packages/kotti-ui/source/kotti-popover/KtPopover.vue
RueRivoli: ok correct

packages/kotti-ui/source/kotti-popover/KtPopover.vue
RueRivoli: ok

packages/kotti-ui/source/kotti-popover/KtPopover.vue
RueRivoli: ok

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
RueRivoli: v-on="$listeners" is used to catch the click event inside ktPopover

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
RueRivoli: no not necessary

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
RueRivoli: yes

packages/kotti-ui/source/kotti-popover/index.ts
```diff
@@ -21,3 +22,15 @@ export const KtPopoverItem = attachMeta(makeInstallable(KtPopoverItemVue), {
 	},
 	typeScript: null,
 })
+
+export const KtPopoverIconTextItem = attachMeta(
+	makeInstallable(KtPopoverIconTextItemVue),
+	{
+		addedVersion: '0.0.8',
```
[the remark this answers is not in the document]
RueRivoli: ok

packages/kotti-ui/source/index.ts
```diff
@@ -155,6 +159,7 @@ export default {
 			KtPagination,
 			KtPopover,
 			KtPopoverItem,
+			KtPopoverIconTextItem,
```
mvangent, line 162: @RueRivoli @FlorianWendelborn @carsoli  Since it is now implemented as an internal fallback, are you sure you want to expose it on the external api? Other than that code is looking good to me at this stage.

RueRivoli: Yes, it would make sense. But if I remove it from the export in kotti-ui/source/index.ts, the KtPopoverIconTextItem s disappear in the example of the fallback case. Just an empty popover is displayed. That's why I kept it like this.

RueRivoli: Yes, so import it in KtPopover directly is better

mvangent (approved):

packages/kotti-ui/source/kotti-popover/KtPopover.vue
```diff
@@ -19,13 +26,27 @@
 import { createPopper } from '@popperjs/core'
 import { mixin as clickaway } from 'vue-clickaway'
 
+import KtPopoverIconTextItem from './KtPopoverIconTextItem.vue'
+
+const optionIsValid = (option) =>
```
carsoli, line 31: so, let's tackle this line for line

what if your `option` is a `null` - what if you're given an array of `null` or `undefined` - this line for starters will fail
so you need to above anything guard against `option` being `undefined`/ `null`, 
then what if your option has a `string` type, for example .. well you also need to guard then that this option is an object, no?
Your first check should be
> `typeof option === 'object'` && object !== null`
the `object` `typeof` check will guard against all wrong types and the the second part is immediately augemented because `null` passes as object

packages/kotti-ui/source/kotti-popover/KtPopover.vue
```diff
@@ -19,13 +26,27 @@
...
+	(!option.isDisabled || typeof option.isDisabled === 'boolean') &&
```
carsoli, line 32: what if you get `option: { isDisabled: '' , ... }` 
do you want to allow this?
you want to be more strict because `!` will negate any primitive (so empty string will pass, number zero will pass

so what do we actually want? you either want to not get `isDisabled` prop at all on your `option` object or you want to get it as a boolean, you don't want API mis-use and sources of bugs.
so your check should be as strict as `isDisabled` is undefined or `isDisabled` has a `boolean` type

just a note: 
```typeof undefined === 'undefined``` so you can run the typeof check and see if it's either this or that

File: packages/kotti-ui/source/kotti-popover/KtPopover.vue [hunk 11]
```diff
@@ -19,13 +26,27 @@
...
+	option.onClick &&
```
carsoli, line 33: same as the `isDisabled` you don't want to force the user to pass an `onClick` 
if the user doesn't want the onClick option, then his item is not clickable, easy

my suggestion would be similiar then
either typeof `onClick` is `undefined` (so not passed)
or is a `function`

packages/kotti-ui/source/kotti-popover/KtPopover.vue
```diff
@@ -19,13 +26,27 @@
...
+	typeof option.onClick === 'function' &&
+	option.label &&
```
carsoli, line 35: this check means you disregard an empty string label
your API shouldn't prevent the user from having an empty string for a label if they want to
so this check is unneeded

usually for primitives, you just check that you get the type exactly of that primitive

packages/kotti-ui/source/kotti-popover/KtPopover.vue
```diff
@@ -75,6 +96,14 @@ export default {
 		this.destroyPopper()
 	},
 	methods: {
+		clickHandler(option) {
+			if (
+				!option.isDisabled &&
+				option.onClick &&
```
carsoli, line 102: you'll only need this check bc your validator guards against the type of `onClick` if it exists

packages/kotti-ui/source/kotti-popover/KtPopover.vue
```diff
@@ -9,7 +9,14 @@
...
 		<div v-if="showPopper" ref="content" :class="popperClass">
 			<slot :close="handleClickaway" name="content">
-				{{ content }}
+				<KtPopoverIconTextItem
+					v-for="(option, index) in options"
+					:key="index"
+					:icon="option.icon"
+					:isDisabled="option.isDisabled"
+					:label="option.label"
+					@click="clickHandler(option)"
```
carsoli, line 18: i noticed you can't call this `handleClick` b/c there is another `handleClick`
there's also a `handleClickaway`
I'd say stick to the naming convention of this file at least
which names all handlers to be prefixed with `handle`

so i propose u cann this `handleItemClick`

while you rename `handleClick` to `handleAnchorClick` maybe

packages/kotti-ui/source/kotti-popover/KtPopover.vue
carsoli, line 31: i noticed you want to support `icon` on the `option` as well but you don't check for it in the validator

check if an icon is passed (not `undefined`) and if it is, there's a util function `isYocoIcon` that you can run immediately

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+	<div
+		class="kt-popover-item"
+		:class="{
+			'kt-popover-item--is-disabled': isDisabled,
+		}"
+		v-on="$listeners"
+	>
+		<i v-show="icon" class="yoco" v-text="icon" />
```
carsoli, line 9: why `show` and not `if`
why `render` at all if `icon` is `null` (default)

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
```diff
@@ -0,0 +1,54 @@
...
+			type: Boolean,
+		},
+		label: {
+			type: String,
+		},
+	},
+}
+</script>
+
+<style lang="scss">
```
carsoli, line 32: `scoped` please

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
```diff
@@ -0,0 +1,54 @@
...
+
+<script>
+export default {
+	name: 'KtPopoverIconTextItem',
+	props: {
+		icon: {
+			type: String,
+			default: null,
+		},
+		isDisabled: {
```
carsoli, line 22: so your rule of thumb for props could be that you either support the combo of [`type` and `default`] or [`type` and `required`]
so add a fallback `default` since you can't make this `required`

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
```diff
@@ -0,0 +1,54 @@
...
+			type: Boolean,
+		},
+		label: {
+			type: String,
```
carsoli, line 26: same but you want to fallback to `null` here, not `''` b/c like I said, the user app might want to have `''` for whatever reason
and maybe even v-if on https://github.com/3YOURMIND/kotti/pull/363/files#diff-dbaf4ef6086fcffffa4d4c846ad5bc90cf8fc93db9a38dbc8f299f197c498ec0R10 that it's not `null`

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
```diff
@@ -0,0 +1,54 @@
...
+<style lang="scss">
+@import '../kotti-style/_variables.scss';
+.kt-popover-item {
+	display: flex;
+	padding: $unit-2;
+	margin: $unit-h;
+	line-height: 20px;
+	&:hover {
+		cursor: pointer;
+		background: $lightgray-300;
```
carsoli, line 41: for the colors specially, we r moving towards using css variables now 
see https://3yourmind.github.io/kotti/tokens/introduction and check how it's used in other places of `kotti-ui` for references

I'd propose you contact the designers for this b/c i'm always confused as to what should be the color for  certain things

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
```diff
@@ -0,0 +1,54 @@
...
+		border-radius: $border-radius;
+	}
+	.yoco {
+		align-self: center;
+		padding-right: $unit-1;
+		font-size: 20px;
+	}
+	&--is-disabled {
+		color: #ccc;
+		cursor: auto;
```
carsoli, line 51: i'd propose `cursor: not allowed`

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
```diff
@@ -0,0 +1,54 @@
...
+		background: $lightgray-300;
+		border-radius: $border-radius;
+	}
+	.yoco {
+		align-self: center;
+		padding-right: $unit-1;
+		font-size: 20px;
+	}
+	&--is-disabled {
+		color: #ccc;
```
carsoli, line 50: stick with the css variables as well
contact the designers if need be, but no hardcoded colors

breaks theming for one

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
carsoli: https://3yourmind.github.io/kotti/tokens/colors see this also

packages/kotti-ui/source/kotti-popover/index.ts
```diff
@@ -21,3 +22,13 @@ export const KtPopoverItem = attachMeta(makeInstallable(KtPopoverItemVue), {
 	},
 	typeScript: null,
 })
+
+export const KtPopoverIconTextItem = attachMeta(
+	makeInstallable(KtPopoverIconTextItemVue),
+	{
+		addedVersion: '3.0.0',
```
carsoli, line 29: that's not the version you're about to support this on .. 
it's going to be `3.0.0-alpha.19` if nobody publishes before we merge this but double check the last released version and adjust to it

packages/kotti-ui/source/kotti-popover/KtPopover.vue
FlorianWendelborn: @carsoli why would we ever want an item that’s not disabled but not clickable? I’m sorta fine with not necessarily having an `onClick` when it’s disabled though

packages/kotti-ui/source/kotti-popover/KtPopover.vue
FlorianWendelborn: @carsoli I think it’s fine to check for a missing label. Why would we want an option in this dropdown that doesn’t have any content? The user will be confused and I see no way that this would ever be anything but a mistake

packages/kotti-ui/source/kotti-popover/KtPopoverIconTextItem.vue
FlorianWendelborn: And for future reference, unless you have a **very strong reason**, please **always** `scope`

packages/kotti-ui/source/kotti-popover/index.ts
FlorianWendelborn: Please use `3.0.0` though

FlorianWendelborn:
@RueRivoli where and how can we test this? I don’t see any documentation changes

packages/kotti-ui/source/kotti-popover/KtPopover.vue
FlorianWendelborn: @carsoli clarified that this is due to also supporting icons, so we may want icon-only entries. Please disregard my comment.

PR #383, opened 2021-03-25
Update single select to contain possibility of a create option

RueRivoli:
### Aim
update KtFieldSingleSelect to have a create option

### Expected visual:
https://www.figma.com/file/2fayetfljWKphvid6wB2Dy/New-Kotti-Component-%3A-Form-%2F-Select-%2F-Open-with-Link?node-id=1%3A217

### Expected behaviour
When a prop ```action``` is passed to KtFieldSingleSelect, a separator is added at the end of the select and a link is placed under the separator with a label.

### DIFF
--- a/packages/documentation/pages/usage/components/form-fields.vue
+++ b/packages/documentation/pages/usage/components/form-fields.vue
@@ -111,6 +111,14 @@
 				/>
 				<KtFormControllerObject formKey="additionalProps" v-if="componentDefinition.additionalProps.length > 0">
 					<h4>Additional Props</h4>
+					<KtFieldToggle
+						v-if="componentDefinition.additionalProps.includes('actions')"
+						formKey="actions"
+						isOptional
+						helpText="List of actions that can be triggered from the end of the dropdown"
+						label="actions"
+						type="switch"
+					/>
 					<KtFieldSingleSelect
 						v-if="componentDefinition.additionalProps.includes('toggleType')"
 						formKey="toggleType"
@@ -388,7 +396,7 @@ const components: Array<{
 		supports: { clear: false, decoration: true, tabIndex: true },
 	},
 	{
-		additionalProps: ['collapseTagsAfter'],
+		additionalProps: ['collapseTagsAfter', 'actions'],
 		formKey: 'multiSelectValue',
 		name: 'KtFieldMultiSelect',
 		supports: { clear: true, decoration: true, tabIndex: false },
@@ -406,7 +414,7 @@ const components: Array<{
 		supports: { clear: false, decoration: false, tabIndex: true },
 	},
 	{
-		additionalProps: [],
+		additionalProps: ['actions'],
 		formKey: 'singleSelectValue',
 		name: 'KtFieldSingleSelect',
 		supports: { clear: true, decoration: true, tabIndex: false },
@@ -520,20 +528,20 @@ const generateCode = (component: ComponentValue) =>
 const radioGroupOptions: Kotti.FieldRadioGroup.Props['options'] = [
 	{ label: 'Key 1', value: 'value1' },
 	{ label: 'Key 2', value: 'value2', tooltip: 'Some tooltip' },
-	{ label: 'Key 3', value: 'value3' },
 	{
 		isDisabled: true,
-		label: 'Key 4',
+		label: 'Key 3',
 		tooltip: 'This option is disabled',
-		value: 'value4',
+		value: 'value3',
 	},
+	{ label: 'Key 4', value: 'value4' },
 ]
 
 const singleOrMultiSelectOptions: Kotti.FieldSingleSelect.Props['options'] = [
 	{ label: 'Key 1', value: 'value1' },
 	{ label: 'Key 2', value: 'value2' },
-	{ label: 'Key 3', value: 'value3' },
-	{ isDisabled: true, label: 'Key 4', value: 'value4' },
+	{ isDisabled: true, label: 'Key 3', value: 'value3' },
+	{ label: 'Key 4', value: 'value4' },
 ]
 
 const toggleGroupOptions: Kotti.FieldToggleGroup.Props['options'] = [
@@ -562,6 +570,7 @@ export default defineComponent({
 
 		const settings = ref<{
 			additionalProps: {
+				actions: Kotti.FieldToggle.Value
 				autoComplete: 'current-password' | 'new-password'
 				collapseTagsAfter: Kotti.FieldNumber.Value
 				hideChangeButtons: boolean
@@ -596,6 +605,7 @@ export default defineComponent({
 			validation: Kotti.Field.Validation.Result['type']
 		}>({
 			additionalProps: {
+				actions: false,
 				autoComplete: 'current-password',
 				collapseTagsAfter: null,
 				hideChangeButtons: false,
@@ -738,10 +748,21 @@ export default defineComponent({
 					hideChangeButtons: settings.value.additionalProps.hideChangeButtons,
 				})
 
-			if (['KtFieldMultiSelect', 'KtFieldSingleSelect'].includes(component))
+			if (['KtFieldMultiSelect', 'KtFieldSingleSelect'].includes(component)) {
 				Object.assign(additionalProps, {
 					options: singleOrMultiSelectOptions,
 				})
+
+				if (settings.value.additionalProps.actions)
+					Object.assign(additionalProps, {
+						actions: [
+							{
+								label: 'Create Item',
+								onClick: () => alert('actions[0].onClick called'),
+							},
+						],
+					})
+			}
 
 			if (['KtFieldRadioGroup'].includes(component))
 				Object.assign(additionalProps, {
@@ -795,7 +816,6 @@ export default defineComponent({
 				validation: settings.value.validation,
 			}),
 		)
-
 		return {
 			component: computed(
 				(): { meta: Kotti.Meta; name: string } =>
--- a/packages/kotti-ui/source/kotti-field-select/KtFieldMultiSelect.vue
+++ b/packages/kotti-ui/source/kotti-field-select/KtFieldMultiSelect.vue
@@ -44,6 +44,17 @@
 					:label="option.label"
 					:value="option.value"
 				/>
+				<div v-if="actions.length" class="kt-field-select__actions">
+					<div class="el-select-dropdown__item">
+						<div
+							v-for="(action, index) in actions"
+							:key="index"
+							class="kt-field-select__actions__item"
+							@click="action.onClick"
+							v-text="action.label"
+						/>
+					</div>
+				</div>
 			</ElSelect>
 		</div>
 		<template v-slot:actionIcon="{ classes, handleClear, showClear }">
--- a/packages/kotti-ui/source/kotti-field-select/KtFieldSingleSelect.vue
+++ b/packages/kotti-ui/source/kotti-field-select/KtFieldSingleSelect.vue
@@ -25,6 +25,18 @@
 					:label="option.label"
 					:value="option.value"
 				/>
+
+				<div v-if="actions.length" class="kt-field-select__actions">
+					<div class="el-select-dropdown__item">
+						<div
+							v-for="(action, index) in actions"
+							:key="index"
+							class="kt-field-select__actions__item"
+							@click="action.onClick"
+							v-text="action.label"
+						/>
+					</div>
+				</div>
 			</ElSelect>
 		</div>
 		<template v-slot:actionIcon="{ classes, handleClear, showClear }">
--- a/packages/kotti-ui/source/kotti-field-select/constants.ts
+++ b/packages/kotti-ui/source/kotti-field-select/constants.ts
@@ -21,6 +21,20 @@ const KOTTI_FIELD_SELECT_PROPS = {
 			),
 	},
 	placeholder: { default: null, type: String },
+	actions: {
+		default: () => [],
+		type: Array,
+		validator: (value: unknown): value is Shared.Props['actions'] =>
+			Array.isArray(value) &&
+			value.every(
+				// eslint-disable-next-line
+				(action: any): action is Shared.Action =>
+					typeof action === 'object' &&
+					action !== null &&
+					typeof action.label === 'string' &&
+					typeof action.onClick === 'function',
+			),
+	},
 }
 
 export const KOTTI_FIELD_SELECT_SUPPORTS: KottiField.Supports = {
--- a/packages/kotti-ui/source/kotti-field-select/styles.scss
+++ b/packages/kotti-ui/source/kotti-field-select/styles.scss
@@ -29,6 +29,25 @@
 		}
 	}
 
+	&__actions {
+		&:before {
+			display: block;
+			height: 1px;
+			margin: 0 0.6rem;
+			content: '';
+			background: var(--gray-20);
+		}
+
+		&__item {
+			color: var(--interactive-01);
+			cursor: pointer;
+
+			&:hover {
+				background-color: var(--ui-01);
+			}
+		}
+	}
+
 	&--multiple {
 		$vertical-tag-gap: 2px;
 		$horizontal-tag-gap: 4px;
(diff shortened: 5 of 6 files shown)


packages/kotti-ui/source/kotti-field-select/constants.ts
```diff
@@ -31,6 +31,18 @@ export const KOTTI_FIELD_SELECT_SUPPORTS: KottiField.Supports = {
 
 export const KOTTI_FIELD_SINGLE_SELECT_PROPS = {
 	...KOTTI_FIELD_SELECT_PROPS,
+	actions: {
+		default: () => null,
```
carsoli, line 35: would never recommend a default `null` for an array

packages/kotti-ui/source/kotti-field-select/KtFieldSingleSelect.vue
```diff
@@ -25,6 +25,16 @@
 					:label="option.label"
 					:value="option.value"
 				/>
+				<div v-if="actions" class="kt-field-select-actions">
```
carsoli, line 28: condition should be explicit to only show when array length is more than 0 so: `actions.length`, 
i see that ur default prop value is null, which justifies ur choice of condition but i vote for `[]` and checking on length b/c you expect an array to fallback to `[]` and then it's clear to the reader w/o going through further code, that this is an array

as far as i'm concerned, that's not BEM, so maybe:`kt-field-select__actions`

packages/kotti-ui/source/kotti-field-select/KtFieldSingleSelect.vue
```diff
@@ -25,6 +25,16 @@
...
+					<div
+						v-for="action in actions"
+						:key="action.value"
+						class="kt-field-select-actions-item"
```
carsoli, line 32: following BEM, this would then be `kt-field-select__actions__item`

packages/kotti-ui/source/kotti-field-select/constants.ts
```diff
@@ -31,6 +31,18 @@ export const KOTTI_FIELD_SELECT_SUPPORTS: KottiField.Supports = {
...
+		type: Array,
+		validator: (value: unknown[]): value is KottiFieldSingleSelect.Action[] =>
```
carsoli, line 37: value should be `unknown` only even, not unknown array

packages/kotti-ui/source/kotti-field-select/constants.ts
```diff
@@ -31,6 +31,18 @@ export const KOTTI_FIELD_SELECT_SUPPORTS: KottiField.Supports = {
...
+			value.every(
```
carsoli, line 38: first, we check that `value` is an `Array` (`Array.isArray`)
then we check that every value matches what it should
b/c u typed it as an `unknown[]` you already assumed, you get some array
how do you guarantee that this is the case

packages/kotti-ui/source/kotti-field-select/constants.ts
```diff
@@ -31,6 +31,18 @@ export const KOTTI_FIELD_SELECT_SUPPORTS: KottiField.Supports = {
...
+				(action: any): action is KottiFieldSingleSelect.Action =>
+					typeof action === 'object' &&
+					action !== null &&
```
carsoli, line 41: you seem to depend on `action.value` in the template as your key but don't check that such an attribute lives on your `action` object here

packages/kotti-ui/source/kotti-field-select/styles.scss
```diff
@@ -29,6 +29,25 @@
...
 	}
 
+	&-actions {
+		&:before {
+			position: relative;
+			display: block;
+			height: 1px;
+			margin: 0 16px 0 16px;
+			content: '';
+			background: var(--gray-20);
```
carsoli, line 39: i assume all of this is to add a `1px` grey seperator, no?
why the position relative?

packages/kotti-ui/source/kotti-field-select/types.ts
```diff
@@ -32,7 +32,13 @@ export namespace Shared {
 }
 
 export namespace KottiFieldSingleSelect {
-	export type Props = KottiField.Props<Value> & Shared.Props
+	export type Props = KottiField.Props<Value> &
+		Shared.Props & { actions: KottiFieldSingleSelect.Action[] }
+
+	export type Action = {
+		label: string
```
carsoli, line 39: your type definition doesn't match the template usage of `action.value`

packages/kotti-ui/source/kotti-field-select/KtFieldSingleSelect.vue
```diff
@@ -25,6 +25,16 @@
 					:label="option.label"
 					:value="option.value"
 				/>
+				<div v-if="actions" class="kt-field-select-actions">
+					<div
+						v-for="action in actions"
+						:key="action.value"
```
carsoli, line 31: this seems pretty unintentional after going through your code, you can for a fixed string  + `index`

packages/kotti-ui/source/kotti-field-select/constants.ts
```diff
@@ -31,6 +31,18 @@ export const KOTTI_FIELD_SELECT_SUPPORTS: KottiField.Supports = {
 
 export const KOTTI_FIELD_SINGLE_SELECT_PROPS = {
 	...KOTTI_FIELD_SELECT_PROPS,
+	actions: {
+		default: () => null,
+		type: Array,
+		validator: (value: unknown[]): value is KottiFieldSingleSelect.Action[] =>
+			value.every(
+				(action: any): action is KottiFieldSingleSelect.Action =>
```
carsoli, line 39: any reason for choosing `any` over `unknown`

packages/kotti-ui/source/kotti-field-select/constants.ts
```diff
@@ -31,6 +31,18 @@ export const KOTTI_FIELD_SELECT_SUPPORTS: KottiField.Supports = {
 
 export const KOTTI_FIELD_SINGLE_SELECT_PROPS = {
 	...KOTTI_FIELD_SELECT_PROPS,
```
carsoli, line 33: if you intend to extend this to `KtFieldMultiSelect` as well, then move the prop validator and defintion to `KOTTI_FIELD_SELECT_PROPS`, it should already work, and you can in theory use the same template code
you would need a sub-component to render in both that contains the template code and the styles

packages/kotti-ui/source/kotti-field-select/styles.scss
```diff
@@ -29,6 +29,25 @@
 		}
 	}
 
+	&-actions {
```
carsoli, line 32: don't forget to change this if you follow BEM, likewise for any other class accessors below

packages/kotti-ui/source/kotti-field-select/styles.scss
carsoli: I don't see an `absolute` wrapper directly but maybe that's the popper's body?

packages/kotti-ui/source/kotti-field-select/constants.ts
FlorianWendelborn: Additionally, you don’t need to use a function for primitive defaults (like `null`)

Yeah, this should be `() => []`

packages/kotti-ui/source/kotti-field-select/KtFieldSingleSelect.vue
```diff
@@ -25,6 +25,16 @@
...
+						class="kt-field-select-actions-item"
+						@click="action.onClick"
+					>
+						{{ action.label }}
```
FlorianWendelborn, line 35: Would prefer `v-text="action.label"` instead

packages/kotti-ui/source/kotti-field-select/constants.ts
FlorianWendelborn: I’d say this should definitely be extended to `KtFieldMultiSelect` as well. Will be really confusing if it’s not working when someone tries this

As far as I can see, this should be trivial to support for both at the same time

@RueRivoli do you want to do this or should I take over?

packages/documentation/pages/usage/components/form-fields.vue
```diff
@@ -738,7 +742,11 @@ export default defineComponent({
 				Object.assign(additionalProps, {
 					options: singleOrMultiSelectOptions,
 				})
-
+			if (component === 'KtFieldSingleSelect') {
+				Object.assign(additionalProps, {
+					actions: singleSelectActions,
```
FlorianWendelborn, line 747: This is right now always active, implying that it’s a required field. You should add it as an option in the form field generator instead

packages/kotti-ui/source/kotti-field-select/styles.scss
```diff
@@ -29,6 +29,25 @@
...
+		&:before {
+			position: relative;
+			display: block;
+			height: 1px;
+			margin: 0 16px 0 16px;
```
FlorianWendelborn, line 37: The `margin`s and `padding`s don’t fit with the design. I realize you just integrated them into the existing Element UI styles though.

I think this is a great opportunity to align the design with the implementation.

Do you want to handle this, or should I start on it?

FlorianWendelborn (changes requested):

packages/kotti-ui/source/kotti-field-select/types.ts
RueRivoli: the use of action.value was a mistake from me. action doesn't contain value

packages/kotti-ui/source/kotti-field-select/styles.scss
RueRivoli: Will look at it in a bit. I focus on the previous comment first.

packages/kotti-ui/source/kotti-field-select/constants.ts
RueRivoli: value doesnt exist in action. It was a mistake from me

packages/kotti-ui/source/kotti-field-select/constants.ts
RueRivoli: Getting error label is not defined on type object with unknown

packages/documentation/pages/usage/components/form-fields.vue
```diff
@@ -111,6 +111,17 @@
...
 				<KtFormControllerObject formKey="additionalProps" v-if="componentDefinition.additionalProps.length > 0">
 					<h4>Additional Props</h4>
+					<KtFieldSingleSelect
+						v-if="componentDefinition.additionalProps.includes('actions')"
+						formKey="actions"
+						isOptional
+						label="action"
+						:options="[
+							{ label: 'no action', value: '' },
+							{ label: 'create', value: 'create' },
```
FlorianWendelborn, line 121: This is basically a `KtFieldSingleSelect` that’s used as a boolean. Perhaps `KtFieldToggleGroup` or `KtFieldToggle` is preferable?

packages/documentation/pages/usage/components/form-fields.vue
```diff
@@ -561,6 +572,7 @@ export default defineComponent({
 
 		const settings = ref<{
 			additionalProps: {
+				actions: string
```
FlorianWendelborn, line 575: Should be `Kotti.Field*.Value`

packages/documentation/pages/usage/components/form-fields.vue
```diff
@@ -595,6 +607,7 @@ export default defineComponent({
 			validation: Kotti.Field.Validation.Result['type']
 		}>({
 			additionalProps: {
+				actions: '',
```
FlorianWendelborn, line 610: `null`

packages/documentation/pages/usage/components/form-fields.vue
```diff
@@ -734,10 +747,22 @@ export default defineComponent({
 					hideChangeButtons: settings.value.additionalProps.hideChangeButtons,
 				})
 
-			if (['KtFieldMultiSelect', 'KtFieldSingleSelect'].includes(component))
+			if (['KtFieldMultiSelect', 'KtFieldSingleSelect'].includes(component)) {
 				Object.assign(additionalProps, {
 					options: singleOrMultiSelectOptions,
 				})
+				if (settings.value.additionalProps.actions !== '') {
+					const action = [
```
FlorianWendelborn, line 755: This looks like it can easily be inlined

packages/kotti-ui/source/kotti-field-select/KtFieldMultiSelect.vue
```diff
@@ -44,6 +44,18 @@
...
+				<div v-if="actions.length">
+					<div class="kt-field-select__actions"></div>
+					<div class="el-select-dropdown__item">
+						<div
+							v-for="(action, index) in actions"
+							:key="`action_${index}`"
+							class="kt-field-select__actions__item"
+							@click="action.onClick"
+							v-text="action.label"
+						></div>
```
FlorianWendelborn, line 56: Can be self-closing

packages/kotti-ui/source/kotti-field-select/KtFieldSingleSelect.vue
```diff
@@ -25,6 +25,19 @@
...
+				<div v-if="actions.length">
+					<div class="kt-field-select__actions"></div>
+					<div class="el-select-dropdown__item">
+						<div
+							v-for="(action, index) in actions"
+							:key="`action_${index}`"
+							class="kt-field-select__actions__item"
+							@click="action.onClick"
+							v-text="action.label"
+						></div>
```
FlorianWendelborn, line 38: Can be self-closing

packages/kotti-ui/source/kotti-field-select/KtFieldSingleSelect.vue
```diff
@@ -25,6 +25,19 @@
 					:label="option.label"
 					:value="option.value"
 				/>
+
+				<div v-if="actions.length">
+					<div class="kt-field-select__actions"></div>
```
FlorianWendelborn, line 30: Why does this empty tag exist?

packages/kotti-ui/source/kotti-field-select/types.ts
```diff
@@ -20,9 +20,15 @@ export namespace Shared {
 
 	export type Props = {
 		options: Shared.Entry[]
+		actions: Shared.Action[]
```
FlorianWendelborn, line 23: Should be sorted for readability

packages/kotti-ui/source/kotti-field-select/KtFieldMultiSelect.vue
```diff
@@ -44,6 +44,18 @@
 					:label="option.label"
 					:value="option.value"
 				/>
+				<div v-if="actions.length">
+					<div class="kt-field-select__actions"></div>
```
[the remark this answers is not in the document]
RueRivoli: to create a separator before and keep the same padding on the div "create" as the other dropdown items

packages/kotti-ui/source/kotti-field-select/types.ts
```diff
@@ -19,10 +19,16 @@ export namespace Shared {
 	}
 
 	export type Props = {
+		actions: Shared.Entry[]
```
FlorianWendelborn, line 22: Is this actually correct?

packages/kotti-ui/source/kotti-field-select/styles.scss
FlorianWendelborn: I don’t think this is resolved yet.

packages/documentation/pages/usage/components/form-fields.vue
```diff
@@ -111,6 +111,13 @@
 				/>
 				<KtFormControllerObject formKey="additionalProps" v-if="componentDefinition.additionalProps.length > 0">
 					<h4>Additional Props</h4>
+					<KtFieldToggle
+						v-if="componentDefinition.additionalProps.includes('actions')"
+						formKey="actions"
+						isOptional
+						label="create new"
```
FlorianWendelborn, line 118: Should be reflective of the prop and may or may not need a `helpText` depending on if it’s obvious what’s happening with the label

FlorianWendelborn (approved):
Fixed the remaining bits & squashed

PR #391, opened 2021-04-09
Feature enable popover text icon items to be tracked with data test

RueRivoli:
### Description
Write a new prop data-set in internal component IconTextItem to track actions on items.

### DIFF
--- a/packages/kotti-ui/source/kotti-popover/KtPopover.vue
+++ b/packages/kotti-ui/source/kotti-popover/KtPopover.vue
@@ -12,7 +12,7 @@
 				<IconTextItem
 					v-for="(option, index) in options"
 					:key="index"
-					:data-test="option.dataTest"
+					:dataTest="option.dataTest"
 					:icon="option.icon"
 					:isDisabled="option.isDisabled"
 					:label="option.label"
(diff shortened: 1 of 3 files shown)


packages/kotti-ui/source/kotti-popover/KtPopover.vue
```diff
@@ -12,6 +12,7 @@
 				<IconTextItem
 					v-for="(option, index) in options"
 					:key="index"
+					:data-test="option.dataTest"
```
carsoli, line 15: does `IconTextItem` accept the data-test 

what do u do with it, this doesn't eventually achieve ur goal
b/c u pass it down but you don't eventualy bind it on a lower level

carsoli: NVM - attributes are by default auto intehrited, unless a flag specifies otherwise, so I'll approve :)

carsoli (approved):

packages/kotti-ui/source/kotti-popover/KtPopover.vue
FlorianWendelborn: While the reasoning may not be correct, I still agree with what you’re saying though. Eventually, we may accidentally break this feature if it’s handled from only the parent’s side. I’ll add `dataTest` to the props

FlorianWendelborn (approved):

Issue #190, opened 2020-04-07
KtButton: Re-Apply Fix for Buttons Being Implicitly Type=Submit


FlorianWendelborn:
Had to be reverted as a workaround, should be re-applied ASAP.

PR #392, opened 2021-04-09
fix(KtButton): composition api and types, fix implicit submit, loading -> isLoading

HansAuger:
* **BREAKING**: `loading` => `isLoading`
* **BREAKING**: `isSubmit` for old bug-causing implicit submit behavior (#190)
* refactored to `@vue/composition-api`
* refactored SCSS


packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -24,35 +28,23 @@ export default {
...
-	data() {
+	setup(props, { emit, slots }) {
+		const hasSlot = computed(() => Boolean(slots.default))
 		return {
-			isHover: false,
+			handleClick: (event) => emit(KottiButton.Event.CLICK, event),
+			hasSlot,
+			mainClasses: computed(() => ({
+				'kt-button': true,
+				[props.type]: true,
```
carsoli, line 38: this can translate to a `null` key, as in
```
mainClasses = { 
//... other key value pairs ,

null: true 
}
```
in theory, wouldn't make a difference, but i hate it :(

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -24,35 +28,23 @@ export default {
...
+				icon: props.icon,
+				'icon-only': props.icon && !hasSlot.value && !props.label,
+				'kt-button--is-block': props.isBlock,
+				'kt-button--is-multiline': props.isMultiline,
+				sm: props.size === 'small',
```
carsoli, line 43: not on you but
i think we pass `sm`, `md`, `lg` to the `size` prop and apparently we only either `sm` or not

would propose 
`[props.size]: true`

HansAuger: hmmm , [documentation](https://3yourmind.github.io/kotti/usage/components/button) claims differently. Will have a look

packages/kotti-ui/source/kotti-button/types.ts
```diff
@@ -0,0 +1,19 @@
+import { Yoco } from '@3yourmind/yoco'
+
+export namespace KottiButton {
+	export type PropsInternal = {
+		icon: Yoco.Icon | ''
```
FlorianWendelborn, line 5: Should be `| null`

packages/kotti-ui/source/kotti-button/types.ts
```diff
@@ -0,0 +1,19 @@
...
+		isBlock: boolean
+		isMultiline: boolean
+		label: string | null
+		loading: boolean
+		size: 'small' | null
```
FlorianWendelborn, line 10: Should use an `enum` and not have a `null`

packages/kotti-ui/source/kotti-button/types.ts
```diff
@@ -0,0 +1,19 @@
...
+		type: 'primary' | 'secondary' | 'text' | 'danger'
```
FlorianWendelborn, line 11: Should be an `enum`

packages/kotti-ui/source/kotti-button/types.ts
```diff
@@ -0,0 +1,19 @@
...
+	}
+
+	export type Props = Partial<PropsInternal>
+
+	export enum Event {
```
FlorianWendelborn, line 16: Doesn’t fit well with our existing event-related namespaces

packages/kotti-ui/source/types/kotti.ts
```diff
@@ -31,6 +31,7 @@ export {
 export { KottiModal as Modal } from '../kotti-modal/types'
 export { KottiNavbar as Navbar } from '../kotti-navbar/types'
 export { KottiRow as Row } from '../kotti-row/types'
+export { KottiButton as Button } from '../kotti-button/types'
```
FlorianWendelborn, line 34: Not sorted

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -11,10 +11,14 @@
 	</component>
 </template>
 
-<script>
-export default {
+<script lang="ts">
+import { computed, defineComponent } from '@vue/composition-api'
+
+import { KottiButton } from './types'
```
FlorianWendelborn, line 17: Missing newline

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -24,35 +28,23 @@ export default {
 		size: { default: null, type: String },
 		type: { default: null, type: String },
 	},
-	data() {
+	setup(props, { emit, slots }) {
+		const hasSlot = computed(() => Boolean(slots.default))
```
FlorianWendelborn, line 32: Either should be inlined or have a newline afterwards

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -24,35 +28,23 @@ export default {
...
 		return {
-			isHover: false,
+			handleClick: (event) => emit(KottiButton.Event.CLICK, event),
+			hasSlot,
+			mainClasses: computed(() => ({
+				'kt-button': true,
+				[props.type]: true,
+				icon: props.icon,
+				'icon-only': props.icon && !hasSlot.value && !props.label,
```
FlorianWendelborn, line 40: This class seems weirdly named

packages/kotti-ui/source/kotti-button/KtButton.vue
FlorianWendelborn, line 43: Should be `kt-button--size-${size}`

FlorianWendelborn (changes requested):

packages/documentation/pages/usage/components/form-fields.vue
```diff
@@ -412,6 +419,7 @@ const components: Array<{
 		formKey: 'singleSelectValue',
 		name: 'KtFieldRadioGroup',
 		supports: { clear: false, decoration: false, tabIndex: true },
+		slots: ['header', 'body'],
```
FlorianWendelborn, line 422: Why is this here?

packages/kotti-ui/source/kotti-button/types.ts
```diff
@@ -0,0 +1,28 @@
...
+export namespace KottiButton {
+	export enum Type {
+		PRIMARY = 'primary',
+		SECONDARY = 'secondary',
+		TEXT = 'text',
+		DANGER = 'danger',
+	}
+
+	export enum Size {
+		SMALL = 'small',
```
FlorianWendelborn, line 12: I think this is a breaking change

packages/kotti-ui/source/kotti-button/types.ts
```diff
@@ -0,0 +1,28 @@
...
+		MEDIUM = 'medium',
```
FlorianWendelborn, line 13: I think this is a breaking change

packages/kotti-ui/source/kotti-button/types.ts
```diff
@@ -0,0 +1,28 @@
...
+		LARGE = 'large',
```
FlorianWendelborn, line 14: I think this is a breaking change

packages/kotti-ui/source/kotti-button/types.ts
HansAuger: Can you explain what about this is a breaking change? I suppose you think that originally you would pass `"sm"` instead of `"small"` , mainly because @carsoli stated the same in a previous comment. But neither the [documentation](https://3yourmind.github.io/kotti/usage/components/button) nor the [previous code](https://github.com/3YOURMIND/kotti/pull/392/files#diff-fa0908517b6e4a5446821f5f904942271f3a861afdece9b31bf7f0f4d25c92d4L40) would suggest this is the case. Am I missing something here?

packages/kotti-ui/source/kotti-button/types.ts
HansAuger: Can you explain what about this is a breaking change? I suppose you think that originally you would pass `"lg"` instead of `"large"` , mainly because @carsoli stated the same in a previous comment. The [documentation](https://3yourmind.github.io/kotti/usage/components/button) wouldn't suggest so and I think I'm  actually fixing a bug here because large buttons somehow were dropped. As you can see in the documentation the medium button and the large button have the same height.

packages/kotti-ui/source/kotti-button/types.ts
HansAuger: True, `"medium"` (and btw neither `"md"`) is a supported prop. Will remove

packages/documentation/pages/usage/components/form-fields.vue
HansAuger: Why is any of this file changed at all?

packages/kotti-ui/source/kotti-button/types.ts
FlorianWendelborn: Wow, I assumed that surely nobody would’ve bothered to explicitly transform `small` into `sm` just to "save" 3 characters in the CSS. In that case, this is fine and preferable.

packages/kotti-ui/source/kotti-button/types.ts
FlorianWendelborn: I think it’s much better if we support `medium` and default to it instead of `null`. I’ll finalize this and merge it now

FlorianWendelborn (approved):

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -68,29 +76,46 @@ export default {
...
+		.yoco {
+			font-size: 1rem;
+		}
+
+		&:not(.kt-button--has-content) {
+			padding: 0 var(--unit-2);
+		}
 	}
 
 	// increase selector specificity by using .kt-button twice
```
carsoli, line 121: WTF :D

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -105,119 +130,116 @@ export default {
...
-	background-color: var(--button-main-color);
-	border-color: var(--button-main-color-dark);
+	&--type {
+		&-danger {
+			color: var(--danger);
 
-	&:hover {
-		background-color: var(--button-main-color-dark);
-	}
+			.kt-circle-loading {
```
carsoli, line 163: this dude seems to always inherit color - maybe instead of duplications, move it one level above and `color: inherit`

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -105,119 +130,116 @@ export default {
...
 		padding-bottom: var(--unit-1);
 		line-height: var(--default-button-line-height);
-		&.icon {
+
+		&.kt-button--has-icon {
 			align-items: baseline;
 			text-align: left;
+
 			.yoco {
 				position: relative;
```
carsoli, line 139: why relativee

can u maybe flex the contents, and that's it?

packages/kotti-ui/source/kotti-button/KtButton.vue
carsoli: this seems off :)

packages/kotti-ui/source/kotti-button/KtButton.vue
FlorianWendelborn: It’s for the crazy `isMultiline` BS. Really don’t want to touch it, sorry.

packages/kotti-ui/source/kotti-button/KtButton.vue
FlorianWendelborn: It’s `border-color` though. I don’t think it’s easily possible to inherit this via CSS

carsoli (approved):

PR #395, opened 2021-04-12
fix(#394): KtPopover CSS Stacking Context

FlorianWendelborn:


HansAuger (approved):

carsoli:
Functional Testing:

As per the recommendation/warning of Florian, I've already used the build output from kotti as a patch package on our user-app

and it works flawlessly (include the case of having a popper on top of another)

carsoli:
> @carsoli did you explicitly test the case where a form with a select is inside a popover?

I explicitly tested the case where a select is inside a popover.. no form included, but I will wrap it in a form now although I'm not sure why a KtForm makes any difference? @FlorianWendelborn

FlorianWendelborn:
@carsoli the form shouldn’t be important, but glad to hear the select isn’t broken anymore. In case you’re still in a state where you can easily test, a `KtFieldDate*` would also be an interesting case to attempt

carsoli:
@FlorianWendelborn / @santiagoballadares / @HansAuger 
Screenshot from 2022-05-09 15-00-14


what I meant by edge case of popper within popper

FlorianWendelborn:
I just tested the render link and the Select/Date are still entirely broken as far as I can tell

For example, even with the recommended `CLICK`, I’m landing in this state:

image

With `HOVER`, it’s even worse/unusable

packages/documentation/pages/usage/components/popover.vue
```diff
@@ -1,199 +1,151 @@
...
+			>
+				<template #helpText>
+					There are 4 different positions for popovers:
+					<code>right</code>, <code>left</code>, <code>top</code>, and
+					<code>bottom</code>, or <code>auto</code>. All of which can be
+					appended with <code>-start</code> and <code>-end</code> to define
+					placement.
+					<br />
+					<a href="https://atomiks.github.io/tippyjs/v6/all-props/#placement"
+						>See Placement Options here</a
```
FlorianWendelborn, line 59: Add a space before/after the words here to fix the autoformatting being overly restrictive

yarn.lock
```diff
@@ -3528,10 +3528,10 @@
     "@octokit/types" "^6.0.3"
     universal-user-agent "^6.0.0"
 
-"@octokit/openapi-types@^8.3.0":
-  version "8.3.0"
-  resolved "https://registry.yarnpkg.com/@octokit/openapi-types/-/openapi-types-8.3.0.tgz#8bc912edae8c03e002882cf1e29b595b7da9b441"
-  integrity sha512-ZFyQ30tNpoATI7o+Z9MWFUzUgWisB8yduhcky7S4UYsRijgIGSnwUKzPBDGzf/Xkx1DuvUtqzvmuFlDSqPJqmQ==
+"@octokit/openapi-types@^11.2.0":
```
FlorianWendelborn, line 3531: Any idea why this was changed? Looks pretty random

Can you check if the types are still fine in the changelog? The pipelines AFAIK can't check this right now

packages/kotti-ui/source/kotti-popover/components/IconTextItem.vue
```diff
@@ -44,11 +36,26 @@ export default defineComponent<{
...
+				&.kt-popover-options-item--is-selected {
+					color: var(--interactive-01-hover);
+				}
+			}
+		}
+	}
+
 	&--is-disabled {
-		cursor: not-allowed;
+		// `--is-disabled` has same specificity as `--is-clickable`.
```
FlorianWendelborn, line 56: This can actually be fixed by moving `cursor:pointer` into the `:not` below it I think

carsoli:
@santiagoballadares the two popover behave differently on hover, because once the cursor is outside the borders of the original popover (once you start hovering over right side of date range popper), the popper closes -- because it's a triggered on hover


@FlorianWendelborn The click behavior is fixed. For now, the default behavior of the click will be the toggle (i.e. clicking the triggere element again), and the user has two ways of closing the popper explicitly on certain click actions
- by using the tippy element reference and calling `.close()` on it 
- by passing a prop called `clickBehavior` as `KottiPopover.ClickBehavior.HIDE_ON_CLICK_OUTSIDE`

yarn.lock
carsoli: nothing wrong on the `changelog.vue`

carsoli:
https://github.com/3YOURMIND/kotti/issues/624

carsoli (approved):

PR #397, opened 2021-04-13
improvement(KtFieldNumber): Buttons covering input contents 

santiagoballadares:


packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
```diff
@@ -311,6 +311,7 @@ export default defineComponent({
 	display: flex;
 	align-items: center;
 	font-variant-numeric: tabular-nums;
+	margin: 0 -0.5rem;
```
carsoli, line 314: you're probably counter-acting this line https://github.com/3YOURMIND/kotti/blob/a90f3150378ea25379f5b825a9642f49c0d2951b/packages/next/kotti-field/KtField.vue#L176

if so, either link this in a comment so we always have the line, or even better, make some variable for the input-container padding so that when we change this in the future it's immediately updated for the `KtFieldNumber`

obviously 0.5 is not 0.8 but i guess u get my point still, you make it 5/8th of it, I don't mind, whatever u see fit
> my point is merely that it requires one LOC change in the future if need be :)

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
```diff
@@ -94,7 +94,7 @@ export default defineComponent({
 					maximum: null,
 					minimum: props.minimum,
 					offset: -props.step,
-					value: props.value,
+					value: field.currentValue,
```
carsoli, line 97: given that this is the most important part of this PR, i'd heavily recommend changing the title to be fix KtFieldNumber `isDecrementEnabled` b/c that's the thing we should definitely know has been fixed above all, right?

santiagoballadares: Agreed

FlorianWendelborn:
@santiagoballadares I’m gonna split this up so we can merge the `KtTable` changes already. This is two different PRs in one IMO.

FlorianWendelborn:
Drafted until #401 and #402 are merged

packages/kotti-ui/source/kotti-field/KtField.vue
```diff
@@ -99,6 +103,7 @@ import FieldHelpText from './components/FieldHelpText.vue'
 import { KottiField } from './types'
 
 export default defineComponent<{
+	containerPadding: string | null
```
carsoli, line 112: heavily opposing this. 
> Most of our goal when working on the new fields/form was to reduce the size of the API, keeping it short and concise (having exactly the amount of props it needs to not compromise API simplicity over flexibility).

There is a tendency to misuse an API that is huge, since the proper intended usage can get lost in translation.

Another thing is, this leaves room for inconsitent UI already. something the designers are heavily opposed to. 
> one middle ground solution I see is having reactive button sizes with a minimum width depending on the container size

OR: (discuss with designers) 
hiding the maximum when the field container is small (and introducing it as a tippy or as help text)

OR even to the label of the field - think about
```
amount (/100):
[ KT FIELD NUMBER ] 
```

carsoli: you also have the ability to `hideMaximumButtons` so you can have a media-query in the user app that hides the buttons entirely

but I don't this as the only solution that will attend to ur current problem, keep the API at minimum and provided a consistent UI on the long run

FlorianWendelborn: Yeah, personally also not too happy with this, which is one of the main reasons why I decided to split-out the non-controversial changes. So that we can already merge those simple fixes without being blocked by this discussion etc.

FlorianWendelborn: @carsoli the label is also problematic as it’s translated — and thus may take up _a lot_ more space for other languages with higher redundancy

_\*cough\* German_

FlorianWendelborn: @carsoli what do you think about using a negative margin on the buttons instead? That’d require no additional `KtField` API and could be done entirely in CSS

FlorianWendelborn: It’d also prevent this misalignment issue as the buttons wouldn’t even be shown in that case — thus they couldn’t add negative margins: 
image

FlorianWendelborn:
I think this shouldn’t be closed as the buttons do waste a lot of space right now, especially since most of the places where this is useful are inside tables where space is limited. Re-opening.

carsoli:
Is this still relevant @santiagoballadares ?

santiagoballadares:
> 


The origin was a bug/issue in Yoda and no one has complained again but it could happen in the future
We could discuss what solutions are possible and go from there

FlorianWendelborn:
Whatever you end-up deciding, please squash this PR as it contains 40 commits right now

Issue #377, opened 2021-03-19
Expose generic types for Selects `options`


carsoli:
`KtFieldSingleSelect` & `KtFieldMultiSelect` both accept `options` array as a prop
the `value` key on each `option` by default extends `number | string | symbol | boolean | null` but it would be great if we could expose a generic that would limit the `options` to one of those types depending on the usage. 
In turn, utilization of Kotti types is doable without having to type cast.

sample code proposed
```
export type Entry<V extends string | number | boolean | symbol | null = string | number | boolean | symbol | null> = {
		isDisabled?: boolean
		label: string
		value: V
	}
```

PR #407, opened 2021-04-18
fix(#377): Expose generic types for Selects `options`

HansAuger:


packages/kotti-ui/source/kotti-field-select/types.ts
```diff
@@ -12,10 +12,12 @@ export namespace KottiFieldMultiSelect {
 }
 
 export namespace Shared {
-	export type Entry = {
+	export type DefaultValue = string | number | boolean | symbol | null
```
FlorianWendelborn, line 15: I think this doesn’t need to be exported, as `Kotti.Field*Select.Value` already is enough to refer to it if needed. Let’s keep it minimal and inline it instead

FlorianWendelborn, line 15: I think this doesn’t need to be exported, as `Kotti.Field*Select.Value` already is enough to refer to it if needed. Let’s keep it minimal and inline it instead

FlorianWendelborn (changes requested):

FlorianWendelborn (approved):

FlorianWendelborn:
Thanks @HansAuger, merged

PR #415, opened 2021-04-24
refact(KtBreadcrumb): typescript, composition api

HansAuger:
Nothing too fancy

### DIFF
--- a/packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
+++ b/packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
@@ -1,22 +1,22 @@
 <template>
-	<div class="breadcrumb">
+	<div class="kt-breadcrumb">
 		<ol>
 			<li
 				v-for="(breadcrumb, index) in breadcrumbs"
 				:key="index"
-				class="breadcrumb__text"
+				class="kt-breadcrumb__text"
 				:class="textClasses(breadcrumb)"
 			>
 				<span @click="handleClick(breadcrumb)" v-text="breadcrumb.title" />
 				<span v-if="showSeparator(index)">
 					<i
-						v-if="separator.style === 'icon'"
+						v-if="separator.style === KottiBreadcrumb.SeparatorType.ICON"
 						class="yoco"
 						v-text="separator.value"
 					/>
 					<span
-						v-if="separator.style === 'text'"
-						class="breadcrumb__text-separator"
+						v-if="separator.style === KottiBreadcrumb.SeparatorType.TEXT"
+						class="kt-breadcrumb__text-separator"
 						v-text="separator.value"
 					/>
 				</span>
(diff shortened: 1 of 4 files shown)


packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
```diff
@@ -25,11 +25,34 @@
...
+<script lang="ts">
+import { defineComponent } from '@vue/composition-api'
+import { isBoolean, isString } from 'lodash'
+
+import { isYocoIcon } from '../validators'
+
+import { KottiBreadcrumb } from './types'
+
+const breadcrumbValidator = (
+	value: KottiBreadcrumb.Breadcrumb,
```
FlorianWendelborn, line 37: This should be `unknown` or `any`, `x is y` functions are trying to assert that `y` is a sub-type of `x`, otherwise you could just return `true`

FlorianWendelborn, line 37: This should be `unknown` or `any`, `x is y` functions are trying to assert that `y` is a sub-type of `x`, otherwise you could just return `true`

packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
```diff
@@ -38,33 +61,36 @@ export default {
 					value: 'chevron_right',
```
FlorianWendelborn, line 61: Might as well use the `Yoco` enum here I guess.

packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
```diff
@@ -38,33 +61,36 @@ export default {
...
 				}
 			},
+			validator: (
+				value: KottiBreadcrumb.Style,
```
FlorianWendelborn, line 65: `unknown` or `any`

packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
```diff
@@ -38,33 +61,36 @@ export default {
...
-		},
+	setup(props, { root }) {
+		return {
+			handleClick: (item: KottiBreadcrumb.Breadcrumb) => {
+				if (!item.isCompleted || item.to || !root.$router) return
+				root.$router.push(item.to)
+			},
+			showSeparator: (index: number) => index < props.breadcrumbs.length - 1,
+			textClasses: (item: KottiBreadcrumb.Breadcrumb) => ({
+				'kt-breadcrumb__text--completed': item.isCompleted,
```
FlorianWendelborn, line 81: Should be `--is-completed`

FlorianWendelborn (changes requested):

packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
```diff
@@ -25,11 +25,34 @@
...
+			type: Array,
+			required: true,
+			validator: (value: unknown) =>
+				Array.isArray(value) &&
+				value.length > 0 &&
+				value.every(breadcrumbValidator),
+		},
 		separator: {
 			type: Object,
 			default() {
```
carsoli, line 58: I can't comment on https://github.com/3YOURMIND/kotti/pull/415/files#diff-2b4d6c2a4746f4f5fdeda076d4b86e007548a025cb9d9c1dc1bf8e50dcb6c7b3R60 but I'd make an `enum` of the style 

and expose this on the types as well ?
or at least expose the type for `Kotti.BreadCrumb.Seperator`to be `{ style: 'icon', value: Yoco.Icon } | { style: 'text', value: 'string' }`

carsoli: Already done, nvm 👏🏼

packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
```diff
@@ -25,11 +25,34 @@
...
+export default defineComponent<KottiBreadcrumb.PropsInternal>({
 	name: 'KtBreadcrumb',
 	props: {
-		breadcrumbs: { type: Array, required: true },
+		breadcrumbs: {
+			type: Array,
+			required: true,
+			validator: (value: unknown) =>
+				Array.isArray(value) &&
+				value.length > 0 &&
```
carsoli, line 53: would you rather error on prop validation if `breadcrumbs` is an empty array at one point in time (perhaps before mounting a component in the user-app) or just not loading the breadcrumbs if that's the case ? b/c in the latter, this line wouldn't be needed, and instead the `v-for` in the template would render nothing

packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
```diff
@@ -25,46 +25,76 @@
...
-		breadcrumbs: { type: Array, required: true },
+		breadcrumbs: {
+			type: Array,
+			required: true,
+			validator: (value: unknown) =>
+				Array.isArray(value) && value.every(breadcrumbValidator),
+		},
 		separator: {
 			type: Object,
 			default() {
```
FlorianWendelborn, line 58: Should use arrow function & type annotate the result. Otherwise, TS will not check if you keep the default up-to-date in future iterations on `KtBreadcrumb`

packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
```diff
@@ -25,46 +25,76 @@
...
+	isString(value.to)
+
+export default defineComponent<KottiBreadcrumb.PropsInternal>({
 	name: 'KtBreadcrumb',
 	props: {
-		breadcrumbs: { type: Array, required: true },
+		breadcrumbs: {
+			type: Array,
+			required: true,
+			validator: (value: unknown) =>
```
FlorianWendelborn, line 53: Missing return type annotation

FlorianWendelborn (changes requested):
LGTM besides those minor remarks. Please also resolve conversations if you fixed them. It’s harder for reviewers to resolve them as they don’t even show up on the diff anymore when outdated.

FlorianWendelborn (approved):

packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
```diff
@@ -25,46 +25,74 @@
...
-		},
-		handleClick(item) {
-			if (!item.isCompleted || !item.to || !this.$router) return
-			this.$router.push(item.to)
-		},
+	setup(props) {
+		return {
+			handleClick: (item: KottiBreadcrumb.Breadcrumb) => {
+				if (!item.isCompleted) return
+				item.onClick()
```
FlorianWendelborn, line 78: How does this line not error? `item` only has `.to` in this PR somehow

PR #416, opened 2021-04-24
Kt breadcrumb onClick Proposal

HansAuger:
Introduces a breaking change.

Use `onClick` event handler instead of `to`, which leads to tight coupling with Vue's router.

Theoretically we can keep the `to`s around, if we want to stay backwards compatible. But seemingly we don't and that would be unnecessary redundancy.

It's probably worthwhile to mention, that in our codebase we hack around this problem bu having a second breadcrumbs component, that we use within the basket wizard


FlorianWendelborn (approved):
Was always against the tight coupling, especially for re-usability for apps that may not even want to use `vue-router`

carsoli (approved):

carsoli:
agree on the proposal, approved, won't merged till #415 is finalized

PR #417, opened 2021-04-29
improvement(KtBanner): composition api; put css into component; sanitize;…

HansAuger:
Nothing too fancy
However, introduces some breaking changes:
prop `isGrey` renamed to `isGray`
prop `switchText` renamed to `expandLabel`
prop `switchCloseText` renamed to `expandCloseLabel`
remove `glyph` slot

removes "banner" class, so yoda HACK in `service-panel/src/features/materials/views/Base.vue` needs to be addressed


packages/kotti-ui/source/kotti-banner/index.ts
```diff
@@ -12,8 +12,14 @@ export const KtBanner = attachMeta(makeInstallable(KtBannerVue), {
...
 	},
 	slots: {
-		expand: { description: null, scope: null },
-		glyph: { description: null, scope: null },
+		expand: {
+			description: 'Content when Banner is expended',
+			scope: null,
+		},
+		glyph: {
+			description: 'Use if you want to showw something else than a yoco icon',
```
FlorianWendelborn, line 20: 1. Typo in `showw`
2. Can you check if we even use this? This might be removeable

packages/kotti-ui/source/kotti-banner/index.ts
```diff
@@ -12,8 +12,14 @@ export const KtBanner = attachMeta(makeInstallable(KtBannerVue), {
 			'https://www.figma.com/file/0yFVivSWXgFf2ddEF92zkf/Kotti-Design-System?node-id=1439%3A5',
 	},
 	slots: {
-		expand: { description: null, scope: null },
-		glyph: { description: null, scope: null },
+		expand: {
+			description: 'Content when Banner is expended',
```
FlorianWendelborn, line 16: Typo in `expanded`

packages/kotti-ui/source/kotti-banner/KtBanner.vue
```diff
@@ -1,81 +1,109 @@
...
 	<div :class="bannerClass">
-		<div class="collapse">
-			<div class="glyph">
-				<span v-if="icon">
+		<div class="kt-banner__collapse">
+			<div class="kt-banner__glyph">
+				<slot name="glyph">
 					<i class="yoco" v-text="icon" />
-				</span>
-				<span v-else>
```
FlorianWendelborn, line 8: Technically a breaking change, as it changes the behavior if both the icon and slot are passed. Needs to be documented in the next release (and therefore PR description) and needs to be sanity-checked in yoda.

packages/kotti-ui/source/kotti-banner/KtBanner.vue
```diff
@@ -1,81 +1,109 @@
...
+export default defineComponent<KottiBanner.PropsInternal>({
 	name: 'KtBanner',
+	components: { KtButton },
 	props: {
-		isGrey: {
-			type: Boolean,
-			default: false,
-		},
-		message: String,
+		actionText: { type: String, default: '' },
```
FlorianWendelborn, line 37: Should default to `null` to indicate that it’s not always filed. Also needs adjusted `v-if` above and type definitions then

packages/kotti-ui/source/kotti-banner/KtBanner.vue
```diff
@@ -1,81 +1,109 @@
...
-		},
-		switchCloseText: {
-			type: String,
-			default: 'Close',
+			default: Yoco.Icon.ANNOUNCE,
+			validator: isYocoIcon,
 		},
+		isGray: { type: Boolean, default: false },
+		message: { type: String, required: true },
+		switchCloseText: { type: String, default: 'Close' },
```
FlorianWendelborn, line 45: Is this heavily utilized and often kept as-is in yoda? If yes, then ideally this would be covered by kotti translations

packages/kotti-ui/source/kotti-banner/KtBanner.vue
```diff
@@ -1,81 +1,109 @@
...
+		switchText: { type: String, default: 'View' },
```
FlorianWendelborn, line 46: Is this heavily utilized and often kept as-is in yoda? If yes, then ideally this would be covered by kotti translations

packages/kotti-ui/source/kotti-banner/KtBanner.vue
```diff
@@ -1,81 +1,109 @@
...
+	display: flex;
+	flex-direction: column;
+	padding: var(--unit-2) var(--unit-4);
+	background: var(--white);
+	border: 1px solid var(--gray-20);
+	border-radius: var(--border-radius);
+
+	&--is-gray {
+		background: var(--gray-20);
+		border: 0;
```
FlorianWendelborn, line 73: Should be `border: none` AFAIK

Also, please make sure that not having a border doesn’t influence the size/padding here. Could easily yield off-by-one-pixel errors

packages/kotti-ui/source/kotti-banner/KtBanner.vue
```diff
@@ -1,81 +1,109 @@
...
+	&__collapse {
+		display: flex;
+		flex-direction: row;
+		align-items: center;
+		width: 100%;
+	}
+
+	&__glyph {
+		width: 1.6rem;
+		line-height: 1;
```
FlorianWendelborn, line 85: Is this an alias for `1em` or how does this work?

FlorianWendelborn (changes requested):

packages/kotti-ui/source/kotti-banner/index.ts
HansAuger: It's not used. Will remove since I'm not a fan

packages/kotti-ui/source/kotti-banner/KtBanner.vue
HansAuger: will be removed anyways but basically means `1 * font-size`
see https://developer.mozilla.org/en-US/docs/Web/CSS/line-height

packages/kotti-ui/source/kotti-banner/KtBanner.vue
HansAuger: Well, the breaking change will be, that we abolish this slot entirely. It is not used in yoda whatsoever btw

packages/kotti-ui/source/kotti-banner/KtBanner.vue
HansAuger: will change to `1em` for clarity

packages/kotti-ui/source/types/kotti.ts
```diff
@@ -1,6 +1,7 @@
 export { KottiAvatar as Avatar } from '../kotti-avatar/types'
 export { KottiAvatarGroup as AvatarGroup } from '../kotti-avatar-group/types'
 export { KottiBreadcrumb as Breadcrumb } from '../kotti-breadcrumb/types'
+export { KottiBanner as Banner } from '../kotti-banner/types'
```
FlorianWendelborn, line 4: Should be sorted

packages/kotti-ui/source/kotti-banner/types.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { SpecifyRequiredProps } from '../types/utilities'
+
+export namespace KottiBanner {
+	export type PropsInternal = {
+		actionText: string | null
+		icon: Yoco.Icon
+		isGray: boolean
+		message: string
+		expandCloseLabel: string
```
FlorianWendelborn, line 11: Shouldn’t this be `string | null` so you can default to the translations?

packages/kotti-ui/source/kotti-banner/types.ts
```diff
@@ -0,0 +1,21 @@
...
+		expandLabel: string
```
FlorianWendelborn, line 12: Shouldn’t this be `string | null` so you can default to the translations?

packages/kotti-ui/source/kotti-banner/types.ts
```diff
@@ -0,0 +1,21 @@
...
+	}
+
+	export type Props = SpecifyRequiredProps<PropsInternal, 'message'>
+
+	export type Translations = {
+		expandLabel: string
```
FlorianWendelborn, line 18: Should be sorted

packages/kotti-ui/source/kotti-banner/index.ts
```diff
@@ -12,8 +12,10 @@ export const KtBanner = attachMeta(makeInstallable(KtBannerVue), {
 			'https://www.figma.com/file/0yFVivSWXgFf2ddEF92zkf/Kotti-Design-System?node-id=1439%3A5',
 	},
 	slots: {
-		expand: { description: null, scope: null },
-		glyph: { description: null, scope: null },
+		expand: {
```
FlorianWendelborn, line 15: Should probably be `default` now?

packages/documentation/pages/usage/components/banner.vue
```diff
@@ -54,14 +51,12 @@ interface.
 <KtBanner
 	message="You material is not available to users"
```
FlorianWendelborn, line 52:
```suggestion
	message="Your material is not published yet"
```

packages/kotti-ui/source/kotti-banner/KtBanner.vue
```diff
@@ -1,81 +1,106 @@
...
+		const translations = useTranslationNamespace('KtBanner')
+
 		return {
-			isExpand: false,
+			bannerClass: computed(() => ({
+				'kt-banner': true,
+				'kt-banner--is-gray': props.isGray,
+			})),
+			expandable: computed(() => Boolean(slots.expand)),
+			isExpand: ref(false),
```
FlorianWendelborn, line 58: Would rename to `isExpanded`

FlorianWendelborn (changes requested):

packages/kotti-ui/source/kotti-banner/index.ts
HansAuger: True

FlorianWendelborn (approved):

PR #422, opened 2021-04-30
updating selected dependendies

HansAuger:


FlorianWendelborn:
I’m actually gonna close this as it’s not necessary to fix and would take quite some time to adjust the eslint stuff

PR #426, opened 2021-05-01
fix(documentation): remove props that pass null to generic component …

### LINKED ISSUE #347: KtFieldNumber: setting step or minimum can  break behavior
**Describe the bug**
For KtFieldNumber the props minimum and step have a weird kind of interconnection:

** To Reproduce **
1. In the kotti docs  go to `KtFieldNumber`
2. If not setting minimum and  setting step=1 , no issue
3. If setting minmum=0 and step=1, the fields will truncate the numbers after the decimals, aka you can not write 8,1 . The length of truncation is dependend on the number of  decimal points in the step value, for example if setting step=0.01 the value  can be 8,01 but not 8,001
4.  If setting minimum=0 and not setting step , all input is prohibited

**Expected behavior**
- step should never influence the number of allowed digits after the comma (it  does not do this if you provide no minimum, even providing only a maximum is  no problem)

 - OS: macOS
 - Browser Firefox
 - Version 85.0b9

HansAuger:
Fixes #347

As it turns out  this bug happens only in the documentation, because without being aware of it we are passing `:step="null"` when binding the computed props to the generic KtField component.

We probably never want to do this, since passing any props will override the props default value, so props that pass null can be filtered out.


FlorianWendelborn:
> We probably never want to do this, since passing any props will override the props default value, so props that pass null can be filtered out.

That’s a good point. @carsoli and I actually recently fixed the `:step="null"` behavior...

I’m not 100% sure though if this is actually safe for everything. I’d much rather properly handle this in a complete refactor of the code generator. It was never meant to scale to the amount of usage it now has... there’s lots of random exceptions in the code

PR #427, opened 2021-05-01
fix(#330): Conserve decimal separator when deleting

### LINKED ISSUE #330: KtFieldNumber: Deletes Decimal Separator when Deleting
1. Type `1.2`
2. Hit backspace (yields `1`, expected: `1.`)
3. Hit backspace (yields `null`, expected `1`)

HansAuger:
Closes #330

The idea is to only change the representational string when necessary, aka the number it represents is different from the current number

### DIFF
--- a/packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
+++ b/packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
@@ -122,7 +122,7 @@ export default defineComponent({
 
 		watch(
 			() => field.currentValue,
-			(newNumber, oldNumber) => {
+			(newNumber) => {
 				const newString = toString(newNumber)
 				const truncatedNumber = toNumber(newString)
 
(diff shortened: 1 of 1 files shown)


packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
```diff
@@ -240,9 +241,9 @@ export default defineComponent({
 					})
 
 				if (isTypedNumberValid) {
+					internalStringValue.value = valueWithoutLeadingZeroes
```
FlorianWendelborn, line 244: Why was this changed? This looks like a bug to me

HansAuger: It is needed to make sure that `internalStringValue` has assumed the string value of the latest input before potentially triggering `field.setValue`.

That on the other hand is important, because we use `internalStringValue` to fix the bug, and in order to work it must have assumed the string value of the latest input.

In question of how this looks like a bug, can you make your suspicion more tangible. As a preemptive explanation, note that before my change both `if` and `else` will sooner or later update the internalStringValue to whatever they see fit. So the only problem that I see is a flicker, when both line 244 and line 149 are run. However, that would mean that `nextNumber` is different from `internalStringValue.value` in numerical value, which shouldn't be possible.

FlorianWendelborn: Wouldn’t this prematurely change the `internalStringValue`, so that it stays as-is even if the parent rejects the attempted value change?

HansAuger:
Can be closed after the recent changes to `KtFieldNumber`

PR #430, opened 2021-05-02
429 add kt label value

Isokaeder:
Proposal for `KtLabelValue`


HansAuger:
I asked design to take a look at the current design and adjust if they see reason to do so.

PR #447, opened 2021-06-01
feat(#439): add interactivity and position to KtLine

### LINKED ISSUE #439: Extend KtLine features
We have designs that put the text to the right and support links as text. That should be implemented in Kotti.

Designs: https://www.figma.com/file/0yFVivSWXgFf2ddEF92zkf/Kotti-Design-System?node-id=95%3A771

HansAuger:
Closes #439

### DIFF
--- a/packages/kotti-ui/source/kotti-line/KtLine.vue
+++ b/packages/kotti-ui/source/kotti-line/KtLine.vue
@@ -1,13 +1,13 @@
 <template>
 	<div class="kt-line">
-		<div class="kt-line__line" />
-		<div v-if="text" class="kt-line__text" v-text="text" />
-		<div v-if="text" class="kt-line__line" />
+		<div v-if="position !== Position.LEFT" class="kt-line__line" />
+		<div v-if="text" :class="textClasses" @click="handleClick" v-text="text" />
+		<div v-if="position !== Position.RIGHT" class="kt-line__line" />
 	</div>
 </template>
 
 <script lang="ts">
-import { defineComponent } from '@vue/composition-api'
+import { computed, defineComponent } from '@vue/composition-api'
 
 import { makeProps } from '../make-props'
 
@@ -16,6 +16,18 @@ import { KottiLine } from './types'
 export default defineComponent<KottiLine.PropsInternal>({
 	name: 'KtLine',
 	props: makeProps(KottiLine.propsSchema),
+	setup(props, { emit }) {
+		return {
+			handleClick: () => {
+				if (props.isInteractive) emit('click')
+			},
+			textClasses: computed(() => ({
+				'kt-line__text': true,
+				'kt-line__text--is-interactive': props.isInteractive,
+			})),
+			Position: KottiLine.Position,
+		}
+	},
 })
 </script>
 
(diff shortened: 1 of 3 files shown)


packages/kotti-ui/source/kotti-line/KtLine.vue
```diff
@@ -1,23 +1,46 @@
...
 </template>
 
 <script lang="ts">
-import { defineComponent } from '@vue/composition-api'
+import { computed, defineComponent } from '@vue/composition-api'
 
 import { KottiLine } from './types'
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const validatePosition = (value: any): value is KottiLine.Position =>
```
carsoli, line 15: can work with `unknown`

packages/kotti-ui/source/kotti-line/KtLine.vue
```diff
@@ -33,15 +56,21 @@ export default defineComponent<KottiLine.InternalProps>({
...
 	&__text {
 		flex: 1 0 auto;
 		padding: 0 0.4rem;
 		font-size: 0.6rem;
-		color: var(--gray-30);
+		color: var(--text-03);
 		text-align: center;
+
+		&--is-interactive {
+			font-size: 0.8rem;
```
carsoli, line 70: why the interactive `font-size` bigger


if the purpose was to make it bigger during `hover` then you're missing `:hover` for example, but I'd refrain from this and just have the `interactive-hover` token (i can't remember the exact token name but we have one for `hover` and one for `active` i.e. when you r clicking the link)

HansAuger: It's actually like this in the design https://www.figma.com/file/0yFVivSWXgFf2ddEF92zkf/Kotti-Design-System?node-id=95%3A771

HansAuger: I don't know why it be like that. It makes sense that the interactive ones are big, because they are supposed to be clickable. So I guess they wanted to have that, without futzing with the old styles because they normal title is not supposed to draw your attention like that.

packages/kotti-ui/source/kotti-line/KtLine.vue
FlorianWendelborn: Actually, in most cases that yields a lot of annoying type errors in those validation functions 😔

Agree with it on an idealistic level though, would be nice if it could be used as it's the best fit, just not practically feasible

packages/kotti-ui/source/kotti-line/KtLine.vue
```diff
@@ -1,23 +1,46 @@
...
 
 import { KottiLine } from './types'
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const validatePosition = (value: any): value is KottiLine.Position =>
+	[
+		KottiLine.Position.LEFT,
+		KottiLine.Position.CENTER,
+		KottiLine.Position.RIGHT,
+	].includes(value)
```
FlorianWendelborn, line 20:
```suggestion
	Object.values(KottiLine.Position).includes(value)
```

FlorianWendelborn (approved):
LGTM besides minor details

packages/kotti-ui/source/kotti-line/KtLine.vue
HansAuger: typescript will actually say that `.includes()` is not allowed because value is not of type `Position` :(

packages/kotti-ui/source/kotti-line/KtLine.vue
carsoli: Can you double check with Designers that this is intentional

I'm not sure if this was design-driven or if you found it out because I am not aware of a Yoda issue that matches

more often than not, you need to get back to the designers who don't intentionally add everything

Also - the text is still not clickable according to the code changes

PR #448, opened 2021-06-01
Resolve(#390): fix sizing issue via css vertical-align

### LINKED ISSUE #390: Adding Icon to button affects its vertical alignment
**Describe the bug**
Adding an icon to a button via the `icon` prop affects the vertical alignment, it goes about 5px up. You can see this on the docs:

**To Reproduce**
Put two buttons into the same row, make one of them use an icon. notice the alignment getting out of whack.

**Expected behavior**
Buttons with icons and Buttons without icons are aligned cleanly

**Screenshots**
Screenshot 2021-04-09 at 00 33 02

- OS: macOS
- Browser: Firefox
- Version: 88.0b7

**Smartphone (please complete the following information):**
Not checked on smartphone

**Aditional Context**
Some Observations:
 - Removing the `<i>` tag via display: none in the browser dev tools fixes the alignment issue. 
 - Changing the `font-size` value of `kt-button.icon` in the browser dev tools changes the alignment without fully bringing the buttons to alignment

HansAuger:
We set KtButton to be `inline-flex` items without taking care of vertical alignment.  This led to strange misalignments between buttons with differing content (like buttons with an icon vs buttons without an icon).

Setting `vertical-align` fixes this issue

Old, revolting, ridiculed with errors:
Screenshot 2021-04-09 at 00 33 02

New, shiny, impeccable:
Screenshot 2021-06-01 at 16 18 09

Closes #390

### DIFF
--- a/packages/kotti-ui/source/kotti-button/KtButton.vue
+++ b/packages/kotti-ui/source/kotti-button/KtButton.vue
@@ -83,6 +83,11 @@ export default defineComponent<KottiButton.PropsInternal>({
 
 	font-weight: 600;
 
+	// Since this is an inline element (display: inline-flex), we need to
+	// care about vertical-alignment. The default setting would cause
+	// different alignments between buttons with or without icon.
+	vertical-align: middle;
+
 	cursor: pointer;
 	user-select: none;
 


packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -80,6 +80,7 @@ export default defineComponent<KottiButton.PropsInternal>({
 	display: inline-flex;
 	align-items: center;
 	justify-content: center;
+	contain: content;
```
FlorianWendelborn, line 83: 1. How did you find out about this? I've never seen this feature
2. This doesn't appear to be supported by Safari. Do you have an alternative solution? If no, then this is better than nothing

FlorianWendelborn (approved):

packages/kotti-ui/source/kotti-button/KtButton.vue
HansAuger: 1. Nothing too sophisticated, I just did highly uncoordinated experiments in the browser dev tools
2. Changed the fix to use `vertical-align`, that should have a bit of a broader support

HansAuger: I am not entirely sure why `vertical-align: middle` works (along with a few alternative values like `text-bottom` or `text-top`). Playing around with it I noticed that if you unset `display: inline-flex` for `.kt-button` the texts of the buttons would align. So I guess it's something to do with the order in which `vertical-align` and `flex` is calculated and drawn by browsers.

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -82,6 +82,7 @@ export default defineComponent<KottiButton.PropsInternal>({
 	justify-content: center;
 
 	font-weight: 600;
+	vertical-align: middle;
```
FlorianWendelborn, line 85: Since this was a pretty weird bug, I think it makes sense to add a small comment on why this is necessary

mvangent (approved):

PR #450, opened 2021-06-14
feat(kt-field-single-select-remote): 

HansAuger:


packages/kotti-ui/source/kotti-field-select/KtFieldSingleSelectRemote.vue
```diff
@@ -0,0 +1,170 @@
...
+			<KtField
+				v-bind="{ field }"
+				class="kt-field-select kt-field-select-single-remote"
+				:getEmptyValue="() => null"
+				:helpTextSlot="$slots.helpText"
+			>
+				<SingleSelectInput
+					:label="selectedLabel"
+					:placeholder="placeholder"
+					:queryString="queryValue"
```
mvangent, line 13: this is used for searching on the backend, correct? if so, how about making a debounce configurable as a prop to the SingleSelectInput, so that queryInputChanged emit is debounced and the remote is not hit when typing a word?

carsoli: I'd say that's on the user-app, not on Kotti. The component should remain purely controlled, it emits when it needs to, if the user-app wants to ignore the events, it debounces.

HansAuger: Agreeing with @carsoli , it is not necessarily true - although likely - that this component is used in an async context or that the search operation is costly.  Debouncing should  be handled by whoever is implementing this.

mvangent: yeah. I was assuming it is always true, and probably it will be. I agree it is a bell/whistle, but 99% of the times it is used, so why not make it dry and let the component handle it was and is my thinking. By the way, it does not mean a debounce is always used, default value can be that it is 0. It is not a blocker for me, just wanted to point out the user app is going to always implement a debounce from the outside now for a component that by it's definition seems to always need one.

FlorianWendelborn:
@mvangent can you please rebase instead of merging? Thanks

mvangent:
> @mvangent can you please rebase instead of merge? Thanks

@FlorianWendelborn certainly, mea culpa. I was simply hitting the button configured over here. Can it be set to rebase instead of merge if a branch needs to get up to date with master?

FlorianWendelborn:
> Can it be set to rebase instead of merge if a branch needs to get up to date with master?

@mvangent last time I checked, no. Otherwise, I assure you I’d already have configured it

FlorianWendelborn:
@mvangent this one would be possible to add, but it requires using `/rebase` instead of having a button: <https://github.com/marketplace/actions/automatic-rebase>

mvangent:
> @mvangent this one would be possible to add, but it requires using `/rebase` instead of having a button: https://github.com/marketplace/actions/automatic-rebase

@FlorianWendelborn , LGTM, let's add and use it?

packages/kotti-ui/source/kotti-field-select/KtFieldSingleSelectRemote.vue
carsoli: if all the usages on user-app would debounce, it's on user-app to implement either a utility (hook) that handles the emitted event, debounces, makes the BE call, passes back a flag that it's loading/fetching and then the negation when it's done, and the data that it receives (so the options) and in this case it would also disable the select while loading from accepting further query, if it has to.

All of this ^ is not to show off but to state that it's not going to be just the "debounce" behavior that's needed but rather a whole logical implementation, and this is more and more proof that it's fit to be done on the user app
A custom component instead of the utility can be implemented as well.

I still disagree with the debounce proposal, but don't mind if another argument can be raised that tilts my opinion in favor of this.
@mvangent

mvangent: Like I said, not a blocker, so please go ahead as you see fit, I trust you find the best solution @carsoli and I don't disagree with your argumentation. Debounce is just a good way of making sure the user has enough time to search for the term they are actually searching for imo. And since this component is called selectremote, and a need for delay seemed implied by definition I wanted to raise it as a point. I can see why you want to do it on the wrapping layer. Thanks for sharing your thinking on it.

carsoli: I can't deny that other libraries don't already make this mistake but in essence they defy the whole point of this being a controlled component fully when they fiddle with the logic internally, and the way I see it it's always going to be a source of bugs at one point.

I can take the responsibility on Yoda's side or one of you guys @HansAuger @FlorianWendelborn even, after the urgent change is made for internal order ofc

packages/kotti-ui/source/kotti-field-select/components/ActionIconNext.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+	<div
+		:class="classes"
+		role="button"
+		@mouseenter="hoverOnClearIcon = true"
+		@mouseleave="hoverOnClearIcon = false"
+	>
+		<i class="yoco" @click.stop="handleClear" v-text="icon" />
```
carsoli, line 8: we might want to not handleClear if it's not clear icon in the first place - testing and changing this myself if needed

packages/kotti-ui/source/kotti-popover/components/IconTextItem.vue
```diff
@@ -19,36 +21,48 @@ export default defineComponent<{
...
-@import '../../kotti-style/_variables.scss';
+@import '../../kotti-field/mixins';
 
 .kt-popover-options-item {
+	@include no-outline;
+
 	display: flex;
 	align-items: center;
-	padding: $unit-2;
+	padding: var(--unit-2);
```
carsoli, line 46: were those made into css variables, checking that this works

packages/kotti-ui/source/kotti-field-select/components/ActionIconNext.vue
carsoli: Will only execute `handleClear` if the `canClear` condition is met

it doesn't make a difference in the Functional behavior at all, it just prevents an extra execution when the field is empty and the clearIcon is not shown (so an optimization and potentially a guard against future bugs)

carsoli (approved):

carsoli:
i need one more approve from either @FlorianWendelborn or @HansAuger to merge

thank you

packages/kotti-ui/source/kotti-field-select/constants.ts
```diff
@@ -37,6 +37,12 @@ const KOTTI_FIELD_SELECT_PROPS = {
 	},
 }
 
+export const KOTTI_FIELD_REMOTE_SELECT_SUPPORTS: KottiField.Supports = {
+	clear: true,
+	decoration: true,
+	tabIndex: true, // TODO:verify that this will work
```
FlorianWendelborn, line 43: TODO spotted

packages/kotti-ui/source/kotti-field-select/constants.ts
```diff
@@ -54,3 +60,13 @@ export const KOTTI_FIELD_MULTI_SELECT_PROPS = {
 		type: Number,
 	},
 }
+
+export const KOTTI_FIELD_REMOTE_SINGLE_SELECT_PROPS = {
+	...KOTTI_FIELD_SELECT_PROPS,
+	// TODO: update docs to show this (add to additional props)
```
FlorianWendelborn, line 66: I think this was fixed already

packages/kotti-ui/source/kotti-popover/components/IconTextItem.vue
```diff
@@ -3,12 +3,14 @@
 		class="kt-popover-options-item"
 		:class="{
 			'kt-popover-options-item--is-disabled': isDisabled,
+			'kt-popover-options-item--is-selected': isSelected,
 		}"
 		:data-test="dataTest"
+		:tabindex="isDisabled ? -1 : 0"
```
FlorianWendelborn, line 9: Does `-1` work? Otherwise, I'd prefer `undefined`

FlorianWendelborn (approved):
Pretty much flawless now, I like it

packages/kotti-ui/source/kotti-popover/components/IconTextItem.vue
carsoli: https://www.techonthenet.com/html/elements/global_attributes.php#:~:text=tabindex,reachable%20through%20sequential%20keyboard%20navigation

carsoli: if I set it to be `undefined`, the element is not focusable, then the `event.relatedTarget` that's captured by the `blur` `event` from the `input` element (see https://github.com/3YOURMIND/kotti/pull/450/files#diff-e7f6fccf48a52fe6a4fffaa684158edc9d0c5a85bb8c153a41f959505889ebfaR182) is the component that is the parent of the tippy content so the function  returns `false` and so we close the dropdown

resolving, will leave as is.

PR #452, opened 2021-06-15
fix(Kt-Field-*-Select): hideClear actionIcon not showing

carsoli:
we needed to capture `mouseenter` and `mouseleave` events for the Icon to show on hovering on the action icon.
### Details: 
With `KtFieldSingleSelect` and `KtFieldMultiSelect`, we overwrote the `actionIcon` slot because not only do we have a `clear` icon that's interactive and empties the field but also chevrons that change depending on the state of visibility of the popup.
To achieve this, we captured `mouseenter/mouseleave` events, and if the `hideClear` was false, we showed the clear on hover 
> hideClear is only false when the field is not "empty" 

The mistake must have been that someone found the `mouseenter/mouseleave` pointless at one point and removed them b/c this used to work (that someone could have very well been me)

Now that `KtField` captures those events, we can see the clear behavior again

### Notes: 
There still is a problem with element-ui's clear icon showing - out of scope


FlorianWendelborn:
Turns out this doesn’t really fix anything. Closing as discussed with @carsoli.

PR #456, opened 2021-07-09
Refactor kt accordion

HansAuger:
Nothing too fancy

Breaking change, css classes rename from `accordion*` to `kt-accordion*`

### DIFF
--- a/packages/kotti-ui/source/kotti-accordion/KtAccordion.vue
+++ b/packages/kotti-ui/source/kotti-accordion/KtAccordion.vue
@@ -15,12 +15,7 @@
 				<i class="yoco" v-text="isOpen ? 'minus' : 'plus'" />
 			</div>
 		</div>
-		<div
-			class="accordion__content"
-			:class="isOpen ? 'is-open' : 'is-close'"
-			:style="`--height: ${contentHeight}px`"
-			@click="setHeight"
-		>
+		<div class="accordion__content" :class="isOpen ? 'is-open' : 'is-close'">
 			<div ref="contentInner" class="inner">
 				<slot />
 			</div>
@@ -28,8 +23,12 @@
 	</div>
 </template>
 
-<script>
-export default {
+<script lang="ts">
+import { defineComponent, ref } from '@vue/composition-api'
+
+import { KottiAccordion } from './types'
+
+export default defineComponent<KottiAccordion.PropsInternal>({
 	name: 'KtAccordion',
 	props: {
 		icon: { default: null, type: String },
@@ -37,37 +36,21 @@ export default {
 		isFullyClickable: { default: false, type: Boolean },
 		title: { required: true, type: String },
 	},
-	data() {
+	setup(props) {
+		const isOpen = ref(!props.isClosed)
+
+		const toggle = () => {
+			isOpen.value = !isOpen.value
+		}
 		return {
-			isOpen: true,
-			contentHeight: null,
-			interval: null,
+			headerToggle: () => {
+				if (props.isFullyClickable) toggle()
+			},
+			isOpen,
+			toggle,
 		}
 	},
-	mounted() {
-		this.isOpen = !this.isClosed
-		// This is the best solution right now
-		// It will handle the case when the content change
-		const reloadingTime = 500
-		this.interval = setInterval(this.setHeight, reloadingTime)
-	},
-	beforeDestroy() {
-		clearInterval(this.interval)
-	},
-	methods: {
-		setHeight() {
-			const height = this.$refs.contentInner.clientHeight
-			this.contentHeight = height
-		},
-		headerToggle() {
-			if (this.isFullyClickable) this.toggle()
-		},
-		toggle() {
-			this.setHeight()
-			this.isOpen = !this.isOpen
-		},
-	},
-}
+})
 </script>
 
 <style lang="scss">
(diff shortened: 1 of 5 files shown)


packages/kotti-ui/source/kotti-accordion/hooks.ts
```diff
@@ -0,0 +1,53 @@
+import { ref, Ref } from '@vue/composition-api'
```
FlorianWendelborn, line 1:
```suggestion
import { ref, Ref } from '@vue/composition-api'

```

packages/kotti-ui/source/kotti-accordion/KtAccordion.vue
```diff
@@ -1,145 +1,138 @@
...
 		title: { required: true, type: String },
 	},
-	data() {
+	setup(props) {
+		const contentInner = ref<HTMLElement | null>(null)
+
+		const { isContentOpen, toggle } = useSlideAnimaion(contentInner, {
+			duration: 300,
+			isInitiallyClosed: props.isClosed,
+		})
```
FlorianWendelborn, line 43:
```suggestion
		})

```

packages/kotti-ui/source/kotti-accordion/KtAccordion.vue
```diff
@@ -1,145 +1,138 @@
...
+
+import { useSlideAnimaion } from './hooks'
+import { KottiAccordion } from './types'
+
+export default defineComponent<KottiAccordion.PropsInternal>({
 	name: 'KtAccordion',
 	props: {
 		icon: { default: null, type: String },
 		isClosed: { default: false, type: Boolean },
 		isFullyClickable: { default: false, type: Boolean },
```
FlorianWendelborn, line 37: I think we should actually drop support and make this the default. From what I can tell, there's no good reason why we wouldn't want to always have this set to true. Right now, AFAIK, it's unused as well…

packages/kotti-ui/source/kotti-accordion/KtAccordion.vue
```diff
@@ -1,145 +1,138 @@
...
+				isContentOpen.value ? Yoco.Icon.MINUS : Yoco.Icon.PLUS,
+			),
 		}
 	},
-	mounted() {
-		this.isOpen = !this.isClosed
-		// This is the best solution right now
-		// It will handle the case when the content change
-		const reloadingTime = 500
-		this.interval = setInterval(this.setHeight, reloadingTime)
```
FlorianWendelborn, line 52: Why is this no longer necessary?

packages/kotti-ui/source/kotti-accordion/KtAccordion.vue
```diff
@@ -1,145 +1,138 @@
...
-.accordion__title__icon,
-.accordion__title__text {
-	display: inline-block;
-}
+	&__content {
+		margin-top: -1px; // prevent border overlap when closed
+		overflow: hidden;
+		border: 1px solid var(--ui-02);
+		border-top: none;
+		transition: height 200ms linear;
```
FlorianWendelborn, line 115: This might conflict with the animation stuff you did, but not sure

FlorianWendelborn (approved):
CR looks pretty good and readable, didn't FT

packages/kotti-ui/source/kotti-accordion/KtAccordion.vue
carsoli: double this opinion

packages/kotti-ui/source/kotti-accordion/hooks.ts
```diff
@@ -0,0 +1,53 @@
...
+	easing: 'ease',
+	fill: 'forwards',
+	isInitiallyClosed: true,
+})
+
+export const useSlideAnimaion = (
+	element: Ref<HTMLElement | null>,
+	options: Options,
+) => {
+	const { isInitiallyClosed, ...restOptions } = options
```
carsoli, line 17: i think u can just skip this entire body if element is null

so u can start with an `if` condition here, if it's null, just return `{ isContentOpen: false, toggle: () => {})`

instead of all `element.value?.` you got there

packages/kotti-ui/source/kotti-accordion/hooks.ts
```diff
@@ -0,0 +1,53 @@
...
+	const down = async (): Promise<void> => {
+		isContentOpen.value = true
+		await triggerAnimation(true)
+	}
+
+	const toggle = (): Promise<void> => {
+		return isContentOpen.value ? up() : down()
+	}
+
+	return { up, down, toggle, isContentOpen }
```
carsoli, line 52: why do u return `up` and `down`

packages/kotti-ui/source/kotti-accordion/KtAccordion.vue
HansAuger: This one is debatable. What this does in the current version of the Accordion, now every time the content height of the content changes this will trigger an animation. I personally dislike both the visual feel of that an the implementation. Now a content height change is will be drawn immediately without any transition.

Change my mind

FlorianWendelborn: IIRC, this code was never about any fancy animation or whatever. It was because the max height was previously specified due to the need to interpolate between 0 and height for the opening.

So, basically, without this code, when the content expands it just wouldn’t expand the accordion.

FlorianWendelborn:
@carsoli can you approve & merge?

PR #457, opened 2021-07-09
fix(KtFilters): Don't mutate isAddingFilter prop

santiagoballadares:


FlorianWendelborn:
Does the parent already handle this emit or why isn't there a second file in this PR @santiagoballadares?

FlorianWendelborn (approved):

carsoli (approved):

Issue #454, opened 2021-06-28
Make KtAvatar small per default in KtComment


kpm-berlin:
**Is your feature request related to a problem? Please describe.**
The new design rule for all comments is a small avatar size.

**Describe the solution you'd like**
Add small property `:size="sm"` to `KtAvatar` in `KtComment`

**Describe alternatives you've considered**
Add a prop `avatarSize` to`KtComment` which is passed to `KtAvatar`

**Additional context**
A similar component for history will be developed soon that shall have the same avatar size (small).

PR #459, opened 2021-07-12
fix(454): remove avatar class, flex reply icons inline with button/text

kpm-berlin:
Follow-up PR to #454 

Changelog:

- remove `comment__avatar` class
- flex reply button text and icon in-line
- flex reply message and icon in-line

### DIFF
--- a/packages/kotti-ui/source/kotti-comment/KtComment.vue
+++ b/packages/kotti-ui/source/kotti-comment/KtComment.vue
@@ -1,6 +1,6 @@
 <template>
 	<div class="comment">
-		<KtAvatar class="comment__avatar" size="sm" :src="userAvatar" />
+		<KtAvatar size="sm" :src="userAvatar" />
 		<div class="comment__wrapper">
 			<div class="comment__info">
 				<div class="info__name" v-text="userName" />
@@ -157,3 +157,10 @@ export default {
 	},
 }
 </script>
+
+<style lang="scss" scoped>
+.action__reply {
+	display: flex;
+	align-items: center;
+}
+</style>
--- a/packages/kotti-ui/source/kotti-comment/components/CommentReply.vue
+++ b/packages/kotti-ui/source/kotti-comment/components/CommentReply.vue
@@ -111,3 +111,9 @@ export default {
 	},
 }
 </script>
+<style lang="scss" scoped>
+.comment-reply__message {
+	display: flex;
+	align-items: center;
+}
+</style>


packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 	<div class="comment">
```
carsoli, line 2: absolutely not on you but could u also add `kt-` prefix to this class, we r moving towards this in all our refactors, since this should be distinguished as a kotti class, and we will add this as a breaking change in the followup release, in case some user app relies on this

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -1,6 +1,6 @@
...
-		<KtAvatar class="comment__avatar" size="sm" :src="userAvatar" />
+		<KtAvatar size="sm" :src="userAvatar" />
 		<div class="comment__wrapper">
```
carsoli, line 4: accordingly `kt-comment__wrapper`

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -1,6 +1,6 @@
...
 			<div class="comment__info">
```
carsoli, line 5: `kt-comment__info`

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -1,6 +1,6 @@
...
 				<div class="info__name" v-text="userName" />
```
carsoli, line 6: `kt-comment__info__name`

carsoli: i can't comment further below but `kt-comment__info__time`

carsoli: and `kt-comment__message`

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -157,3 +157,10 @@ export default {
 	},
 }
 </script>
+
+<style lang="scss" scoped>
+.action__reply {
```
carsoli, line 162: as with the class restructuring i'd propose, you also look into how you can make this follow the BEM convention, so it's not `action__reply` but whatever it is in the hierarchy of `kt-comment`

carsoli:
merging this for now

the class restructuring to prefix with `kt-` and also follow BEM
as well as using `KtFieldTextArea` instead of `text-area` with `form-group` classes and `form-input` classes
can follow up in a separate tech PR

carsoli (approved):

PR #473, opened 2021-07-21
fix(KtNavbar): sections' links should support new tab opening

carsoli:
the current usage in the docs and our user app still relies on the router, so the fix is still not in place fully
but assuming someone wants to use those a-tags directly as links, that should still be allowed


packages/kotti-ui/source/kotti-navbar/components/NavbarMenu.vue
```diff
@@ -12,6 +12,8 @@
 				class="kt-navbar-menu__item"
 				:class="{ active: link.isActive, narrow: isNarrow }"
 				:href="link.link ? link.link : null"
+				rel="noopener noreferrer"
+				target="_blank"
```
FlorianWendelborn, line 16: Wouldn't that **always** open in new tab?

FlorianWendelborn:
Superseded by #557. Closing.

PR #475, opened 2021-08-06
ci(unimported): Auto-detect Dead Code

FlorianWendelborn:
FYI @smeijer here’s one of the public repos that’ll use `unimported`. Majority will be in private repos though


packages/kotti-ui/tsconfig.json
```diff
@@ -12,7 +12,6 @@
 		"strict": true,
 		"jsx": "react",
 		"jsxFactory": "h",
-		// FIXME: remove skipLibCheck when [https://github.com/ElemeFE/element/issues/17329] is resolved
```
FlorianWendelborn, line 15: @smeijer this is perhaps something that could be fixed in `unimported`. It doesn’t seem to support comments in JSON. I know, not technically allowed in the JSON standard, however neither rollup, typescript, vscode, etc. had any issues with this. Removed it for `unimported` for now as it’s not a big deal in this particular case

smeijer: Thanks. I'll fix that in a future release. It's being tracked via this issue. https://github.com/smeijer/unimported/issues/49#issuecomment-894072532

carsoli (approved):

PR #483, opened 2021-08-31
feature(#481): Add maximumSelectable Prop to KtFieldMultiSelect

### LINKED ISSUE #481: KtFieldMultiSelect: `maximumSelectableOptions` prop
Some of our multi-select fields want to be able to set a limit on the selection of choices. 
In real-life: this would happen, if you have a form, where you ask for feedback about the top 3 choices for some property, and you wouldn't want your users to be able to select more.

**Describe the solution you'd like**
Add a new prop for `KtFieldMultiSelect` that sets a limit on the choices, and give it a clear name.

**Describe alternatives you've considered**
None

**Additional context**
There should be some indication for the user that there is a maximum of allowed selectables.

FlorianWendelborn:
Closes #481

### DIFF
--- a/packages/kotti-ui/source/kotti-field-select/KtFieldMultiSelect.vue
+++ b/packages/kotti-ui/source/kotti-field-select/KtFieldMultiSelect.vue
@@ -15,6 +15,7 @@
 			<ElSelect
 				ref="elSelectRef"
 				v-bind="elMultipleSelectProps"
+				:multipleLimit="maximumSelectable"
 				@input="onChange"
 				@visible-change="(showPopper) => (isDropdownOpen = showPopper)"
 			>
--- a/packages/documentation/pages/usage/components/form-fields.vue
+++ b/packages/documentation/pages/usage/components/form-fields.vue
@@ -1,309 +1,367 @@
-<template lang="md">
-<ComponentInfo v-bind="{ component }" />
+<template>
+	<div>
+		<ComponentInfo v-bind="{ component }" />
 
-<KtTranslationContext :locale="settings.locale">
-	<div class="overview">
-		<div class="overview__component">
-			<h4>Component</h4>
-			<KtForm v-model="values">
-				<component
-					:is="componentRepresenation.name"
-					:validator="componentRepresenation.validator"
-					v-bind="componentRepresenation.props"
-					@update:query="updateQuery"
-				>
-					<div slot="helpText" v-if="componentRepresenation.hasHelpTextSlot">
-						<div>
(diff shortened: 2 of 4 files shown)


packages/kotti-ui/source/kotti-field-select/KtFieldMultiSelect.vue
```diff
@@ -95,6 +96,7 @@ export default defineComponent({
 	props: {
 		...KOTTI_FIELD_PROPS,
 		...KOTTI_FIELD_MULTI_SELECT_PROPS,
+		maximumSelectable: { default: Number.MAX_SAFE_INTEGER, type: Number },
```
carsoli, line 99: this belongs on `KOTTI_FIELD_MULTI_SELECT_PROPS` though

carsoli: Also, I don't agree with the `MAX_SAFE_INTEGER` either for the default value  
this should be the maximum of the options.length which can't be added to the default inline defintion obviously  
so we most likely need a computed `defaultMaximumSelectable` to be calculated always off the `options.length`  as a fallback to `null` default

1) default becomes `() => null` 
2) we have a `defaultMaximumSelectable` computed off options.length
3) what we end up passing to `element-ui` is `props.maximumSelectable ?? defaultMaximumSelectable.value`

carsoli: https://github.com/ElemeFE/element/blob/655d89fe0e0d9c4c7c4ca2541f5cf90437690a76/packages/select/src/option.vue#L74
if u look at this computed from their source code, they disable the code when the `limitIsReached` which should be the options.length actually in our case IMO

carsoli: I also noticed we had the same default for `collapseTagsAfter` and i don't know if i want to change that because this is internal logic from Kotti's side and doesn't matter really if it can be MAX INT at one point

carsoli: alternatively, don't pass it at all if it's not passed to Kotti - so v-bind to `undefined`

carsoli: or ignore this entire thing, ur call - not a blocker 

My fear is future updates in which we will miss a weird side effect from their source code

carsoli (approved):

PR #492, opened 2021-09-04
Refactor kt drawer

HansAuger:
typescript and composition api

Breaking Changes:
- prop `wide` renamed to `isWide`
- switched meaning of boolean prop `closeOutside` (default: `true`) to `disallowCloseOutside`. That means, old usages of this must be changed in the following way:
```
:closeOutside="false" ~> disallowCloseOutside
```


carsoli:
picking this up

packages/kotti-ui/source/kotti-accordion/KtAccordion.vue
```diff
@@ -46,8 +46,6 @@ export default {
 	},
 	mounted() {
 		this.isOpen = !this.isClosed
-		// This is the best solution right now
```
carsoli, line 49: what is this change related to @HansAuger 
can we discard this

packages/kotti-ui/source/kotti-drawer/KtDrawer.vue
```diff
@@ -1,203 +1,195 @@
...
-	align-self: flex-end;
-	width: 100%;
-}
+	/* Vue Transition classes */
 
-.drawer-enter {
-	opacity: 0;
-	transition: opacity 0.5s;
-}
+	&-enter {
```
carsoli, line 149: is this why we can't add `scoped` to the style sheet

can we `::v-deep` on each of the sub-classes (won't work with scss syntax) instead?

carsoli (approved):

FlorianWendelborn (approved):
CR LGTM

packages/kotti-ui/source/kotti-drawer/KtDrawer.vue
```diff
@@ -1,203 +1,198 @@
...
-		},
-		defaultWidth: {
-			type: String,
-			default: null,
-		},
-		expandWidth: {
-			type: String,
-			default: null,
-		},
+		disallowCloseOutside: { type: Boolean, default: false },
```
carsoli: @HansAuger I actually thought about this again

I'd say, just make it `closeOutside` and `default` as `false`

If any case, we are breaking the API, might as well just replace the usages of the accordion where `closeOutside` wasn't mentioned, to explicitly mention it, and replace the other with `false`

@FlorianWendelborn any opinions?

FlorianWendelborn: I think the default should be whatever behavior we prefer. I think if there's no special reason to disallow close-outside, it should be turned on

Maybe the wording can be changed somehow else (e.g. associated to that the user doesn't want to accidentally cancel here)

packages/kotti-ui/source/kotti-accordion/KtAccordion.vue
FlorianWendelborn: I’m gonna discard this so we can move on

packages/kotti-ui/source/kotti-drawer/KtDrawer.vue
FlorianWendelborn: It’s about the `:root`. I changed it to only make that part non-scoped

PR #494, opened 2021-09-07
feature(#285): Add Zod to Massively Improve Validation

### LINKED ISSUE #285: Add Prop Validation EVERYWHERE
Just saw a `KtCol span="2.5"`. We absolutely need validators for everything.

- [x] Validate Numbers to be Integers if Required
- [x] Validate With Whitelists
- [x] Validate Yoco Icons

~~I think at this point it makes sense to create a `@3yourmind/vue-validators` package or similar.~~

This will now use `zod`

FlorianWendelborn:
Closes #285

---

⚠️ **Remember to publish a new yoco version once this is merged and ready for publish**


HansAuger:
lgtm so far. Not all components got a redo, but we can do that  incrementally. Approved

HansAuger (dismissed):

packages/kotti-ui/source/kotti-button-group/types.ts
```diff
@@ -1,7 +1,7 @@
 import { SpecifyRequiredProps } from '../types/utilities'
 
 export namespace KottiButtonGroup {
-	export type PropsInternal = {}
+	export type PropsInternal = Record<string, never>
```
HansAuger, line 4: I rebased on current master and this line was making trouble. Apparently `{}` is a deprecated way to declare an empty object acording to `eslint`. Maybe something that's new in TS 4?

FlorianWendelborn: Look up the rule, it's because {} or object basically fit literally anything in JS. I'm not sure what the best way to declare an empty object is, but usually it isn't actually an empty object and instead is a record or an object with lots of optional values

FlorianWendelborn: I think the `Record<string, never>` is the best way to type it BTW. Good find.

packages/kotti-ui/source/kotti-user-menu/types.ts
```diff
@@ -1,19 +1,25 @@
...
+	export type Section = z.infer<typeof sectionSchema>
+
+	export const propsSchema = z.object({
+		sections: z.array(sectionSchema),
+		userAvatar: z.string().nullable().default(null),
+		userName: z.string().nullable().default(null),
+		userStatus: z.string(),
+	})
+
+	export type PropsInternal = z.input<typeof propsSchema>
```
FlorianWendelborn, line 23: output and input are swapped here

FlorianWendelborn:
Well, it’s not really a draft, it’s just that I found some ways to improve it while I’m waiting for @carsoli’s review :(

PR #504, opened 2021-09-21
add-zod(5): Add makeProps

FlorianWendelborn:
Rebased on top of #505, #506, #507, and #508

---

## Breaking Changes:

- `Kotti.UserMenu.Section` now enforces `title` as `string | null`


packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,54 @@
+import { cloneDeep } from 'lodash'
+import { z } from 'zod'
+
+/**
+ * Easily validate vue props, improve error messages drastically
+ *
+ * This augments/replaces Vue’s prop.validator approach by using zod and therefore actually
+ * prints errors to the console that can actually help developers rather than just say
+ * "hey you failed the validator, good luck!" as vue would do by default.
```
carsoli: rather than the non-distinct `vue-warn` message about a validator failing for a component.

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,54 @@
+import { cloneDeep } from 'lodash'
```
carsoli, line 2: do u think the file can be called prop-validator or prop-utils instead?

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,54 @@
...
+	const validator = (value: unknown) => {
+		const result = schema.pick({ [key]: true }).safeParse({ [key]: value })
+
+		if (result.success) return true
+
+		/* eslint-disable no-console */
+		console.group(`propValidator found issues with prop “${key}”`)
+
+		// HACK: 'error' in result is necessary as `ts-jest` doesn’t see that result.success was already properly checked to be falsy and that error now exists
+		if ('error' in result) console.error(result.error)
```
carsoli, line 23: It's a stylistic preference

but now when I have those if-conditions particularly for ts not seeing proper types because they are not asserted and are evaluated at runtime

what I do is:

`console.error((result as Exclude<...>).error)`

if you can achieve a similar type asserion, then fix the type of the result with an assertion right below the check for success

~
ok apparently if you type assert, eslint/ts will remove what it redeems as an unnecessary assertion according to the compiler

packages/kotti-ui/source/kotti-user-menu/KtUserMenu.vue
```diff
@@ -55,32 +55,38 @@ import { mixin as clickaway } from 'vue-clickaway'
 
 import { KtAvatar } from '../kotti-avatar'
 import { IS_NAVBAR_NARROW } from '../kotti-navbar/constants'
+import { propValidator } from '../props'
 import { Kotti } from '../types'
 
 import { KottiUserMenu } from './types'
 
-const linkIsValid = (link: KottiUserMenu.SectionLink) => Boolean(link.title)
```
carsoli, line 62: who wrote those validator expressions? ok, apparently you.

you assumed the type without pre-checking if it's not `unknown` - so for example by having another `Object.prototype.call`

packages/kotti-ui/source/kotti-user-menu/types.ts
```diff
@@ -1,16 +1,27 @@
-import { KottiNavbar } from '../kotti-navbar/types'
+import { z } from 'zod'
+
 import { SpecifyRequiredProps } from '../types/utilities'
 
 export namespace KottiUserMenu {
-	export type SectionLink = KottiNavbar.SectionLink
-	export type Section = KottiNavbar.Section
+	export const sectionLinkSchema = z.object({
+		title: z.string(),
```
carsoli, line 5: according to the usage

there is `link.link` so add `link: z.string()`

carsoli: apparently there is a handler for external control of the click event, so it's `string | undefined`

packages/kotti-ui/source/kotti-user-menu/types.ts
```diff
@@ -1,16 +1,27 @@
...
+	})
+	export type SectionLink = z.infer<typeof sectionLinkSchema>
+
+	export const sectionSchema = z.object({
+		links: z.array(sectionLinkSchema).nonempty(),
+		title: z.string().nullable(),
```
carsoli, line 11: does `nullable()` imply `null | undefined`

carsoli: I think the user-app used to pass `undefined` for a title

and `nullable` apparently makes it `| null`

so it's a breaking change, we need this as a note to make changes in the user-app to avoid console errors by a kotti update

packages/kotti-ui/source/kotti-user-menu/types.ts
```diff
@@ -1,16 +1,27 @@
...
+	export type Section = z.infer<typeof sectionSchema>
 
-	export type PropsInternal = {
-		sections: Array<KottiUserMenu.Section>
-		userAvatar: string | null
-		userName: string | null
-		userStatus: string
-	}
+	export const propsInternalSchema = z.object({
+		sections: z.array(sectionSchema),
```
carsoli, line 17: do u maybe wna also make this `.nonEmpty()`

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
```diff
@@ -55,7 +55,7 @@
 import { computed, defineComponent, provide, ref } from '@vue/composition-api'
 import { mixin as clickaway } from 'vue-clickaway'
 
-import { Kotti } from '../types'
+import { propValidator } from '../props'
 
 import NavbarLogo from './components/NavbarLogo.vue'
 import NavbarMenu from './components/NavbarMenu.vue'
```
carsoli, line 61: can't comment on the line below

create an issue to delete `NavbarNotification` and the prop `notifications` on the `KtNavbar` 
as part of the next published release of Kotti

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
```diff
@@ -74,15 +74,35 @@ export default defineComponent<KottiNavbar.PropsInternal>({
...
+			default: null,
+			type: Object,
+			validator: propValidator(KottiNavbar.propsInternalSchema, 'notification'),
+		},
+		quickLinks: {
+			default: () => [],
+			type: Array,
+			validator: propValidator(KottiNavbar.propsInternalSchema, 'quickLinks'),
+		},
+		sections: {
```
carsoli: note to self

sections structure

```
Array<
 {
  title: string (not optional), 
  links: Array<{
      icon: Yoco (not optional), 
      title: string (not optional), 
      link: string | undefined, 
      isActive: boolean
     }>
  }
>
```

packages/kotti-ui/source/kotti-navbar/types.ts
```diff
@@ -1,39 +1,52 @@
...
+		link: z.string(),
+		title: z.string(),
+	})
+	export type QuickLink = z.infer<typeof quickLinkSchema>
+
+	export const sectionLinkSchema = z.object({
+		icon: yocoIconSchema,
+		isActive: z.boolean(),
+		title: z.string(),
+		to: z.unknown().optional(),
```
carsoli, line 22: All the usages in the user-app rely on a `to` attribute on the section
> but it's internally used on the click handler that we eventually pass to `KtNavbar`
> so there is no breaking usage

HOWEVER, exposing the `to` will confuse users who inspect the type not the actual implementation

SO:
1) remove the `to` 
2) introduce the `link` prop which is a `string | undefined` that we bind to `href` in `NavbarMenu.vue`

The user either uses the `link` and relies on our `href` implementation, or they have whatever internal API that they like and exclusively use the click handler

This comment can be resolved with this:
https://github.com/3YOURMIND/kotti/pull/504/commits/257ea964d015d82203c72e3631fc4b40b316e228#r713940508

packages/kotti-ui/source/kotti-navbar/types.ts
```diff
@@ -1,39 +1,52 @@
...
+	})
+	export type SectionLink = z.infer<typeof sectionLinkSchema>
+
+	export const sectionSchema = z.object({
+		title: z.string().nullable(),
+		links: z.array(sectionLinkSchema),
```
carsoli, line 28: an extra check could be the `nonempty()`

carsoli: it's otherwise unfruitful to use this component? 🤷🏼

packages/kotti-ui/source/kotti-button/types.ts
```diff
@@ -10,23 +11,26 @@ export namespace KottiButton {
...
-		isBlock: boolean
-		isLoading: boolean
-		isMultiline: boolean
-		isSubmit: boolean
-		label: string | null
-		size: KottiButton.Size
-		type: KottiButton.Type
-	}
+	export const propsInternalSchema = z.object({
+		icon: yocoIconSchema,
```
carsoli: `.nullable()`

packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
```diff
@@ -25,47 +25,34 @@
...
-				value: any,
-			): value is KottiBreadcrumb.Props['separator'] =>
-				value !== null &&
-				typeof value === 'object' &&
-				((value.style === KottiBreadcrumb.SeparatorType.ICON &&
-					isYocoIcon(value.value)) ||
-					(value.style === KottiBreadcrumb.SeparatorType.TEXT &&
-						isString(value.value))),
+			type: Object,
+			validator: propValidator(
```
carsoli: followup PR  

consider deleting the way we define the `separator` prop and simplifying it into just a text-separator

packages/kotti-ui/source/kotti-line/types.ts
```diff
@@ -1,9 +1,12 @@
+import { z } from 'zod'
+
 import { SpecifyRequiredProps } from '../types/utilities'
 
 export namespace KottiLine {
-	export type InternalProps = {
-		text: string
-	}
+	export const propsInternalSchema = z.object({
+		text: z.string(),
```
carsoli: `.nullable()`

packages/kotti-ui/source/props.ts
carsoli: still suggest this even after being in the last commit that touches it

packages/kotti-ui/source/props.ts
```diff
@@ -1,29 +1,16 @@
...
- * }
- */
-export const propValidator = <
-	SCHEMA extends z.ZodObject<z.ZodRawShape>,
-	KEY extends keyof z.infer<SCHEMA>,
->(
+const DEBUG_MAKE_PROPS = false as const // enable to print debug log
+const DEBUG_WALK_SHAPE = false as const // enable to print debug log
+
+const propValidator = <SCHEMA extends z.ZodTypeAny>(
```
carsoli, line 10: ok so previously the schema was `z.ZodObject<X>`
where X was `z.ZodRawShape`

type declaration for raw shape is that it's an object of values of type `ZodTypeAny`:  
```
export declare type ZodRawShape = {
    [k: string]: ZodTypeAny;
};
```

HOW?

I would have understood if 
1) previously schema was 
`z.ZodZodObject<z.ZodTypeAny>`
2) now it became
`z.ZodRawShape`

none of this makes sense

I kind thing it should be simply
`SCHEMA extends z.ZodRawShape`

packages/kotti-ui/source/props.ts
```diff
@@ -52,3 +39,137 @@ export const propValidator = <
 
 	return validator
 }
+
+// eslint-disable-next-line sonarjs/cognitive-complexity
+const walkShape = <SHAPE extends z.ZodTypeAny>(shape: SHAPE): Set<string> => {
```
carsoli, line 46: how is this not rawShape 😕

packages/kotti-ui/source/props.ts
```diff
@@ -52,3 +39,137 @@ export const propValidator = <
...
+	}
+
+	return types
+	/* eslint-enable no-console */
+}
+
+type MapValueType<T> = T extends Map<unknown, infer V> ? V : never
+
+const ignoredZodTypes = new Set(['ZodDefault', 'ZodNullable', 'ZodOptional'])
+const zodToVueTypes = new Map<
```
carsoli, line 90: `zodToVueSingleType`

packages/kotti-ui/source/props.ts
```diff
@@ -52,3 +39,137 @@ export const propValidator = <
...
+	string,
+	Exclude<PropType<unknown>, Array<unknown>>
+>([
+	['ZodArray', Array],
+	['ZodBoolean', Boolean],
+	['ZodFunction', Function],
+	['ZodNativeEnum', String], // educated guess
```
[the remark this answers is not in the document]
carsoli: please clarify that we only support `string` enums then

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,198 @@
...
+	['ZodString', String],
+])
+
+/**
+ * @summary Easily define & validate vue props, improve error messages drastically
+ *
+ * @description
+ * This augments/replaces Vue’s props entirely, including the need to add prop.validator
+ * by using zod and therefore actually prints errors to the console that can help developers
+ * rather than just say "hey you failed the validator, good luck!" as vue would do by default.
```
carsoli, line 109: there was some comment on an inital commit on this line, with a different wording, feel free to use it or not.

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * ## Known limitations:
+ *
+ * 1. Deeply defined defaults are only applied when manually `schema.safeParse`-ing the data (only top-level defaults are possible to transform to Vue)
+ * 2. While this function also automatically types the return types via as PropType<> (which in theory
```
carsoli, line 114: via "as PropType<>"
or 
via asserting to PropType<>

carsoli: 2. This function automatically types the return types of props using "as `PropType<>`" (which, in turn, makes the annotation of Vue props unnecessary). However, PropType<> appears to be bugged, as tested in @vue-composition/api (versions: 0.6.1 & 1.1.5).

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,198 @@
...
+ *   should mean that annotating the prop types explicitly is no longer necessary) in `@vue/composition-api`
+ *   (`0.6.1` and `1.1.5`), `PropType<>` appears to be bugged.
+ *   For example KtBreadcrumb’s props get inferred to be any
+ *   So, unfortunately, this may only help in the future, once the Vue team fixes this
+ * 3. `z.function().default()` does not accept `undefined` while parsing the schema manually
```
carsoli, line 119:
```
3. `z.safeParse()` fails for type: `z.function()` if we specify the default, without explicitly chaining `optional()` on the function type. See {@link https://github.com/colinhacks/zod/issues/647}
Nevertheless, it is unconsequential for Vue prop validation, because Vue doesn't execute the `validator` if the prop is not passed (i.e. is undefined).
```

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,198 @@
...
+ * // types.ts
+ * export namespace KottiUserMenu {
+ *   const propsSchema = z.object({
+ *     // ...
+ *   })
+ *   type Props = z.input<typeof propsSchema>
+ *   type PropsInternal = z.output<typeof propsSchema>
+ * }
+ */
+export const makeProps = <SCHEMA extends z.ZodObject<z.ZodRawShape>>(
```
carsoli, line 139: I think I can verify that:
1) `z.ZodObject<z.ZodRawShape>` is equivalent to
2) `z.ZodRawShape`

in our case, because this is somehow only useful if we use recursive schemas IMU

carsoli: from all the prop type defintions, you're using `z.object` constructor, so I will not tire myself with understanding this further

since an object constructor simply should return a `ZodObject` as a wrapping type for each of the props type 

nevertheless:

the generic passed to the object being a rawshape, means that each of the props is an object and not a direct type/(schema in the terms of zod)


will see how it's used further then resolve this comment if fit

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,198 @@
...
+	schema: SCHEMA,
+): {
+	[KEY in keyof SCHEMA['shape']]: Omit<PropOptions, 'required' | 'type'> & {
+		required: z.output<SCHEMA>[KEY] extends undefined ? false : true
```
carsoli, line 143: Florian: `z.input` not `z.output`

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,198 @@
...
+		type: PropType<z.output<SCHEMA>[KEY]>
```
carsoli, line 144: same ☝🏼

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,198 @@
...
+			 * unintended side-effects
+			 */
+			const isOptional =
+				zodTypeSet.has('ZodDefault') || zodTypeSet.has('ZodOptional')
+
+			const prop: PropOptions<unknown, boolean> = {
+				validator: propValidator(shape, key),
+			}
+
+			if (isOptional) prop.default = shape._def.defaultValue
```
carsoli, line 169: check if zod adds the `undefined` by default to the `default` attribute
```
const i = z.string().optional() 
```
and i don't use a default 

i expect
```i._def.defaultValue.toBe(undefined)```

File: packages/kotti-ui/source/props.ts [hunk 26]
```diff
@@ -0,0 +1,198 @@
...
+ * These types do not influence the generation of Vue’s `propName.type` Arrays and are therefore discarded
+ */
+const ignoredZodTypes = new Set(['ZodDefault', 'ZodNullable', 'ZodOptional'])
+
+/**
+ * This maps the internal zod name of a type to the the constructor that Vue expects in `propName.type`
+ */
+const zodToVueTypes = new Map<
+	string,
+	Exclude<PropType<unknown>, Array<unknown>>
```
carsoli, line 92: if they expose it, I personally think it might be better to use `PropConstructor`

Vue's definition of `PropType<T>` is `PropConstructor | PropConstructor[]`
and you don't want the array of whatever PropConstructor is 

Assumptions:
- they actually expose it
-  TS doesn't scream about it
-  you find it more readable

carsoli: Otherwise don't change

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,198 @@
...
+				zodTypeSet.has('ZodDefault') || zodTypeSet.has('ZodOptional')
+
+			const prop: PropOptions<unknown, boolean> = {
+				validator: propValidator(shape, key),
+			}
+
+			if (isOptional) prop.default = shape._def.defaultValue
+			else prop.required = true
+
+			const propType = [...zodTypeSet]
```
carsoli, line 172: can u rename to `propTypes`

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,198 @@
...
+						typeof zodToVueTypes
+					>
+				})
+
+			prop.type = propType.length === 1 ? propType[0] : propType
+
+			return [key, prop]
+			/* eslint-enable no-console */
+		}),
+	) as {
```
carsoli, line 193: why did you need the assertion

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,198 @@
...
+		[KEY in keyof SCHEMA['shape']]: Omit<PropOptions, 'required' | 'type'> & {
+			required: z.output<SCHEMA>[KEY] extends undefined ? false : true
+			type: PropType<z.output<SCHEMA>[KEY]>
```
carsoli, line 196: wouldn't `type` here be the same thing that was excluded from PropType
so
```Exclude PropType<z.output<SCHEMA>[KEY]>, Array<unknown>```

and as proposed in a separate comment
`PropConstructor<z.output<SCHEMA>[KEY]>`

File: packages/kotti-ui/source/props.ts (hunk 26 above)
Comment (carsoli) [reply packages/kotti-ui/source/props.ts:92]: Not exported:
Personally make a separate type here:
`type PropConstructor = Exclude<PropType<unknown>, Array<unknown>>`

with a comment explaining, that we would've otherwise used the declared type from `Vue` if it were exported

File: packages/kotti-ui/source/props.ts (hunk 29 above)
Comment (carsoli) [reply packages/kotti-ui/source/props.ts:196]: No, because of this line: https://github.com/3YOURMIND/kotti/pull/504/files#diff-fbcbf2aa3c29073ef5702e2fef52a3667008fcdc1a0b78c3825ec65bdbee1ed7R188

File: packages/kotti-ui/source/props.ts [hunk 30]
```diff
@@ -0,0 +1,198 @@
...
+			for (const typeSet of options.map((x) => walkShape(x)))
+				for (const type of typeSet) types.add(type)
+
+			if (DEBUG_WALK_SHAPE) console.groupEnd()
+		} else {
+			if (DEBUG_WALK_SHAPE) console.log(`walkTypes: pushing “${typeName}”`)
+			types.add(typeName)
+		}
+
+		shapePointer = shapePointer._def.innerType
```
carsoli, line 75: as far as the definition I saw is implemented
`innerType` is only on 3 extensions of `ZodTypeDef` which is the type of `_def` here
1) ZodNullableDef
2) ZodOptionalDef
3) ZodDefaultDef

so maybe, add a comment here explaining that the deep recursion is to walk the types of `null` | `undefined`, and whatever the `default` is

carsoli: Florian; will use a switch on `ZodTypeAny`

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	return types
+	/* eslint-enable no-console */
+}
+
+/**
+ * These types do not influence the generation of Vue’s `propName.type` Arrays and are therefore discarded
+ */
+const ignoredZodTypes = new Set(['ZodDefault', 'ZodNullable', 'ZodOptional'])
```
carsoli, line 85: use the enum for ZodType if u find it

packages/kotti-ui/source/props.ts
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * This maps the internal zod name of a type to the the constructor that Vue expects in `propName.type`
+ */
+const zodToVueTypes = new Map<
+	string,
+	Exclude<PropType<unknown>, Array<unknown>>
+>([
```
carsoli, line 93: since you return a map with keys that are a subset of what ZOD supports, use this 
`ZodFirstPartyTypeKind.ENUM_KEY`

packages/kotti-ui/source/kotti-banner/types.ts
```diff
@@ -1,20 +1,18 @@
...
-
 export namespace KottiBanner {
-	export const propsInternalSchema = z.object({
-		actionText: z.string().nullable(),
-		expandCloseLabel: z.string().nullable(),
-		expandLabel: z.string().nullable(),
-		icon: yocoIconSchema,
-		isGray: z.boolean(),
+	export const propsSchema = z.object({
+		actionText: z.string().nullable().default(null),
```
carsoli, line 6: optional().default(null)

packages/kotti-ui/source/kotti-breadcrumb/types.ts
```diff
@@ -29,14 +27,14 @@ export namespace KottiBreadcrumb {
...
 	export type Separator = z.infer<typeof separatorSchema>
 
-	export const propsInternalSchema = z.object({
+	export const propsSchema = z.object({
 		breadcrumbs: z.array(breadcrumbSchema),
-		separator: separatorSchema,
+		separator: separatorSchema.default(() => ({
+			style: SeparatorType.ICON as const,
+			value: Yoco.Icon.CHEVRON_RIGHT,
+		})),
```
carsoli, line 35: Florian: move `as const` to this instead of on :33

packages/kotti-ui/source/kotti-user-menu/types.ts
carsoli: i.e. 
add `link: z.string().optional()`

packages/documentation/components/NavBar.vue
```diff
@@ -86,17 +86,7 @@ export default defineComponent({
...
 			],
 			sections: menu.map(
-				(
-					section,
-				): {
-					title: string | null
-					links: Array<{
-						icon: Yoco.Icon
-						title: string
-						path: string
```
carsoli, line 96: florian: rename path to to

packages/kotti-ui/source/props.ts
carsoli: UP TO YOU

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
carsoli: @carsoli TODO tomorrow

packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
carsoli: create issue for this overkill of a prop 
@carsoli @FlorianWendelborn

packages/kotti-ui/source/props.ts
```diff
@@ -52,3 +39,137 @@ export const propValidator = <
...
+			 * However, this currently doesn’t seem to work with z.function().default()
+			 * and this also internally just calls safeParse on the object — evaluating
+			 * the default function, which wastes performance AND could cause
+			 * unintended side-effects
+			 */
+			const isOptional =
+				zodTypeSet.has('ZodDefault') || zodTypeSet.has('ZodOptional')
+
+			const prop: PropOptions<unknown, boolean> = {
+				validator: propValidator(shape, key),
```
carsoli, line 166: please rename those variables as follows

`shape` ==> `propType`
`key` ==> `propName`

carsoli: Florian: `propType` ==> `propSchema`

packages/kotti-ui/source/props.ts
carsoli: `walkShape` ==> `walkSchema` ?
`SHAPE` ==> `SCHEMA`

packages/kotti-ui/source/props.ts
carsoli: UP TO YOU

packages/kotti-ui/source/props.ts
carsoli: UP TO YOU

packages/kotti-ui/source/props.ts
carsoli: UP TO YOU

packages/kotti-ui/source/props.ts
carsoli: check in unit tests @carsoli

packages/kotti-ui/source/props.ts
FlorianWendelborn: Actually, as it turns out the `required` should use `z.input` and the `type` should use `z.output`.

packages/documentation/components/NavBar.vue
```diff
@@ -86,28 +86,22 @@ export default defineComponent({
...
+									? route.value.path === '/'
+									: route.value.path.startsWith(`/${subsection.path}`),
+							title: subsection.title,
+							to: `/${subsection.path}${
+								subsection.pages.length >= 1
+									? `/${subsection.pages[0].path}`
+									: ''
+							}`,
+						}),
+					) as unknown as Kotti.Navbar.Section['links'], // TS doesn’t understand the non-empty arrays without the cast due to the .map
```
FlorianWendelborn, line 104: Kind-of an issue with the non-empty arrays.

packages/documentation/components/NavBar.vue
```diff
@@ -86,28 +86,22 @@ export default defineComponent({
...
-							subsection.pages.length >= 1 ? `/${subsection.pages[0].path}` : ''
-						}`,
-						isActive:
-							subsection.path === ''
-								? route.value.path === '/'
-								: route.value.path.startsWith(`/${subsection.path}`),
-					})),
+				(section): Kotti.Navbar.Section => ({
+					links: section.subsections.map(
+						(subsection): Kotti.Navbar.SectionLink & { to: string } => ({
```
FlorianWendelborn, line 91: Removal of `to: z.unknown().optional()` made this necessary

packages/kotti-ui/source/kotti-user-menu/types.ts
```diff
@@ -1,19 +1,28 @@
...
 
 export namespace KottiUserMenu {
-	export type SectionLink = KottiNavbar.SectionLink
-	export type Section = KottiNavbar.Section
+	export const sectionLinkSchema = z
+		.object({
+			link: z.string().optional(),
+			title: z.string(),
+		})
+		.passthrough()
```
FlorianWendelborn, line 9: I flagged theses as `passthrough` because we allow custom properties on them. Unfortunately, this didn’t fix the type definition though. Perhaps worth making an issue at `zod` for

packages/documentation/components/NavBar.vue
carsoli: that's what I was looking for actually, it gives clarity, that this is an usage-specific API

packages/documentation/components/NavBar.vue
carsoli: Do you think we should move away from them then? 

I was only hoping we use them as an extra warning that those are required, but if for every usage, we would need a type cast, then maybe not?


The weird thing is, if the arrays are enforced to be non-empty, the map shouldn't be problematic, unless it's used like a conditional array function (e.g. filter) but I guess at compilation time this is not detectable, ad most likely the ts compiler alone won't figure that out. 

In any case, let's discuss potentially not using them then, the disadvantages outweigh the benefits.

packages/kotti-ui/source/props.test.ts
```diff
@@ -0,0 +1,451 @@
...
+	}
+
+	successCases.forEach((c) => expect(validator(c)).toBe(true))
+	failCases.forEach((c) => expect(validator(c)).toBe(false))
+}
+
+describe('array', () => {
+	beforeAll(silenceConsole)
+
+	it('z.array(z.number())', () => {
```
carsoli: I know it is concise and potentially clear, but I think the test descriptions should be in pure English.
e.g.
`it('tests array of numbers')`

`it('tests nullable array of numbers')` OR `array of numbers or null` 

etc

packages/kotti-ui/source/props.test.ts
```diff
@@ -0,0 +1,451 @@
...
+		expectRequired(prop)
+		expect(prop.type).toBe(Array)
+		expectValidator(
+			prop,
+			[[], [1], [-1], null],
+			[[null], ['string'], 1, 'test', undefined],
+		)
+	})
+
+	it('z.array(z.number()).default()', () => {
```
carsoli: also here for example, would be nice to have a generic case for `default()` that `it('tests that a prop with default() is implicitly optional`) and then you add the undefined as the success case regardless of the suite prop type

packages/kotti-ui/source/props.test.ts
```diff
@@ -0,0 +1,451 @@
...
+			prop: z.array(z.number()).default(() => []),
+		})
+
+		const { prop } = makeProps(schema)
+
+		expectDefault(prop, [])
+		expect(prop.type).toBe(Array)
+		expectValidator(
+			prop,
+			[[], [1], [-1], undefined],
```
carsoli: would then keep the undefined out of this case

like i said, it was not clear to me at first that zod assume `undefined` as a union to the type if you use `default()` 
> so have a special suite for it with one test case

packages/kotti-ui/source/props.test.ts
```diff
@@ -0,0 +1,451 @@
...
+			prop: z.array(z.number()),
+		})
+		const { prop } = makeProps(schema)
+
+		expectRequired(prop)
+		expect(prop.type).toBe(Array)
+		expectValidator(prop, [[], [1], [-1]], [1, null, 'test', undefined])
+	})
+
+	it('z.array(z.number()).nullable()', () => {
```
carsoli: likewise you can have the generic suite for `nullable` where you clarify that nullable unions the type to the general type, and not to each element

so `nullable` array, would add Array<X> | null BUT NOT Array<X | null>

packages/kotti-ui/source/props.test.ts
carsoli: up to you

packages/kotti-ui/source/props.test.ts
carsoli: up to you

packages/kotti-ui/source/props.test.ts
carsoli: up to you

packages/kotti-ui/source/props.test.ts
carsoli: up to you

packages/kotti-ui/source/props.test.ts
```diff
@@ -0,0 +1,451 @@
...
+
+	it('z.boolean().nullable().default()', () => {
+		const schema = z.object({
+			prop: z.boolean().nullable().default(null),
+		})
+		const { prop } = makeProps(schema)
+
+		expectDefault(prop, null)
+		expect(prop.type).toBe(Boolean)
+		expectValidator(prop, [true, false, null, undefined], ['true'])
```
carsoli: the fail scenario is a string true for a fail scenario, unrelated to `nullable()` and `default()` 

up to you to keep it on this test or move it out

packages/kotti-ui/source/props.test.ts
```diff
@@ -0,0 +1,451 @@
...
+	})
+})
+
+describe('function', () => {
+	beforeAll(silenceConsole)
+
```
carsoli: do we actually make use of `function.args(...)` and `function.returns(`
If yes? can a test be added

packages/kotti-ui/source/props.test.ts
```diff
@@ -0,0 +1,451 @@
...
+	it('z.function().default()', () => {
+		const schema = z.object({
+			prop: z.function().default(() => undefined),
+		})
+
+		const { prop } = makeProps(schema)
+
+		expectDefault(prop, undefined)
+		expect(prop.type).toBe(Function)
+		expectValidator(prop, [() => undefined], ['string', null])
```
carsoli: maybe add `() => null` to your failure scenarios

packages/kotti-ui/source/props.test.ts
```diff
@@ -0,0 +1,451 @@
...
+			prop: z.number(),
+		})
+		const { prop } = makeProps(schema)
+
+		expectRequired(prop)
+		expect(prop.type).toBe(Number)
+		expectValidator(
+			prop,
+			[1, 0],
+			[false, [], 'string', Number.NaN, null, undefined],
```
carsoli: so in typescript `Number.NaN` extends the `Number` interface

in JS, the `typeof` operation returns `typeof NaN` and `typeof Number.NaN` to be `number`

so I don't find this intuitively to be a failure scenario

feel free to point it out in a separate use-case just to point out that zod doesn't support it to be `number`

maybe also try other edge cases - from the `Integer` class like MAX_SAFE_INTEGER
or not

up to you

carsoli: edge cases can be more than the `SAFE_INTEGER`s ofc

packages/kotti-ui/source/props.ts
```diff
@@ -2,11 +2,20 @@ import { PropOptions, PropType } from '@vue/composition-api'
...
 
-import { MapValueType } from './types/utilities'
-
 const DEBUG_MAKE_PROPS = false as const // enable to print debug log
 const DEBUG_WALK_SHAPE = false as const // enable to print debug log
 
+/**
+ * This type is not exported directly by @vue/composition-api
+ */
+type VuePropConstructor = Exclude<PropType<unknown>, Array<unknown>>
```
carsoli: wouldn't it be `Array<PropType<unknown>>`?

packages/kotti-ui/source/props.ts
```diff
@@ -42,62 +51,95 @@ const propValidator = <SCHEMA extends z.ZodTypeAny>(
...
+ */
+const walkSchemaTypes = <SCHEMA extends z.ZodTypeAny>(
+	schema: SCHEMA,
+): Set<z.ZodFirstPartyTypeKind> => {
 	/* eslint-disable no-console */
-	const types = new Set<string>()
+	const typeName = schema._def.typeName as z.ZodFirstPartyTypeKind
 
-	let shapePointer = shape
+	if (DEBUG_WALK_SHAPE)
```
carsoli: minor: u forgot to rename the debug flags to `DEBUG_WALK_SCHEMA_TYPES`

carsoli (approved):
If you attend to any of the comments, please have separate commits/commit, for an easier review process 
Thanks!

packages/kotti-ui/source/props.ts
FlorianWendelborn: It’s possible to do that, but that wouldn’t make this type any better and it’s also weird to exclude `Array<Array<PropConstructor>>` as well. I’m just excluding anything array-like from this type which I think is pretty clear.

packages/kotti-ui/source/props.ts
FlorianWendelborn: Good catch

packages/kotti-ui/source/props.test.ts
```diff
@@ -0,0 +1,451 @@
...
+
+	it('z.object()', () => {
+		const schema = z.object({
+			prop: z.object({ key: z.boolean() }),
+		})
+		const { prop } = makeProps(schema)
+
+		expectRequired(prop)
+		expect(prop.type).toBe(Object)
+		expectValidator(
```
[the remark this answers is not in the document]
FlorianWendelborn: Added to all `z.object()` tests. It should/will fail of course

packages/kotti-ui/source/props.test.ts
FlorianWendelborn: Somewhat addressed. I hope it’s a good compromise

packages/kotti-ui/source/props.test.ts
FlorianWendelborn: Somewhat addressed via `TYPE_SUCCESS` / `TYPE_FAILURE`. I think this makes it easier to understand the tests

packages/kotti-ui/source/props.test.ts
FlorianWendelborn: Somewhat addressed via `TYPE_SUCCESS` / `TYPE_FAILURE`

packages/kotti-ui/source/props.test.ts
FlorianWendelborn: Somewhat addressed via `TYPE_SUCCESS`/`TYPE_FAILURE`

packages/kotti-ui/source/props.test.ts
FlorianWendelborn: Regarding `NaN`, I’m pretty sure most developers aren’t even fully aware of it, especially not aware enough to always treat all their number handling with `NaN` in mind or to actively _expect_ `NaN` to be allowed to be passed to something that needs a number.

I think it’s a very sane default behavior of `zod` to exclude `NaN`. Even the `isNumber` function in the `KtFieldNumber` utilities explicitly forbids `NaN`.

However, I discovered two edge-cases that I actually did **not** expect to behave like they do in `zod`:

1. `Number.POSITIVE_INFINITY` (is valid)
2. `Number.NEGATIVE_INFINITY` (is valid)

packages/documentation/components/NavBar.vue
FlorianWendelborn: I think I found a good solution for this. See `chore(add-zod): Replace .nonempty with .refine(...refinementNonEmpty)`

packages/kotti-ui/source/make-props.test.ts
```diff
@@ -1,51 +1,143 @@
...
+				return 'valid'
+			},
+			pass: passNoDefault && passRequired,
+		}
+	},
+	toBeType(prop: PropOptions<unknown, boolean>, expected: PropType<unknown>) {
+		const actualPropTypes = new Set(castArray(prop.type as PropType<unknown>))
+		const expectedPropTypes = new Set(castArray(expected))
+
+		const pass = isEqual(actualPropTypes, expectedPropTypes)
```
carsoli, line 41: I just tried 
```_.isEqual([1,2], [2,1])
```
and it returns `false`

which might be non-deterministic without an explicit `sort`

packages/kotti-ui/source/make-props.test.ts
```diff
@@ -1,51 +1,143 @@
...
+	},
+	toDefaultTo(prop: PropOptions<unknown, boolean>, expectedDefault: unknown) {
+		if (typeof prop.default !== 'function')
+			return {
+				message: () => 'Expected “prop.default” to be a function',
+				pass: false,
+			}
+
+		const actualDefaultValue = prop.default()
+		const passDefault = isEqual(actualDefaultValue, expectedDefault)
```
carsoli, line 65: minor: for consistency either rename `actualDefaultValue` to `actualDefault` or the argument to `expectedDefaultValue`

I prefer the latter but it's up to you

packages/kotti-ui/source/make-props.test.ts
```diff
@@ -1,51 +1,143 @@
...
+		if (validator === undefined)
+			return {
+				message: () => 'Expected prop.validator to exist',
+				pass: false,
+			}
+
+		const successfulCases = cases.filter((c) => validator(c))
+		const failedCases = cases.filter((c) => !validator(c))
+
+		const wrongCases = this.isNot ? successfulCases : failedCases
```
carsoli, line 105: IMO, this overcomplicates just passing `success` and `failure` scenarios`. 
> the upside, I see, is that it's done like a test library would implement this, since at the end of the day, the API that's exposed will be chainable, thus, human-readable.

doesn't block anything, just don't see how this was a better refactor in terms of readability for the implementation itself, not for the usage of the end mutated API.

even more recognizable in the `pass` condition below https://github.com/3YOURMIND/kotti/pull/504/commits/83b368ff092c18779027c96cd9f717350e6bb9c9?diff=unified&w=1#diff-e0075d751357fe7b9821160bc2f8830ed316ea5140182506249baa0b298a59d3R120

packages/kotti-ui/source/make-props.test.ts
FlorianWendelborn: It should be `false` as you’re comparing two non-equal arrays. Those are `Set`s

carsoli: It should be `false`
> wasn't saying it should be `true`

Those are `Set`s
> what are `Sets`? `castArray` returns an `Array`? 
> my point is more about how you construct the types outside of this function, to my knowledge, most Array prototype functions are "async" - thus, not respecting sorting, which might cause non-determinism

packages/kotti-ui/source/make-props.test.ts
FlorianWendelborn: I initially had this entire function duplicated because using `.not` without switching between success and failure basically makes it wrong as it’s based on `all tests` or `no tests` but it flags `some tests` as well.

So, I basically added these 3 checks here to avoid having it duplicated entirely if that makes sense

packages/kotti-ui/source/zod-refinements.test.ts
```diff
@@ -0,0 +1,24 @@
+import { z } from 'zod'
+
+import { refinementNonEmpty } from './zod-refinements'
+
+describe('refinementNonEmpty', () => {
+	const schema = z.array(z.number()).refine(...refinementNonEmpty)
+
+	it('works for arrays > 1', () => {
```
carsoli, line 8: `with length > 1`*

packages/kotti-ui/source/zod-refinements.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+		expect(result.success).toBe(false)
+
+		// type hint
+		if ('error' in result)
+			expect(result.error.errors).toEqual([
+				{ code: 'custom', message: 'array may not be empty', path: [] },
+			])
+		else throw new Error('error.result doesn’t exist')
+	})
```
carsoli, line 23: `result.error`*

carsoli (approved):

packages/kotti-ui/source/make-props.test.ts
FlorianWendelborn: Also, slight correction: Array prototype functions aren’t `async`, they’re entirely synchonous, but have a different scope than the outside.

E.g. `Array.prototype.forEach` is doing the same thing as `for (const item of array)`, besides that you can’t return/break/throw, etc.. That part is only the case because they use callbacks though, not because they’re actually async/running in the next V8 tick

carsoli: nvm, I didn't see `new Set(` might have been the diff I was on or just my idiocy.

PR #505, opened 2021-09-21
add-zod(1): Upgrade TS to 4.3.5

FlorianWendelborn:
TS 4.4.2 is unfortunately not possible yet, as there’s this rollup+typescript bug that will most likely be fixed in TS 4.4.3:

https://github.com/rollup/plugins/issues/983
https://github.com/microsoft/TypeScript/issues/45633

Co-Authored-By: Moritz Vetter <16950410+HansAuger@users.noreply.github.com>

PR #508, opened 2021-09-21
add-zod(4): Move isYocoIcon to @3yourmind/yoco, Add yocoSchema

FlorianWendelborn:
Rebased on top of #505, #506, and #507

### DIFF
--- a/packages/yoco/source/index.ts
+++ b/packages/yoco/source/index.ts
@@ -1,1 +1,3 @@
+export * from './schema'
 export * as Yoco from './types'
+export * from './validator'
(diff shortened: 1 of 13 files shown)


HansAuger (approved):

packages/yoco/source/index.ts
```diff
@@ -1 +1,3 @@
+export * from './schema'
 export * as Yoco from './types'
```
carsoli, line 2: do u need this after importing the schema?

carsoli: `export * as Yoco from './schema'`

while deleting the line for the types

carsoli (approved):

PR #524, opened 2021-10-03
refactor(KtComment): Remove Payload Weirdness & Fix Internal Event Typo

FlorianWendelborn:

PR #525, opened 2021-10-04
chore(KtComment): Remove Dead Condition

FlorianWendelborn:
Rebased on #524.


packages/kotti-ui/source/kotti-comment/KtCommentInput.vue
```diff
@@ -70,17 +70,12 @@ export default {
 			this.$refs.textarea.style.height = `${height}px`
 		},
 		handleSubmitClick() {
-			if (!this.text && this.text === '\n') {
```
FlorianWendelborn, line 73: How can the text be simultaneously falsy and be a newline?

PR #543, opened 2021-10-14
Refactor kotti pagination

Isokaeder:
The usual shtick, some typescript, some zod, some rearranging of code and whatnot


packages/kotti-ui/source/kotti-pagination/components/PaginationExpanded.vue
```diff
@@ -1,47 +1,46 @@
 <template>
-	<div class="inline-container">
+	<div class="kt-pagination__inline-container">
 		<li
 			v-for="(page, index) in totalPages"
 			:key="index"
-			:class="paginatorClasses(page, 'page-item--active')"
-			@click="$emit('setPage', page)"
+			:class="paginatorClasses(page)"
+			@click="$emit('setPage', page - 1)"
```
FlorianWendelborn, line 7: This looks wrong. Why was this changed?

packages/kotti-ui/source/kotti-pagination/KtPagination.vue
```diff
@@ -1,163 +1,139 @@
...
+				currentPage: currentPage.value,
+				fixedWidth: props.fixedWidth,
+				maximumPage: maximumPage.value,
+				pageSize: props.pageSize,
+				total: props.total,
+			})),
+			previousPage: () => {
+				if (currentPage.value === 0) return
+				currentPage.value -= 1
+				emit('previousPageClicked', currentPage.value + 1)
```
FlorianWendelborn, line 85: This API should probably be simplified. Why would the app care if the previous button was clicked instead of a specific page number? Let’s just emit the normal event

Maybe in a follow-up PR after this one is merged, alternatively let’s make an issue if you don’t want to do this immediately

packages/kotti-ui/source/kotti-pagination/KtPagination.vue
```diff
@@ -1,163 +1,139 @@
...
+			},
+			setCurrentPage: (page) => {
+				if (page === currentPage.value) return
+				currentPage.value = page
+				emit('currentPageChange', currentPage.value + 1)
```
FlorianWendelborn, line 90: This API should probably be simplified. Why would the app care if the previous button was clicked instead of a specific page number? Let’s just emit the normal event

Maybe in a follow-up PR after this one is merged, alternatively let’s make an issue if you don’t want to do this immediately

packages/kotti-ui/source/kotti-pagination/KtPagination.vue
```diff
@@ -1,163 +1,139 @@
...
-	props: {
-		adjacentAmount: { type: Number, default: 1 },
-		fixedWidth: { type: Boolean, default: false },
-		page: { type: Number, default: 1 },
-		pageSize: { type: Number, default: 10 },
-		pagingStyle: { type: String, default: 'expand' },
-		total: { type: Number, required: true },
+	props: makeProps(KottiPagination.propsSchema),
+	setup(props, { emit }) {
+		const currentPage = ref(props.page - 1)
```
FlorianWendelborn, line 45: This API should probably just be controlled so we can ensure that the page is properly handled.

I’d suggest either a follow-up PR after this one is merged, or an issue if you don’t want to do this immediately

FlorianWendelborn (approved):

Isokaeder:
Closed, replaced by #1007

Isokaeder:
Closed, replaced by #1007

PR #547, opened 2021-11-22
add-date-test-attributes-to-kt-field-single-select-remote

carsoli:
IconTextItem already supported dataTest attributes, so I just pass them down

### DIFF
--- a/packages/documentation/pages/usage/components/form-fields.vue
+++ b/packages/documentation/pages/usage/components/form-fields.vue
@@ -840,13 +840,18 @@ export default defineComponent({
 			) {
 				const options =
 					component === 'KtFieldSingleSelectRemote'
-						? singleOrMultiSelectOptions.filter((option) =>
-								option.label
-									.toLowerCase()
-									.includes(
-										(remoteSingleSelectQuery.value ?? '').toLowerCase(),
-									),
-						  )
+						? singleOrMultiSelectOptions
+								.filter((option) =>
+									option.label
+										.toLowerCase()
+										.includes(
+											(remoteSingleSelectQuery.value ?? '').toLowerCase(),
+										),
+								)
+								.map((option) => ({
+									...option,
+									dataTest: `dataTest ${String(option.value)}`,
+								}))
 						: singleOrMultiSelectOptions
 
 				Object.assign(additionalProps, {
--- a/packages/documentation/pages/usage/components/form-fields.vue
+++ b/packages/documentation/pages/usage/components/form-fields.vue
@@ -6,7 +6,7 @@
 			<div class="overview">
 				<div class="overview__component">
 					<h4>Component</h4>
-					<KtForm v-model="values">
+					<KtForm v-model="values" :formId="settings.formId">
 						<component
 							:is="componentRepresenation.name"
 							v-bind="componentRepresenation.props"
@@ -78,6 +78,7 @@
 								{ label: 'Large', value: 'large' },
 							]"
 						/>
+						<KtFieldText formKey="formId" isOptional label="Form ID" />
 						<KtFieldSingleSelect
 							formKey="validation"
 							isOptional
@@ -635,6 +636,7 @@ export default defineComponent({
 				isOptional: Kotti.FieldToggle.Value
 			}
 			component: ComponentNames
+			formId: Kotti.FieldText.Value
 			hasHelpTextSlot: boolean
 			helpDescription: Kotti.FieldText.Value
 			helpText: Kotti.FieldText.Value
@@ -676,6 +678,7 @@ export default defineComponent({
 			hasHelpTextSlot: false,
 			helpDescription: null,
 			helpText: null,
+			formId: null,
 			label: 'Label',
 			leftIcon: null,
 			locale: 'en-US',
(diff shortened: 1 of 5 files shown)


packages/kotti-ui/source/kotti-field-select/types.ts
```diff
@@ -55,6 +55,7 @@ export namespace KottiFieldSingleSelect {
 export namespace KottiFieldSingleSelectRemote {
 	export type Props = KottiField.Props<Value, string | null> &
 		Shared.Props & {
+			dataTest: string | null
```
HansAuger, line 58: I don't see this being used in the component anywhere.

packages/documentation/pages/usage/components/form-fields.vue
```diff
@@ -840,13 +840,18 @@ export default defineComponent({
...
+								.filter((option) =>
+									option.label
+										.toLowerCase()
+										.includes(
+											(remoteSingleSelectQuery.value ?? '').toLowerCase(),
+										),
+								)
+								.map((option) => ({
+									...option,
+									dataTest: `dataTest ${String(option.value)}`,
```
HansAuger, line 853: AFAIK those data-test attributes should be unique, right? So It would make sense to have a marker per field as well so in case there are two fields on the same page, to make them distinguishable. 

In other words I would propose we also use the `dataTest` attribute that you added as a prop for remote single selects here like so `dataTest: `dataTest ${ogFieldDataTest} ${String(option.value)}`

HansAuger (changes requested):

packages/kotti-ui/source/kotti-field-select/types.ts
carsoli: because the `modifiedOptions` that are passed, merely `...option` then add more attributes to it 

but you should see a change on the template level where I bound `dataTest` to `option.dataTest`

packages/documentation/pages/usage/components/form-fields.vue
carsoli: what would `ogFieldDataTest` be

carsoli: oh wait nvm i actually forgot to add it to the additional props to enable user-input here

packages/kotti-ui/source/kotti-field-select/types.ts
carsoli: so basically - what kind of change would be needed logically, even? it's just the prop passed down, then remapped by the `modifiedOptions` declaration

HansAuger: Sure, but I would expect this `dataTest` on the field level, not on the option level. 
Like so:
```html
<KtField
	v-bind="{ field }"
	class="kt-field-select kt-field-select--single-remote"
	:data-test="dataTest"
	:getEmptyValue="() => null"
	:helpTextSlot="$slots.helpText"
>
```

HansAuger: The dataTest that are passed to the options are completely generated and are not derived from the prop that is supposedly passed here.


...by the way, am I stupid or is `isLoadingOptions` also effectively doing nothing in the component?

carsoli: NVM - saw what you meant

packages/documentation/pages/usage/components/form-fields.vue
carsoli: if you want to prefix with the field dataTest, I'd pass the formId for all of them  - on it

packages/kotti-ui/source/kotti-field-select/types.ts
carsoli: I misplaced the data-test in the type definition of the field itself, yes.

it belonged on the options, and I added it and its corresponding logic in later commits.

didn't check `isLoadingOptions` but can do

no the ones on the options are NOT generated automatically. On the contrary, the ones on the fields are AUTOMATICALLY generated from the `formId` and the `formKey` and the hierarchy
> so I removed the `dataTest` from the field-level (my mistake initially)
- if the user wants to tweak the data-test, they can tweak the `formId`
- but there is no allowance through the API to tweak it per field 
- because we always relied on `formPath`

This ^ is a limitation however, because not every field is used within a `KtForm`
so needs fixing in the future

packages/documentation/pages/usage/components/form-fields.vue
carsoli: `ogFieldDataTest` is sadly not exposed on the API

carsoli: and actually doesn't make sense to be - since it's internally generated and serves the purpose of identifying an HTML element - which is usually queried post rendering anyways

carsoli:
Screenshot from 2021-11-22 19-35-27
Screenshot from 2021-11-22 19-35-42
Screenshot from 2021-11-22 19-37-11

@HansAuger

packages/kotti-ui/source/kotti-field-select/types.ts
HansAuger: Got it

HansAuger:
LGTM

HansAuger (approved):

PR #549, opened 2021-11-24
KtNavbar: Add Support for Custom Link Components

HansAuger:
Tries to solve the same problem as #473 but in a more controlled way.

### DIFF
--- a/packages/documentation/pages/usage/layouts/navbar.vue
+++ b/packages/documentation/pages/usage/layouts/navbar.vue
@@ -43,7 +43,7 @@ notificationData: {
 
 ### Menu
 
-`KtNavbar` only supports this menu structure:
+This is how you display navigation links in `KtNavbar`:
 
 ```js
 menuData: [
@@ -51,20 +51,53 @@ menuData: [
 		title: 'Order Management',
 		links: [
 			{
-				link: '#',
-				title: 'Orders',
 				icon: 'invoice',
 				isActive: false,
+				linkProps: {
+					href: '#',
+				},
+				title: 'Orders',
 			},
 			{
-				link: '#',
-				title: 'Quotes',
 				icon: 'request',
 				isActive: false,
+				linkProps: {
+					href: '#',
+				},
+				title: 'Quotes',
 			},
 		],
 	},
 ]
+```
+
+The links in KtNavbar are normal `a` elements. However if you want to use one of
+the popular vue router frameworks, you can exchange this for any individual link
+via `v-bind:is`. The data provided in `linkProps` will be bound to the component:
+
+```js
+const links = [
+	// For Nuxt's router framework
+	{
+		component: 'nuxt-link',
+		icon: 'invoice',
+		isActive: false,
+		linkProps: {
+			to: '#',
+		},
+		title: 'Orders',
+	},
+	// For VueRouter
+	{
+		component: 'router-link',
+		icon: 'request',
+		isActive: false,
+		linkProps: {
+			to: '#',
+		},
+		title: 'Quotes',
+	},
+],
 ```
 
 ### Quick Links
--- a/packages/documentation/pages/usage/layouts/navbar.vue
+++ b/packages/documentation/pages/usage/layouts/navbar.vue
@@ -53,7 +53,7 @@ menuData: [
 			{
 				icon: 'invoice',
 				isActive: false,
-				linkProps: {
+				props: {
 					href: '#',
 				},
 				title: 'Orders',
@@ -61,7 +61,7 @@ menuData: [
 			{
 				icon: 'request',
 				isActive: false,
-				linkProps: {
+				props: {
 					href: '#',
 				},
 				title: 'Quotes',
@@ -83,7 +83,7 @@ const links = [
 		component: 'nuxt-link',
 		icon: 'invoice',
 		isActive: false,
-		linkProps: {
+		props: {
 			to: '#',
 		},
 		title: 'Orders',
@@ -93,7 +93,7 @@ const links = [
 		component: 'router-link',
 		icon: 'request',
 		isActive: false,
-		linkProps: {
+		props: {
 			to: '#',
 		},
 		title: 'Quotes',
--- a/packages/documentation/pages/usage/layouts/navbar.vue
+++ b/packages/documentation/pages/usage/layouts/navbar.vue
@@ -71,9 +71,10 @@ menuData: [
 ]
 ```
 
-The links in KtNavbar are normal `a` elements. However if you want to use one of
-the popular vue router frameworks, you can exchange this for any individual link
-via `v-bind:is`. The data provided in `linkProps` will be bound to the component:
+By default the elements from the `links` prop on `KtNavbar` are bound to an anchor
+HTML element (`<a />`). Therefore, providing `href` attribute on each element suffices
+in most cases. In case of the need for a custom component, provide the `component`
+attribute and any corresponding prop that would need to be bound to it.
 
 ```js
 const links = [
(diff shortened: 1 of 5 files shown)


packages/kotti-ui/source/kotti-navbar/types.ts
```diff
@@ -17,14 +17,13 @@ export namespace KottiNavbar {
...
-		.object({
-			icon: yocoIconSchema,
-			isActive: z.boolean(),
-			link: z.string().optional(),
-			title: z.string(),
-		})
-		.passthrough()
+	export const sectionLinkSchema = z.object({
+		icon: yocoIconSchema,
+		component: z.string().default('a'),
```
carsoli, line 22: sort `component` above `icon`

packages/documentation/pages/usage/layouts/navbar.vue
```diff
@@ -43,30 +43,63 @@ notificationData: {
...
+				title: 'Quotes',
 			},
 		],
 	},
 ]
 ```
 
+The links in KtNavbar are normal `a` elements. However if you want to use one of
+the popular vue router frameworks, you can exchange this for any individual link
+via `v-bind:is`. The data provided in `linkProps` will be bound to the component:
```
carsoli, line 76: from an API POV, you don't mention how we bind the component

you need to mention that the link object requires a `component` attribute

so to rephrase this part

```
By default the elements from the `links` prop on `KtNavbar` are bound to an anchor HTML element (`<a />`). Therefore, providing `href` attribute suffices on each element in the array automatically redirects.

In case of the need for a custom component, provide the `component` attribute and any corresponding prop that would need to be bound to it.
```

packages/kotti-ui/source/kotti-navbar/components/NavbarMenu.vue
```diff
@@ -6,12 +6,13 @@
 				class="kt-navbar-menu__section"
 				v-text="isNarrow ? '' : item.title"
 			/>
-			<a
+			<component
+				:is="link.component ? link.component : 'a'"
```
carsoli, line 10: according to the prop schema `a` is default

but i see it as a fallback , not a Vue prop `default` in the parent of this?

packages/kotti-ui/source/kotti-navbar/types.ts
```diff
@@ -17,14 +17,13 @@ export namespace KottiNavbar {
...
+		isActive: z.boolean(),
+		linkProps: z.record(z.unknown()).optional(),
```
carsoli, line 24: why do you need `record` and not `object`? is this a `zod` limitation you're aware of?

HansAuger:
Sorry for the spam. I am too shtoopid for github apparently...

packages/kotti-ui/source/kotti-navbar/components/NavbarMenu.vue
FlorianWendelborn: You can parse the received props with zod, which will populate your defaults. What you’re describing is a limitation of `makeProps` since `vue` doesn’t support changing data in the validator.

packages/kotti-ui/source/kotti-navbar/types.ts
FlorianWendelborn: I think this is a bit redundant. Any objections to just calling it `props`?

FlorianWendelborn (approved):
I was a bit skeptical about this approach, but after reviewing it I think it’s actually a good way to solve this weirdness. Approved.

FlorianWendelborn:
After this is finalized & merged, could we get a similar PR for the `UserMenu`? Please note that we definitely need to support JS event handlers for links there though due to the logout buttons.

Perhaps this means that for consistency we should still support the click event in this PR as well? You dropped it but I think you don’t have to do this for what this PR is accomplishing.

FlorianWendelborn:
@HansAuger I think you should `autorebase:opt-in` this PR?

packages/kotti-ui/source/kotti-navbar/types.ts
carsoli: I am still considering how we can better restructure this

basically want to combine the `component` and `props` so that you either have `AnchorLink` or a random object

idk if this will be an overkill and actually make us run into more issues
> would have to test myself a bit

but my point is more it's maybe better to either have `a` or something else that can be anything🤷🏼 
> we can even further overkill this with a generic on the component for the linkProps type
so that we end up with 

```
type AnchorLink = `component: 'a' & HTMLLinkElement`
 
sectionLinkSchema<T extends { [key: string]: unknown } = { [key: string]: unknown }> = z.object({
icon: yocoIconSchema,
isActive: z.boolean(), 
type: AnchorLink | T

```

packages/kotti-ui/source/kotti-navbar/types.ts
carsoli: Idk where my mind is. I thought I left a comment but i didn't apparently 😕 

I had a comment about `default` here

I see you default the type, but you don\t actually `default` it on `VueProps` and you had a runtime fallback on the template itself

`:is=link.component?link.component:'a'`
which should just be purely default `a`

carsoli: nvm found it

packages/kotti-ui/source/kotti-navbar/types.ts
FlorianWendelborn: @carsoli actually it’s possible to make this conditional via using a `z.array(z.union([z.object(...withA), z.object(...withoutA)]))

However, I don’t think it’s strictly necessary. We also support JS-based links as well which wouldn’t really pass a `href` and perhaps at some point we need other props as well (e.g. `download` to download a zip from the navbar or so).

Another possibility is changing this to `component: { name: 'a', props: { href: '/' } }`, but once again I’m fine with the current solution besides that `linkProps` is a bit redundant

packages/kotti-ui/source/kotti-navbar/components/NavbarMenu.vue
```diff
@@ -30,17 +30,24 @@
...
 export default defineComponent({
 	name: 'KtNavbarMenu',
 	components: {
 		NavbarTooltip,
 	},
-	props: {
-		isNarrow: { default: false, type: Boolean },
-		sections: { required: true, type: Array },
-	},
+	props: makeProps(KottiNavbarMenu.propsSchema),
```
FlorianWendelborn, line 50: This is not enough to deeply generate defaults for `link.component`. You need to manually parse the relevant props with `zod`. As mentioned in another comment, Vue does not allow modifying props in the validators which means that `makeProps` cannot implement this as a feature.

packages/kotti-ui/source/kotti-navbar/types.ts
carsoli: I didn't have your answer when I typed the 2nd comment @FlorianWendelborn

carsoli (approved):

packages/kotti-ui/source/kotti-navbar/components/NavbarMenu.vue
```diff
@@ -21,24 +22,38 @@
...
 	components: {
 		NavbarTooltip,
 	},
-	props: {
-		isNarrow: { default: false, type: Boolean },
-		sections: { required: true, type: Array },
+	props: makeProps(KottiNavbarMenu.propsSchema),
+	setup(props) {
+		return {
+			parsedSections: KottiNavbarMenu.propsSchema.shape.sections.parse(
```
FlorianWendelborn, line 53: This isn’t reactive

FlorianWendelborn (changes requested):

FlorianWendelborn (approved):

PR #555, opened 2021-11-24
ci(pull-requests): Add actions/labeler

FlorianWendelborn:
This will automatically add labels to PRs based on globs. Right now, this is only used for the `package:*` labels


FlorianWendelborn:
I think this isn’t working on this branch as a security measure. AFAIK only the workflows from `master` are executed for PRs

HansAuger (approved):

carsoli (approved):

PR #557, opened 2021-11-25
feat(KtUserMenu): make link component in KtUserMenu injectible

HansAuger:
Analogous to what was already done for Navbar

### DIFF
--- a/packages/kotti-ui/source/kotti-user-menu/KtUserMenu.vue
+++ b/packages/kotti-ui/source/kotti-user-menu/KtUserMenu.vue
@@ -79,9 +79,7 @@ export default defineComponent<KottiUserMenu.PropsInternal>({
 			isMenuShow,
 			isNarrow,
 			parsedSections: computed(() =>
-				props.sections
-					? KottiUserMenu.propsSchema.shape.sections.parse(props.sections)
-					: [],
+				KottiUserMenu.propsSchema.shape.sections.parse(props.sections),
 			),
 			userInfoClass: computed(() => ({
 				'kt-user-menu__info': true,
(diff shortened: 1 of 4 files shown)


packages/kotti-ui/source/kotti-user-menu/KtUserMenu.vue
```diff
@@ -77,6 +78,11 @@ export default defineComponent<KottiUserMenu.PropsInternal>({
 			clickawayMenu: () => (isMenuShow.value = false),
 			isMenuShow,
 			isNarrow,
+			parsedSections: computed(() =>
+				props.sections
```
FlorianWendelborn, line 82: Why the ternary here?

FlorianWendelborn (approved):

FlorianWendelborn:
Can you `autorebase:opt-in`?

packages/kotti-ui/source/kotti-user-menu/KtUserMenu.vue
HansAuger: I removed it now. The reason for this is a VSCode bbug (maybe?) that confused me about the type
Screenshot 2021-11-25 at 14 22 00
Screenshot 2021-11-25 at 14 22 14

FlorianWendelborn (approved):

packages/kotti-ui/source/kotti-user-menu/KtUserMenu.vue
FlorianWendelborn: Yeah, it’s a visual bug in vetur with `makeProps`. It somehow shows-up as optional even though it’s not. Cause is unknown but doesn’t seem to affect anything besides showing this weirdly

PR #561, opened 2021-11-28
Refactor kt comment: zod

HansAuger:
Step 1/4:

- [ ] use zod for props schema and typing
- [ ] composition api
- [ ] put css into component
- [ ] other clean ups if necessary

After refactor:
- [ ] Integrate Kotti Translations
- [ ] #253
- [ ] #438

Issue #438, opened 2021-05-19
Readonly Mode for comments


Isokaeder:
KtComment has the following features that are not toggleable:
- Reply button
- More Actions Button with Edit and Delete Buttons
- hovering on a reply will make a comment icon appear
- clicking a reply will open a text input

There should  be a readonly mode, where all of these thins are hidden/do not happen

FlorianWendelborn:
Not sure, but perhaps it’s better to just toggle them individually. It may also be possible we want to limit editing of certain comments to certain users etc.

HansAuger:
I'd be fine with  that, will be breaking change though.

I heard today, that we are thinking even about some more  features, like comments that are only readable for certain users. that implies we should have some slots to customize buttons or other parts of the Comment

Issue #253, opened 2020-08-12
KtComments: Consider CSS Solution


FlorianWendelborn:
`white-space: pre-wrap;` instead of `customDangerousEscapeParser` etc.

PR #562, opened 2021-11-28
Refactor kt comment topic/composition api comment

HansAuger:
Step 2/4:

- [x] use zod for props schema and typing (#561)
- [ ] composition api 
  - [ ] KtComment     <--- we are here
  - [x] KtCommentInput   (#563)
  - [ ] CommentReply (#564)
- [ ] put css into component
- [ ] other clean ups if necessary

After refactor:
- [ ] Integrate Kotti Translations
- [ ] #253
- [ ] #438

### DIFF
--- a/packages/kotti-ui/source/kotti-comment/KtComment.vue
+++ b/packages/kotti-ui/source/kotti-comment/KtComment.vue
@@ -75,7 +75,6 @@
 
 <script lang="ts">
 import { computed, defineComponent, ref } from '@vue/composition-api'
-import compact from 'lodash/compact'
 
 import { KtAvatar } from '../kotti-avatar'
 import { KtButton } from '../kotti-button'
@@ -112,28 +111,26 @@ export default defineComponent<KottiComment.PropsInternal>({
 			emit('delete', payload)
 		}
 		return {
-			actionOptions: computed<Array<KottiButton.Props>>(() =>
-				compact([
-					props.isEditable
-						? {
-								label: 'Edit',
-								onClick: () => {
-									if (inlineMessageValue.value === null)
-										inlineMessageValue.value = props.message
-									isInlineEdit.value = true
-								},
-								type: KottiButton.Type.PRIMARY,
-						  }
-						: null,
-					props.isDeletable
-						? {
-								label: 'Delete',
-								onClick: () => handleDelete(props.id),
-								type: KottiButton.Type.DANGER,
-						  }
-						: null,
-				]),
-			),
+			actionOptions: computed<Array<KottiButton.Props>>(() => {
+				const options = []
+				if (isInlineEdit.value) return options
+				if (props.isEditable)
+					options.push({
+						label: 'Edit',
+						onClick: () => {
+							inlineMessageValue.value = props.message
+							isInlineEdit.value = true
+						},
+						type: KottiButton.Type.PRIMARY,
+					})
+				if (props.isDeletable)
+					options.push({
+						label: 'Delete',
+						onClick: () => handleDelete(props.id),
+						type: KottiButton.Type.DANGER,
+					})
+				return options
+			}),
 			cancelInlineEdit: () => {
 				inlineMessageValue.value = null
 				isInlineEdit.value = false
@@ -141,15 +138,15 @@ export default defineComponent<KottiComment.PropsInternal>({
 			handleDelete,
 			handleEditConfirm: () => {
 				isInlineEdit.value = false
-				if (!inlineMessageValue.value) return
+				if (inlineMessageValue.value === null) return
 				const payload: KottiComment.Events.Edit = {
 					id: props.id,
 					message: inlineMessageValue.value,
 				}
 				emit('edit', payload)
 			},
-			handleInlineReplyClick: ({ userId, userName }: UserData) => {
-				userBeingRepliedTo.value = { userId, userName }
+			handleInlineReplyClick: (replyUserData: UserData) => {
+				userBeingRepliedTo.value = replyUserData
 			},
 			handleInlineSubmit: (commentData: KottiComment.Events.Submit) => {
 				userBeingRepliedTo.value = null
--- a/packages/kotti-ui/source/kotti-comment/KtComment.vue
+++ b/packages/kotti-ui/source/kotti-comment/KtComment.vue
@@ -55,17 +55,17 @@
 					:userAvatar="reply.userAvatar"
 					:userId="reply.userId"
 					:userName="reply.userName"
-					@_inlineDeleteClick="(commentId) => handleDelete(commentId, false)"
+					@_inlineDeleteClick="(commentId) => handleDelete(commentId, true)"
 					@_inlineEditSubmit="$emit('edit', $event)"
 					@_inlineReplyClick="handleInlineReplyClick"
 				/>
 			</div>
 			<KtCommentInput
-				v-if="showInlineReply"
+				v-if="userBeingRepliedTo"
 				isInline
 				:parentId="id"
 				:placeholder="replyToText"
-				:replyToUserId="replyToUserId"
+				:replyToUserId="userBeingRepliedTo.userId"
 				:userAvatar="userAvatar"
 				@submit="handleInlineSubmit($event)"
 			/>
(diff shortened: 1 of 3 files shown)


FlorianWendelborn:
@HansAuger can you resolve the conflicts?

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -55,7 +55,7 @@
 					:userAvatar="reply.userAvatar"
 					:userId="reply.userId"
 					:userName="reply.userName"
-					@_inlineDeleteClick="handleDelete($event, 'INLINE')"
+					@_inlineDeleteClick="(commentId) => handleDelete(commentId, false)"
```
FlorianWendelborn, line 58: This was previously `'INLINE'`, now it’s `isInline: false`. Seems wrong

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -73,19 +73,28 @@
...
 
 import CommentReply from './components/CommentReply.vue'
 import KtCommentInput from './KtCommentInput.vue'
 import { KottiComment } from './types'
-import { defaultParser } from './utilities'
 
-export default {
-	defaultParser,
+type UserData = Pick<KottiComment.PropsInternal, 'userName' | 'userId'>
+type SubmitEvent = {
```
FlorianWendelborn, line 91: Should probably be moved to `Kotti.Comment.Events.Submit`

FlorianWendelborn (approved):

FlorianWendelborn:
LGTM besides the boolean weirdness

HansAuger:
@FlorianWendelborn I fixed another thing and removed an internal variable because  it proved to be redundant. Please re-review the last two commits

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -95,75 +99,74 @@ export default {
...
+			emit('delete', payload)
+		}
 		return {
-			showInlineReply: false,
-			replyToText: null,
-			replyToUserId: Number,
-			isInlineEdit: false,
-			inlineMessageValue: '',
+			actionOptions: computed<Array<KottiButton.Props>>(() =>
+				compact([
```
carsoli, line 116: I really am not a fan of overuse of `lodash` utils just to remove `null`s here

this might have made more sense if we had falsible values because it apparently removes `''` and `false` and whatnot

So, I just either `.filter(Boolean)` at the end or not pushng to an array if it's not Editable
> you would have to declare the array as empty initially obviously and then conditionally push

some people have a third approach of pushing an empty array `[]`  in the false condition case  and then `flatMap`ping but I find this equally convoluted personally

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -95,75 +104,74 @@ export default {
...
+						? {
+								label: 'Delete',
+								onClick: () => handleDelete(props.id),
+								type: KottiButton.Type.DANGER,
+						  }
+						: null,
+				]),
+			),
+			cancelInlineEdit: () => {
+				inlineMessageValue.value = null
```
carsoli, line 138: you reset `inlineMessageValue` everytime you cancel the edit action

so just checking why you introduced this check: 
https://github.com/3YOURMIND/kotti/pull/562/commits/9bdc06b7e030544f1149bbd9957f44470b48c8a1#diff-9588d13c829fac8eaaab185abb19d1e652ba33ceec38a11da7f7fe5ac39d7924R126


it seems that every time you click `edit` you want to initialize the inlineEditMessage as `props.message`

but I guess you wanted to handle the case in the UI where the user clicks the `EDIT` while the edit is already active

so that you don't reset what they temporarily typed


I suggest the following then because it makes it clearer

convert the condition to `if(!isInlineEdit.value) return`
because it makes it clear that this is a redudant extra click

and then it won't execute the reset of the message to `props.message` OR the re-initialization of `isInlinedEdit` to `true`

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -95,75 +104,74 @@ export default {
...
+				isInlineEdit.value = false
+			},
+			handleDelete,
```
carsoli, line 141: can you explain why we have BOTH `delete` on the options of the comment and `inline`

the options on the comment are controlled by the user-app

e.g. in AMPI, depending on user permissions the `delete` option is disabled if the user doesn't have deletion rights


So my question is: is this `inlineDelete` controlled??
looks to me like a duplicate feature that most likely is not noticed by the user-app

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -95,75 +104,74 @@ export default {
...
+			handleEditConfirm: () => {
+				isInlineEdit.value = false
+				if (!inlineMessageValue.value) return
```
carsoli, line 144: make `inlineMessageValue.value === null` please

we don't support `''` initialization anymore, so we can be explicit

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -95,75 +104,74 @@ export default {
 		KtCommentInput,
 	},
 	props: makeProps(KottiComment.propsSchema),
-	data() {
-		return {
-			showInlineReply: false,
-			replyToText: null,
-			replyToUserId: Number,
```
carsoli, line 102: one day I'm going to stroke while reading one of those gems

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -95,75 +104,74 @@ export default {
...
-			if (!this.inlineMessageValue) return
-			this.$emit('edit', {
-				id: this.id,
-				message: this.inlineMessageValue,
-			})
-		},
+			},
+			handleInlineReplyClick: ({ userId, userName }: UserData) => {
+				showInlineReply.value = true
+				userBeingRepliedTo.value = { userId, userName }
```
carsoli, line 152: no need to destruct the argument, if we are passing it down as is 

arg can become `replyParentUser: UserData`
and assignment can become `userBeingRepliedTo.value = replyParentUser`

carsoli: or however you want to name the argument

packages/kotti-ui/source/kotti-comment/KtComment.vue
HansAuger: You are thinking one step ahead. `inlineMessageValue` is not bound by a `KtFieldTextArea` but by a `textarea` (which is designed to look like a `KtInlineEdit` lol). So the check is actually intentional since `''` is a possible value that might show up here.

packages/kotti-ui/source/kotti-comment/KtComment.vue
HansAuger: :pepe-hug:

packages/kotti-ui/source/kotti-comment/KtComment.vue
HansAuger: I  am not 100% I get your question. If a comment reply should be deleted, it emits a delete, since The KtComment is always its parent, it re-emits, so it can be handled by the user app. AKA, all `delete/edit/submit` events are emitted by `KtComment` and are processed by whoever implements this.

packages/kotti-ui/source/kotti-comment/KtComment.vue
HansAuger: I thought it made more sense to force those options to be empty when `isInlineEdit.value` is `true`

carsoli: but you didn't force the edit option to not be there, with the condition: `if (inlineMessageValue.value === null)`

you were just guarding before resetting the edit value

which meant you actually prevented the bug that if the user clicks once then starts typing THEN clicks again on the option, it wouldn't have reset the already-typed message to `null`

so I basically said, I actually agree with what you did, I just think it might be more readable to have `if(isInlineEdit) return`

because if the flag that already shows the user is editing is true, then there is no need to reset

carsoli (approved):

carsoli (approved):

PR #564, opened 2021-11-28
Refactor kt comment topic/composition api comment reply

HansAuger:
Step 2/4:

- [x] use zod for props schema and typing (#561)
- [ ] composition api   
  - [ ] KtComment (#562)
  - [ ] KtCommentInput (#563)
  - [ ] CommentReply    <--- We are here
- [ ] put css into component
- [ ] other clean ups if necessary

After refactor:
- [ ] Integrate Kotti Translations
- [ ] #253
- [ ] #438

### DIFF
--- a/packages/kotti-ui/source/kotti-comment/components/CommentReply.vue
+++ b/packages/kotti-ui/source/kotti-comment/components/CommentReply.vue
@@ -8,7 +8,7 @@
 			</div>
 			<div class="comment-reply__body">
 				<div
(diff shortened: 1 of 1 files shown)


HansAuger:
BTW, the usefulness of this component is debatable. It is basically `KtComment` with less features. So imo makes more sense to do a more rigorous refactor where a lot of the overlapping code is thrown out or even `KtComment` is entirely reused recursively.

Issue #423, opened 2021-04-30
Typescript check during build does not work


Isokaeder:
Typescript errors in `.vue` files do not get picked up on build, pipeline succeeds regardless.

FlorianWendelborn:
Turns out this is because of `@rollup/plugin-typescript` not supporting type checking 🥲

<https://github.com/rollup/plugins/tree/master/packages/typescript#faster-compiling>

Not much we can do here, just the normal perks of the Vue ecosystem. I tried out using `@rollup/plugin-sucrase`, but it fails miserably with handling Vue files as expected.

Once again overwhelmed by the awesomeness of the Vue ecosystem.

FlorianWendelborn:
https://github.com/johnsoncodehk/volar/tree/master/packages/vue-tsc is another interesting idea

PR #570, opened 2021-12-15
build(#423): Type checking for vue components

Isokaeder:
Use [rollup-plugin-typescript2](https://www.npmjs.com/package/rollup-plugin-typescript2). to find compilatioin fixes. Since it doesn't integrate in our. rollup config well without breaking the build, only apply the mistakes it finds for now.

OLD:
~~~
Disclaimer, this is still on major version 0, so it might be too early to fully integrate. This is more or less a proposal, in the end we might just take the fixes of typing issue from this PR.

>Rollup plugin for typescript with compiler errors.

Yay, compiler errors!

>This is a rewrite of original rollup-plugin-typescript, starting and borrowing from this fork.

link is broken, but it pointed to a repo of the same author.  My guess is it was integrated by rollup into `rollup/plugin-typescript` since the options of rpt2 is a superset of that.

>This version is somewhat slower than original, but it will print out typescript syntactic and semantic diagnostic messages (the main reason for using typescript after all).

For our case the build step gets tremendously slower (since all vue files are now type-checked which didn't happen before). This is very detrimental to the `yarn watch` workflow in this state. typecheck can be deactivated via setting `check: false` in the plugin options, but still runs slower than or current setup. 

I couldn't make the build work fully yet, right now it complains. about something like this:   :/
```
[!] Error: Could not resolve '../kotti-button' from source/kotti-inline-edit/KtInlineEdit.vue?rollup-plugin-vue=script.js
```
~~~

### DIFF
--- a/packages/kotti-ui/source/kotti-avatar/KtAvatar.vue
+++ b/packages/kotti-ui/source/kotti-avatar/KtAvatar.vue
@@ -17,7 +17,7 @@
 import { useTippy } from '@3yourmind/vue-use-tippy'
 import { Yoco } from '@3yourmind/yoco'
 import { computed, defineComponent, Ref, ref } from '@vue/composition-api'
-import { roundArrow } from 'tippy.js'
+import { Props as TippyProps, roundArrow } from 'tippy.js'
 
 import { makeProps } from '../make-props'
 
@@ -29,13 +29,14 @@ const useTooltip = (name: Ref<string | null>) => {
 
 	useTippy(
 		tooltipTriggerRef,
-		computed(() => ({
+		computed<Partial<TippyProps>>(() => ({
 			appendTo: () => document.body,
 			arrow: roundArrow,
-			content: name.value,
 			offset: [0, ARROW_HEIGHT],
 			theme: 'light-border',
-			...(name.value === null ? { trigger: 'manual' } : {}),
+			...(name.value === null
+				? { trigger: 'manual' }
+				: { content: name.value }),
 		})),
 	)
 
--- a/packages/kotti-ui/source/kotti-button/KtButton.vue
+++ b/packages/kotti-ui/source/kotti-button/KtButton.vue
@@ -28,7 +28,7 @@ export default defineComponent<KottiButton.PropsInternal>({
 		const hasSlot = computed(() => Boolean(slots.default))
 
 		return {
-			handleClick: (event) => emit('click', event),
+			handleClick: (event: Event) => emit('click', event),
 			hasSlot,
 			mainClasses: computed(() => ({
 				'kt-button': true,
--- a/packages/kotti-ui/source/kotti-comment/KtComment.vue
+++ b/packages/kotti-ui/source/kotti-comment/KtComment.vue
@@ -35,7 +35,7 @@
 					<div class="action__options">
 						<a
 							v-for="option in actionOptions"
-							:key="option.type"
+							:key="option.label"
 							@click="option.onClick"
 						>
 							<li>{{ option.label }}</li>
@@ -79,7 +79,6 @@ import { computed, defineComponent, ref } from '@vue/composition-api'
 import { KtAvatar } from '../kotti-avatar'
 import { KtButton } from '../kotti-button'
 import { KtButtonGroup } from '../kotti-button-group'
-import { KottiButton } from '../kotti-button/types'
 import { useTranslationNamespace } from '../kotti-i18n/hooks'
 import { makeProps } from '../make-props'
 
@@ -88,6 +87,10 @@ import KtCommentInput from './KtCommentInput.vue'
 import { KottiComment } from './types'
 
 type UserData = Pick<KottiComment.PropsInternal, 'userName' | 'userId'>
+type CommentAction = {
+	label: string
+	onClick: () => void
+}
 
 export default defineComponent<KottiComment.PropsInternal>({
 	name: 'KtComment',
@@ -105,16 +108,19 @@ export default defineComponent<KottiComment.PropsInternal>({
 		const userBeingRepliedTo = ref<UserData | null>(null)
 		const translations = useTranslationNamespace('KtComment')
 
-		const handleDelete = (commentId: number | string, isInline?: boolean) => {
+		const handleDelete = (
+			commentId: number | string | null,
+			isInline?: boolean,
+		) => {
 			const payload: KottiComment.Events.Delete = {
 				id: commentId,
-				parentId: isInline ? props.id : null,
+				parentId: isInline ? props.id ?? null : null,
 			}
 			emit('delete', payload)
 		}
 		return {
-			actionOptions: computed<Array<KottiButton.Props>>(() => {
-				const options = []
+			actionOptions: computed<Array<CommentAction>>(() => {
+				const options: Array<CommentAction> = []
 				if (isInlineEdit.value) return options
 				if (props.isEditable)
 					options.push({
@@ -123,13 +129,11 @@ export default defineComponent<KottiComment.PropsInternal>({
 							inlineMessageValue.value = props.message
 							isInlineEdit.value = true
 						},
-						type: KottiButton.Type.PRIMARY,
 					})
 				if (props.isDeletable)
 					options.push({
 						label: 'Delete',
-						onClick: () => handleDelete(props.id),
-						type: KottiButton.Type.DANGER,
+						onClick: () => handleDelete(props.id ?? null),
 					})
 				return options
 			}),
@@ -142,7 +146,7 @@ export default defineComponent<KottiComment.PropsInternal>({
 				isInlineEdit.value = false
 				if (inlineMessageValue.value === null) return
 				const payload: KottiComment.Events.Edit = {
-					id: props.id,
+					id: props.id ?? null,
 					message: inlineMessageValue.value,
 				}
 				emit('edit', payload)
--- a/packages/kotti-ui/source/kotti-comment/KtCommentInput.vue
+++ b/packages/kotti-ui/source/kotti-comment/KtCommentInput.vue
@@ -59,10 +59,10 @@ export default defineComponent<KottiCommentInput.PropsInternal>({
 				emit('submit', {
 					message: text.value,
 					replyToUserId: props.replyToUserId,
-					parentId: props.parentId,
+					parentId: props.parentId ?? null,
 				})
 				text.value = null
-				textarea.value.style.height = '1.2rem'
+				if (textarea.value !== null) textarea.value.style.height = '1.2rem'
 			},
 			replyButtonText: computed(() =>
 				props.isInline
--- a/packages/kotti-ui/source/kotti-comment/types.ts
+++ b/packages/kotti-ui/source/kotti-comment/types.ts
@@ -28,12 +28,12 @@ export namespace KottiComment {
 
 	export namespace Events {
 		export type Delete = {
-			id: string | number
+			id: string | number | null
 			parentId: string | number | null
 		}
 
 		export type Edit = {
-			id: string | number
+			id: string | number | null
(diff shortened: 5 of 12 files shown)


packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
```diff
@@ -72,9 +72,11 @@ export default defineComponent<KottiNavbar.PropsInternal>({
 		NavbarNotification,
 		NavbarQuickLink,
 	},
+	// @ts-expect-error: official declarations of @types/vue-clickaway do
+	// not satisfy the typechecker apparently
 	mixins: [clickaway],
 	props: makeProps(KottiNavbar.propsSchema),
-	setup(props, { emit }) {
+	setup(props: KottiNavbar.PropsInternal, { emit }) {
```
HansAuger, line 79: In case somebody is wondering, vetur for some reason decided that props is `unknown` here,  I guess because. of the nonempty `mixins`. somehow, since this happens only for components that rely on 'vue-clickaway'

packages/kotti-ui/rollup.config.js
```diff
@@ -35,6 +35,7 @@ const plugins = ({ enableVisualizer, module }) => [
 		needMap: false,
 	}),
 	typescript({
+		abortOnError: false,
```
HansAuger, line 38: I added this so I could see all compilation error at once.  If omitted, the build will stop at the first error thrown

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
```diff
@@ -72,9 +72,11 @@ export default defineComponent<KottiNavbar.PropsInternal>({
 		NavbarNotification,
 		NavbarQuickLink,
 	},
+	// @ts-expect-error: official declarations of @types/vue-clickaway do
```
FlorianWendelborn, line 75: I'm pretty sure I saw in your VSCode that this was detected as an unnecessary `@ts-expect-error`. It should therefore be a `@ts-ignore` since it's not always erroring

packages/kotti-ui/source/kotti-user-menu/KtUserMenu.vue
```diff
@@ -65,9 +65,11 @@ export default defineComponent<KottiUserMenu.PropsInternal>({
 	components: {
 		KtAvatar,
 	},
+	// @ts-expect-error: official declarations of @types/vue-clickaway do
```
FlorianWendelborn, line 68: Same

packages/kotti-ui/source/kotti-heading/KtHeading.vue
```diff
@@ -36,7 +36,7 @@ export default defineComponent({
 	},
 	setup(props, { emit }) {
 		return {
-			handleClick: (event) => {
+			handleClick: (event: Event) => {
```
FlorianWendelborn, line 39: Type should be more specific

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -28,7 +28,7 @@ export default defineComponent<KottiButton.PropsInternal>({
 		const hasSlot = computed(() => Boolean(slots.default))
 
 		return {
-			handleClick: (event) => emit('click', event),
+			handleClick: (event: Event) => emit('click', event),
```
FlorianWendelborn, line 31: Type should be more specific

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -105,16 +108,19 @@ export default defineComponent<KottiComment.PropsInternal>({
 		const userBeingRepliedTo = ref<UserData | null>(null)
 		const translations = useTranslationNamespace('KtComment')
 
-		const handleDelete = (commentId: number | string, isInline?: boolean) => {
+		const handleDelete = (
+			commentId: number | string | null,
```
FlorianWendelborn, line 112: I don't think null makes sense here. How can you delete a comment without an ID?

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -105,16 +108,19 @@ export default defineComponent<KottiComment.PropsInternal>({
...
+			isInline?: boolean,
+		) => {
 			const payload: KottiComment.Events.Delete = {
 				id: commentId,
```
FlorianWendelborn, line 110: Probably same in this type

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -105,16 +108,19 @@ export default defineComponent<KottiComment.PropsInternal>({
...
-				parentId: isInline ? props.id : null,
+				parentId: isInline ? props.id ?? null : null,
```
FlorianWendelborn, line 117: Can't this be handled via a default on the `id` prop?

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -105,16 +108,19 @@ export default defineComponent<KottiComment.PropsInternal>({
...
 			}
 			emit('delete', payload)
 		}
 		return {
-			actionOptions: computed<Array<KottiButton.Props>>(() => {
-				const options = []
+			actionOptions: computed<Array<CommentAction>>(() => {
```
FlorianWendelborn, line 122: This type annotation is implied and could be removed

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -123,13 +129,11 @@ export default defineComponent<KottiComment.PropsInternal>({
...
 							isInlineEdit.value = true
 						},
-						type: KottiButton.Type.PRIMARY,
 					})
 				if (props.isDeletable)
 					options.push({
 						label: 'Delete',
-						onClick: () => handleDelete(props.id),
-						type: KottiButton.Type.DANGER,
+						onClick: () => handleDelete(props.id ?? null),
```
FlorianWendelborn, line 136: Same defaulting here

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -142,7 +146,7 @@ export default defineComponent<KottiComment.PropsInternal>({
 				isInlineEdit.value = false
 				if (inlineMessageValue.value === null) return
 				const payload: KottiComment.Events.Edit = {
-					id: props.id,
+					id: props.id ?? null,
```
FlorianWendelborn, line 149: Same defaulting

packages/kotti-ui/source/kotti-comment/KtCommentInput.vue
```diff
@@ -59,10 +59,10 @@ export default defineComponent<KottiCommentInput.PropsInternal>({
 				emit('submit', {
 					message: text.value,
 					replyToUserId: props.replyToUserId,
-					parentId: props.parentId,
+					parentId: props.parentId ?? null,
 				})
 				text.value = null
-				textarea.value.style.height = '1.2rem'
+				if (textarea.value !== null) textarea.value.style.height = '1.2rem'
```
FlorianWendelborn, line 65: I'd rather throw and error. This shouldn't actually happen at runtime unless I'm misunderstanding

packages/kotti-ui/source/kotti-comment/KtCommentInput.vue
```diff
@@ -59,10 +59,10 @@ export default defineComponent<KottiCommentInput.PropsInternal>({
 				emit('submit', {
 					message: text.value,
 					replyToUserId: props.replyToUserId,
-					parentId: props.parentId,
+					parentId: props.parentId ?? null,
```
FlorianWendelborn, line 62: Default

packages/kotti-ui/source/kotti-comment/types.ts
```diff
@@ -28,12 +28,12 @@ export namespace KottiComment {
 
 	export namespace Events {
 		export type Delete = {
-			id: string | number
+			id: string | number | null
```
FlorianWendelborn, line 31: Doesn't make sense to delete a comment without an id IMO

packages/kotti-ui/source/kotti-comment/types.ts
```diff
@@ -28,12 +28,12 @@ export namespace KottiComment {
...
 			parentId: string | number | null
 		}
 
 		export type Edit = {
-			id: string | number
+			id: string | number | null
```
FlorianWendelborn, line 36: Doesn't make sense to edit a comment without an id IMO

packages/kotti-ui/source/kotti-filters/KtFilters.vue
```diff
@@ -161,19 +161,18 @@ export default defineComponent<KottiFilters.InternalProps>({
 		}
 		const toggleListVisibility = () => {
 			isListVisible.value = !isListVisible.value
-			if (isListVisible.value) tippyInstanceRef.value.show()
+			if (isListVisible.value) tippyInstanceRef.value?.show()
```
FlorianWendelborn, line 164: Should probably throw an error instead

FlorianWendelborn (changes requested):

packages/kotti-ui/source/kotti-avatar/KtAvatar.vue
```diff
@@ -29,13 +29,14 @@ const useTooltip = (name: Ref<string | null>) => {
 
 	useTippy(
 		tooltipTriggerRef,
-		computed(() => ({
+		computed<Partial<TippyProps>>(() => ({
```
carsoli, line 32: locally check this with @FlorianWendelborn right now

`useTippy` accepts the 2nd argument (`options`) to be `Ref<Partial<Props>>`


can you elaborate on why this was necessary?

packages/kotti-ui/source/kotti-comment/KtCommentInput.vue
carsoli: this is actually dead

carsoli: i also remember leaving a comment on another `KtComment` refactor for this and actually testing it, and it proved dead

but I'd reverify the testing

packages/kotti-ui/source/kotti-field-select/KtFieldSingleSelectRemote.vue
```diff
@@ -124,10 +124,9 @@ export default defineComponent({
 
 		const { tippy } = useTippy(
 			tippyTriggerRef,
-			computed(() => ({
+			computed<Partial<TippyProps>>(() => ({
```
carsoli, line 127: same question as above , why did we need it

packages/kotti-ui/source/kotti-filters/KtFilters.vue
carsoli: I think it's a must

we have it within an `onMounted > watchEffect` for the selects 

but for this, it can just suffice, to have a function that gets the `tippyRef` as an argument and throws if it doesn't exist or return it if it does

packages/kotti-ui/source/kotti-filters/KtFilters.vue
```diff
@@ -161,19 +161,18 @@ export default defineComponent<KottiFilters.InternalProps>({
...
-				tippyInstanceRef.value.hide()
+				tippyInstanceRef.value?.hide()
 				isAddingFilter.value = false
 			}
 		}
 
 		useTippy(
 			listTriggerRef,
-			computed(() => ({
+			computed<Partial<TippyProps>>(() => ({
```
carsoli, line 173: same as https://github.com/3YOURMIND/kotti/pull/570/files#r772348663

PR #571, opened 2021-12-15
feat(kt-filters): support data-test attr

santiagoballadares:


packages/kotti-ui/source/kotti-filters/components/FilterActions.vue
```diff
@@ -1,12 +1,14 @@
...
 	<div class="kt-filter__actions">
 		<ButtonLink
+			:data-test="dataTest ? `${dataTest}-add-filter` : null"
 			:isDisabled="isAddDisabled"
 			:isLoading="isLoading"
 			:label="translations.addFilterLabel"
 			@click="handleAdd"
 		/>
 		<ButtonLink
+			:data-test="dataTest ? `${dataTest}-clear-all` : null"
```
FlorianWendelborn, line 11: Should either be `:dataTest=` or `: undefined` depending on if this is intended to be passed as a prop or as a html attribute

packages/kotti-ui/source/kotti-filters/KtFilters.vue
```diff
@@ -3,12 +3,14 @@
...
 			v-if="hasSearchColumn"
 			:column="searchColumn"
+			:dataTest="dataTest"
 			:filter="searchValue"
 			:isLoading="isLoading"
 			@input="setSearchFilter"
 		/>
 		<div ref="listTriggerRef">
 			<ButtonLink
+				:data-test="dataTest ? `${dataTest}-show-filters` : null"
```
FlorianWendelborn, line 13: Should either be `:dataTest=` or `: undefined` depending on if this is intended to be passed as a prop or as a html attribute

packages/documentation/pages/usage/components/filters.vue
```diff
@@ -8,6 +8,7 @@
     			<KtFilters
     				v-model="filters"
     				:columns="componentProps.columns"
+    				dataTest="kt-filters"
```
FlorianWendelborn, line 11: Won’t work with more than 1 instance of `KtFilters`

FlorianWendelborn (changes requested):

packages/documentation/pages/usage/components/filters.vue
santiagoballadares: This is the usage example of `KtFilters`. Isn't the user supposed to make sure to provide unique strings for `dataTest` when using multiples instances of KtFilters?

FlorianWendelborn (approved):

Issue #249, opened 2020-08-03
Add KtFieldCurrency


killakalle:
- [x] Extend `KtFieldNumber` First:
  - [x] Support parsing all 3 decimal separators as one
  - [x] Support decimal separator through translation context
  - [x] Remove the 3 decimal places cap on the number field 
     - Make it customizable through a prop - and default to `3`
- [ ] Implement KtFieldCurrency
  - [ ] Copy paste the number field
  - [ ] Split the common logic
  - [ ] Support Trailing Zeroes as many as DP 
  - [ ] Drop support for prefix/suffix/icons on KtFieldCurrency
  - [ ] add a utility that decides on the position of the currency (prefix vs suffix)
  - [ ] Drop support for changeButtons on KtFieldCurrency
  
- [ ] Add KtFieldCurrency support in KtFilters

---

**Is your feature request related to a problem? Please describe.**
As an AMPI admin user I want to be able to use Currency fields on forms, so that AMPI users can enter currency amounts.

**Describe the solution you'd like**
- Text input field that allows only number input
- Decimal number input possible with max two decimals (currencies with other decimal rules are not required for the time being)
- Currency symbol (e.g. €) or Currency code (eg. EUR) shown
- Possible to configure currency symbol or code (Eg. EUR, USD, GBP, ...)

**Additional context**
This is an existing input field in AMPI and already in use by customers.

Current currency field in AMPI
![image](https://user-images.githubusercontent.com/30682796/89173881-354fa480-d585-11ea-9060-9d5e0eed788e.png)

---

Related Internal Issues:
- ~~<https://3yourmind.atlassian.net/browse/B3-8561>~~
- <https://3yourmind.atlassian.net/browse/B3-8987>
- <https://3yourmind.atlassian.net/browse/B3-8988>
- <https://3yourmind.atlassian.net/browse/B3-8764>

PR #572, opened 2021-12-16
feat(#249): KtFieldNumber Enhancements

carsoli:
Preliminary Steps before introducing `KtFieldCurrency`

As per description of #249, this is what is covered in this PR's scope:
- [x] Extend `KtFieldNumber` First:
  - [x] Support parsing all 3 decimal separators as one
  - [x] Support decimal separator through translation context
  - [x] Remove the 3 decimal places cap on the number field 
     - Make it customizable through a prop - and default to `3`

### DIFF
--- a/packages/kotti-ui/source/kotti-field-number/constants.ts
+++ b/packages/kotti-ui/source/kotti-field-number/constants.ts
@@ -30,7 +30,7 @@ export const KOTTI_FIELD_NUMBER_PROPS = {
  */
 export const DECIMAL_SEPARATORS_CHARACTER_SET = [
 	'[',
-	...Object.values(DecimalSeparator).map((x) => `\\${x}`),
+	...Object.values(DecimalSeparator),
 	']',
 ].join('')
 
@@ -40,12 +40,15 @@ export const STRINGS_THAT_ARE_TREATED_AS_NULL = [
 	'+',
 	'',
 ]
+
 export const LEADING_ZEROES_REGEX = new RegExp(
 	`^0+([1-9]|0${DECIMAL_SEPARATORS_CHARACTER_SET}?)`,
 )
+
 export const TRAILING_ZEROES_REGEX = new RegExp(
 	`${DECIMAL_SEPARATORS_CHARACTER_SET}0*$|(${DECIMAL_SEPARATORS_CHARACTER_SET}[0-9]*[1-9])0+$`,
 )
+
 export const VALID_REGEX = (decimalPlaces: number) =>
 	new RegExp(
 		`^[+-]?(0?|([1-9][0-9]*))?(${DECIMAL_SEPARATORS_CHARACTER_SET}[0-9]{0,${decimalPlaces}})?$`,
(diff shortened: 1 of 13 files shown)


packages/kotti-ui/source/kotti-i18n/types.ts
```diff
@@ -31,6 +37,7 @@ export namespace KottiI18n {
 	export type Props = {
 		locale: SupportedLanguages
 		messages: DeepPartial<Messages>
+		numberFormat: NumberFormat
```
FlorianWendelborn, line 40: Actually, we should probably allow this to be a DeepPartial for future extensibility

packages/kotti-ui/source/kotti-field-number/constants.ts
```diff
@@ -23,22 +25,28 @@ export const KOTTI_FIELD_NUMBER_PROPS = {
...
 
-export const DECIMAL_PLACES = 3
-// FIXME: should we consider manually setting the decimal separator using the translation context
-export const DECIMAL_SEPARATOR = (1.1).toLocaleString().replace(/\d/g, '')
+/**
+ * RegExp character set for use within other regular expressions
+ */
+export const DECIMAL_SEPARATORS_CHARACTER_SET = [
+	'[',
+	...Object.values(DecimalSeparator).map((x) => `\\${x}`),
```
FlorianWendelborn, line 33: I'm not sure if escaping is necessary in character sets

FlorianWendelborn (approved):

HansAuger:
Not a blocker, but it would make sense to add single select field to `packages/documentation/pages/usage/components/form.vue` another select field to pick a comma separator net to the language. Right now there is no number field on that page to see the effect though, that would need to be added a well.

packages/kotti-ui/source/kotti-field-number/constants.ts
carsoli: changed in latest

packages/kotti-ui/source/kotti-i18n/KtI18nContext.ts
```diff
@@ -8,11 +10,18 @@ const KtI18nContext = defineComponent({
 	props: {
 		locale: { required: true, type: String },
 		messages: { default: () => ({}), type: Object },
```
carsoli, line 12: : KottiI18n.Messages

HansAuger:
LGTM, it behaves a lot better than before.

One thing that still happens is that if you enter a number like `1.001` and then delete the last digit,  you don't get `1.00` but `1`. If you think we can deal with that scenario in the scope of this PR, that would be amazing to fix. But also not a blocker, since this behavior already existed before and all other problems seem to be fixed.

HansAuger (approved):

HansAuger:
@carsoli The only issue I have is that we then do not document this feature at all. Not a super big deal, we also don't document the `messages` prop of `KtI18NContext for example. But something to keep in mind.

carsoli:
> One thing that still happens is that if you enter a number like 1.001 and then delete the last digit, you don't get 1.00 but 1. If you think we can deal with that scenario in the scope of this PR, that would be amazing to fix. But also not a blocker, since this behavior already existed before and all other problems seem to be fixed.

@HansAuger the `.00` would be handled within the currency split - can be exposed as an extra prop IMO but now that you mention it, let me check if it was on the list of tasks just so we don't forget.

Merging for now.

PR #573, opened 2021-12-17
Do not import lodash

HansAuger:
Removes all instances of imports that pull all of lodash into our package.

```ts
// import { someUtil } from 'lodash'
import someUtil from 'lodash/someUtil' 
```

I also removed it from documentation which I know is not necessary, but I thought is was better to be consistent across the app.

### DIFF
--- a/packages/kotti-ui/source/kotti-user-menu/KtUserMenu.vue
+++ b/packages/kotti-ui/source/kotti-user-menu/KtUserMenu.vue
@@ -79,7 +79,7 @@ export default defineComponent<KottiUserMenu.PropsInternal>({
 			isMenuShow,
 			isNarrow,
 			parsedSections: computed(() =>
-				KottiUserMenu.propsSchema.shape.sections.parse(props.sections),
+				KottiUserMenu.propsSchema.shape.sections.safeParse(props.sections),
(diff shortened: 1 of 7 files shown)


FlorianWendelborn:
Did you test if this actually decreases the bundle size?

packages/kotti-ui/source/kotti-user-menu/KtUserMenu.vue
```diff
@@ -79,7 +79,7 @@ export default defineComponent<KottiUserMenu.PropsInternal>({
 			isMenuShow,
 			isNarrow,
 			parsedSections: computed(() =>
-				KottiUserMenu.propsSchema.shape.sections.parse(props.sections),
+				KottiUserMenu.propsSchema.shape.sections.safeParse(props.sections),
```
FlorianWendelborn, line 82: Bug

FlorianWendelborn (changes requested):

HansAuger:
@FlorianWendelborn True.  I did run `build:kotti:esm/build:kotti:cjs` and looked at the resulting bundle and actually saw a small increase. So I guess  this is a failja

PR #598, opened 2022-02-17
feature(KtFilters): Use KtFieldCurrency

carsoli:


packages/documentation/pages/usage/components/filters.vue
```diff
@@ -183,9 +196,6 @@ export default defineComponent({
 							Kotti.Filters.Operation.Currency.LESS_THAN_OR_EQUAL,
 							Kotti.Filters.Operation.Currency.IS_EMPTY,
 						],
-						prefix: settings.value.prefix,
```
carsoli, line 186: what why did you drop it from the return

you still needed for the relevant fields (float)

packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
```diff
@@ -121,21 +124,35 @@ export default defineComponent<{
...
 				: translations.value.andLabel,
 		)
-		const operationOptions = computed<KottiFieldSingleSelect.Props['options']>(
-			() => {
-				if (!props.column) return []
-				return getOperationOptions(props.column)
-			},
-		)
+		const operationOptions = computed(() => {
+			if (!props.column) return []
```
carsoli, line 139: I'd say this check belongs in the `getOperationOptions` function

because wherever it's used, we would need to add this pre-check

so might as well add it to the util and account for a potential `null` arg

packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
```diff
@@ -121,21 +124,35 @@ export default defineComponent<{
...
+			return getOperationOptions(props.column)
+		})
 		const valueComponent = computed(() => {
 			if (!props.column?.type) return null
 			return getValueComponent(props.column.type)
 		})
+		const valueContainerClasses = computed(() => ({
+			'kt-filter__list__row__value-field': true,
+			'kt-filter__list__row__value-field--is-vertically-centered':
+				props.column?.type === KottiFilters.FilterType.BOOLEAN,
```
carsoli, line 149: why this is logically specific to `BOOLEAN` type fields

packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
```diff
@@ -146,39 +163,25 @@ export default defineComponent<{
...
 			switch (props.column?.type) {
-				case KottiFilters.FilterType.CURRENCY:
 				case KottiFilters.FilterType.FLOAT:
 					return props.column.step
 				case KottiFilters.FilterType.INTEGER:
-				// fall through
-				default:
 					return 1
+				default:
+					return null
```
carsoli, line 177: can you type cast all returns from
`valueStep`, `valuePrefix`, `valueSuffix`

this will make us avoid this bug in the future when someone copies the implementation for "additionalProps"

so they can return `null` when the prop is not relevant

packages/documentation/pages/usage/components/filters.vue
santiagoballadares: This is the definition for the Currency column in the usage file for filters. KtFilters uses step, prefix and suffix only when column type is FLOAT.

packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
```diff
@@ -146,39 +163,25 @@ export default defineComponent<{
 			}
 		})
 		const valuePrefix = computed<KottiFieldNumber.Props['prefix']>(() =>
-			props.column?.type === KottiFilters.FilterType.CURRENCY ||
 			props.column?.type === KottiFilters.FilterType.FLOAT
 				? props.column.prefix
 				: null,
 		)
 		const valueStep = computed<KottiFieldNumber.Props['step']>(() => {
```
carsoli, line 170: logical question

was this only exposed from the filters API when we had the issue of having to add the `step` in order to allow for the field validation to work

if u remember internally we had a check `isStepMultiple` that was validating `value` for `KtFieldNumber` and before using `big` we had to pass `step` as a hack for it to work
if currently the problem is adding a value that has different decimal places and you want kotti to validate just that, then you need to expose decimalPlaces

and i wonder if it's now supposed to change this to expose `decimalPlaces` instead

`step` was there to logically allow the `+/-` buttons to add a certain step value

do u actually expose `+/-` or do you use `hideChangeProps` on `KtFieldNumber`

if you use `hideChangeButtons` for field of type `Float` then you don't need to expose `step`, you need to expose `decimalPlaces`
and if you say `decimalPlaces` is `3` then that's what we will support

carsoli: I checked the user-app

you don't use `hideChangeButtons` on number/float types


so my question is narrowed down to

do you need to limit `decimalPlaces` and thus validity of input on float numbers ?
By default it's `3`

I'm guessing internally for the number, you enforce `1` or you enforce the step / min / max to guarantee that's integer increments
> if you enforce the step, please enforce the `decimalPlaces` to `1` because the step by default is `1`

packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
```diff
@@ -121,21 +124,35 @@ export default defineComponent<{
...
 		})
+		const valueContainerClasses = computed(() => ({
+			'kt-filter__list__row__value-field': true,
+			'kt-filter__list__row__value-field--is-vertically-centered':
+				props.column?.type === KottiFilters.FilterType.BOOLEAN,
+		}))
+		const valueCurrency = computed(() =>
+			props.column?.type === KottiFilters.FilterType.CURRENCY
+				? props.column.currency
+				: null,
```
FlorianWendelborn, line 154: Should be `: undefined` so that the prop isn’t passed at all to the components that don’t support it

FlorianWendelborn (approved):
LGTM besides the one small thing I commented

Please squash when you merge since there’s still a WIP commit in the PR

packages/documentation/pages/usage/components/filters.vue
carsoli: so it's not used on template level directly?

packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
carsoli: then it should be the case for `valuePrefix`, `valueSuffix`, `valueStep` as well

packages/documentation/pages/usage/components/filters.vue
santiagoballadares: No, there are computed properties for step, prefix and suffix in FilterRow.vue

packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
santiagoballadares: Right, `step` was added due to the floating point precision issue with the `+/-` buttons. However, I'll keep it as `hideChangeButtons` is not being used.

Will expose `decimalPlaces` as well for `Float` types and `Integer` types will default to: `step = 1` and `decimalPlaces = 0`. 

Also, `prefix`, `suffix`, `decimalPlaces`, `step`, `currency` and `options` will default to `undefined` for types that don't support them.

packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
santiagoballadares: Now `prefix`, `suffix`, `decimalPlaces`, `step`, `currency` and `options` will default to `undefined` for types that don't support them.

packages/documentation/pages/usage/components/filters.vue
```diff
@@ -167,12 +199,13 @@ export default defineComponent({
 							Kotti.Filters.Operation.Float.LESS_THAN_OR_EQUAL,
 							Kotti.Filters.Operation.Float.IS_EMPTY,
 						],
-						prefix: settings.value.prefix,
-						step: 0.001,
-						suffix: settings.value.suffix,
+						prefix: settings.value.numberPrefix ?? undefined,
```
FlorianWendelborn, line 202: Most of these can probably stay `null`. The `undefined` only comes into play when you’re potentially passing these props to components that shouldn’t get them.

I’m fine with keeping this as-is though as it’s not technically wrong

FlorianWendelborn (approved):

PR #599, opened 2022-02-17
fix(KtFieldToggleGroup): value prop is initialized

carsoli:
..as null

which means that on first render, if a value
..is still initialized as null, field.currentValue
could be `null`

therefore, guard against null[option.key]


packages/kotti-ui/source/kotti-field-toggle/KtFieldToggleGroup.vue
```diff
@@ -82,7 +82,7 @@ export default defineComponent<KottiFieldToggleGroup.PropsInternal>({
 			optionsWithChecked: computed(() =>
```
carsoli, line 82: if at any point `props.value` is `null`

and the options array looks like this `[ { label: 'a', value: false }, { label: 'b', value: true } ]`

and this will return ` [ { label: 'a', value: null }, { label: 'b', value: null } ]`

which is one way of doing this rather than let the app handle the initialization for all keys to be `null`

FlorianWendelborn (approved):

Issue #592, opened 2022-02-16
KtField: remove `isCorrectDataType` from KottiFields


carsoli:
_Originally posted by @carsoli in https://github.com/3YOURMIND/kotti/pull/581#discussion_r807818061_

PR #600, opened 2022-02-17
chore(#592): delete isCorrectDataType check

carsoli:
zod validates props now, with a readable error stack
therefore, all the validation logic for props
that were initially introduced can be dropped

### DIFF
--- a/packages/kotti-ui/source/kotti-field/tests.ts
+++ b/packages/kotti-ui/source/kotti-field/tests.ts
@@ -1,6 +1,6 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 require('jsdom-global')()
-import { defineComponent, ref } from '@vue/composition-api'
+import { defineComponent, ref, SetupContext } from '@vue/composition-api'
 import { shallowMount } from '@vue/test-utils'
 import { z } from 'zod'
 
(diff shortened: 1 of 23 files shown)


packages/kotti-ui/source/kotti-field/tests.ts
```diff
@@ -106,7 +70,7 @@ describe('useField', () => {
 	it('props.value gets deepCloned', async () => {
 		const VALUE_REFERENCE = { something: 'something' }
 
-		const wrapper = shallowMount(TestComponentObject, {
```
FlorianWendelborn, line 109: This is wrong, since this test needs props.value to be something worth deep-cloning.

packages/kotti-ui/source/kotti-form/tests.ts
```diff
@@ -139,7 +97,7 @@ describe('KtForm', () => {
 		it('should deepClone values before passing it to field, to prevent deep mutation madness', () => {
 			const DEEP_VALUE_REFERENCE = { deep: true }
 
-			const wrapper = mount(TestForm2, {
```
FlorianWendelborn, line 142: This is also wrong, as this test is testing deep clone behavior as well

FlorianWendelborn (approved):
Approved under the assumption that the 2 issues will be fixed before merging

packages/kotti-ui/source/kotti-field/tests.ts
carsoli: the value reference is still deep cloned, the only difference between `TestComponentObject` and `TestComponent` was the `isCorrectDataType` check

packages/kotti-ui/source/kotti-form/tests.ts
carsoli: same as https://github.com/3YOURMIND/kotti/pull/600#discussion_r809985330

PR #621, opened 2022-05-05
fix(docs/layouts): layouts/introduction had broken links in md

carsoli:
now they redirect properly


carsoli:
Go to Layouts/Introduction 

There are two links in the main page (Navbar / Actionbar )

Previously if you clicked on them, they failed to redirect you because the nuxt-link generated was wrong

now, fixed by having a relative path and you can test the redirection

FlorianWendelborn (approved):

PR #630, opened 2022-05-31
feature(makeProps): Add Support for z.any()

FlorianWendelborn:


carsoli:
@FlorianWendelborn mark as ready when the unit test for this feature is added

thanks

PR #634, opened 2022-06-07
Fix click behavior on kt-popover

carsoli:
remove the clickBehavior prop from KtPopover and change the internal implementation

### DIFF
--- a/packages/kotti-ui/source/kotti-comment/components/CommentActionsOptions.vue
+++ b/packages/kotti-ui/source/kotti-comment/components/CommentActionsOptions.vue
@@ -2,7 +2,6 @@
 	<KtPopover
 		v-if="options.length > 0"
 		class="kt-comment__actions__more-options"
-		:clickBehavior="clickBehavior"
 		:options="options"
 		placement="bottom"
 		trigger="click"
@@ -21,16 +20,11 @@ export default defineComponent<{
 	options: Array<Kotti.Popover.Props['options']>
 }>({
 	name: 'CommentActionsOptions',
-	props: {
-		options: { type: Array, required: true },
-	},
 	components: {
 		KtPopover,
 	},
-	setup() {
-		return {
-			clickBehavior: Kotti.Popover.ClickBehavior.HIDE_ON_CLICK_AWAY,
-		}
+	props: {
+		options: { type: Array, required: true },
 	},
 })
 </script>
--- a/packages/kotti-ui/source/kotti-field-date/KtFieldDateTime.vue
+++ b/packages/kotti-ui/source/kotti-field-date/KtFieldDateTime.vue
@@ -20,12 +20,19 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, ref, Ref, computed } from '@vue/composition-api'
+import {
+	defineComponent,
+	ref,
+	Ref,
+	computed,
+	inject,
+} from '@vue/composition-api'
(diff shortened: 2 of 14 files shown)


packages/kotti-ui/source/kotti-comment/components/CommentActionsOptions.vue
```diff
@@ -24,13 +22,5 @@ export default defineComponent<{
 	props: {
 		options: { type: Array, required: true },
 	},
-	components: {
-		KtPopover,
```
FlorianWendelborn, line 28: Why?

packages/kotti-ui/source/kotti-field/mixins.scss
```diff
@@ -438,7 +438,7 @@
 			/* stylelint-disable */
 			font-size: map.get($fontSizes, $type);
 
-			.kt-field__input-container {
```
FlorianWendelborn, line 441: This seems pretty random

packages/kotti-ui/source/kotti-field/KtField.vue
```diff
@@ -251,13 +251,11 @@ export default defineComponent<{
 					}
 				}
 
-				&:focus-within {
```
FlorianWendelborn, line 254: Why did this have to change?

FlorianWendelborn (approved):

packages/kotti-ui/source/kotti-comment/components/CommentActionsOptions.vue
carsoli: accidental, fixed on the MR where i brought back all the internal imports but forgot to rebase properly i guess

packages/kotti-ui/source/kotti-field/KtField.vue
carsoli: because the hooks hackily append kotti classes to the el-date inner input fields

since we now are context aware, the date-picker is appended to the parent, and if the specificity is not as high, focusing on the parent field was focusing on all the inner input fields (because of focus-within)

test it out in the browser, by bringing back those classes as they were, and you will see how the behavior is altered

packages/kotti-ui/source/kotti-field-date/KtFieldDateTime.vue
```diff
@@ -74,10 +81,17 @@ export default defineComponent<KottiFieldDateTime.PropsInternal>({
...
+
 		return {
 			elDateTimePickerProps: computed(
 				(): Partial<ElDate> => ({
 					...EL_DATE_TIME_PROPS,
+					/**
+					 * @see {@link https://github.com/ElemeFE/element/blob/v2.13.1/packages/date-picker/src/picker.vue#L334)}
+					 * */
+					// @ts-expect-error (exposed through mixin on picker.vue on element-ui's implementation)
+					appendToBody: isInPopover ? false : true,
```
HansAuger, line 94:
```suggestion
					appendToBody: !isInPopover,
```

HansAuger (approved):

packages/kotti-ui/source/kotti-field/mixins.scss
carsoli: it's not, all is related to when you have two `.kt-fields_input-container` that can be influened by the same specifity 

happens when you use the datetime field and datetime range inside popper because the parent becomes the field itself and affects the `input` with this class inside the picker itself

Issue #636, opened 2022-06-14
KtModal not attached to body


santiagoballadares:
**Describe the bug**
KtModal is not attached to body, but rather rendered within its parent component.

**To Reproduce**

1. Go to KtModal page: Components > Modal
2. Click on 'Open Modal' button
3. Open Developer Tools
4. Locate the Modal in the Document tree
5. See that the modal is not attached to body

**Expected behavior**
The modal is attached to body.

**Screenshots**
<img width="1052" alt="Screen Shot 2022-06-14 at 16 39 29" src="https://user-images.githubusercontent.com/744091/173693917-a4970213-c691-43be-992b-16b6ef2d1bf6.png">

**Desktop (please complete the following information):**

- OS: Any
- Browser: Any
- Version: Latest

**Smartphone (please complete the following information):**

- Device: Any
- OS: Any
- Browser: Any
- Version: Latest

PR #637, opened 2022-06-14
fix(#636): Attach KtModal to body

santiagoballadares:


packages/kotti-ui/source/kotti-modal/KtModal.vue
```diff
@@ -29,6 +31,12 @@ export default defineComponent<KottiModal.PropsInternal>({
 	name: 'KtModal',
 	props: makeProps(KottiModal.propsSchema),
 	setup(props, { emit, slots }) {
+		const wrapperRef = ref<HTMLElement | null>(null)
+
+		onMounted(() => {
+			if (wrapperRef.value) window.document.body.appendChild(wrapperRef.value)
```
carsoli, line 37: can u throw if `wrapperRef.value` is null

```
if(wrapperRef.value === null) 
  throw new Error('KtModal: Unbound wrapperRef')
window.document.body.appendChild(wrapperRef.value)
```

carsoli (dismissed):

packages/kotti-ui/source/kotti-modal/KtModal.vue
FlorianWendelborn, line 37: I don't think this is legal to just do. What happens when content updates and Vue needs to re-render something? Is there any place in the Vue documentation that encourages this? How did you find out about this in the first place?

I don't know how it's broken, but this is something that you usually shouldn't be able to do with a VDOM

FlorianWendelborn (dismissed):

packages/kotti-ui/source/kotti-modal/KtModal.vue
FlorianWendelborn: Document shouldn’t change with navigation and memory leaks are removed after a refresh.

Also, please keep in mind that this is just an example of what’s wrong with manipulating VDOM with native DOM operations. Anything could be wrong with this, it’s just the first thing I found.

santiagoballadares: @FlorianWendelborn @carsoli about the memory leak, I missed a removeChild on Unmount. I fixed that and tested locally and with Yoda. Although it seems to work even with data being updated inside the modal, I agree that manipulating the VDom is not a good approach. 
In any case, the PR is now updated with an implementation using Tippy.

packages/kotti-ui/source/kotti-modal/KtModal.vue
```diff
@@ -29,17 +32,70 @@ export default defineComponent<KottiModal.PropsInternal>({
...
+
+					if (contentRef.value === null)
+						throw new Error('KtModal: Unbound contentRef')
+
+					popper.appendChild(contentRef.value)
+
+					return { popper }
+				},
+				trigger: 'manual',
+				zIndex: 1000,
```
carsoli, line 71: I'm only concerned about this, why did you need it

carsoli (dismissed):

carsoli (approved):

carsoli (approved):

PR #638, opened 2022-06-16
feat(KtComment): add m flag to parser regex

carsoli:


FlorianWendelborn:
Why was this changed? As far as I can tell, this should literally do the same thing.

carsoli:
@florianwendelborn I'm yet to test if it makes any change 

but m should be multiline, how is it the same ?

FlorianWendelborn:
@carsoli at least in regexr.com I can’t reproduce any difference between the two.

As far as I understand the `m` flag, all it does it run the regular expression separately on every line in the string. Since we use `g` already, treating the string as one long string with multiple matches for `/n` and treating it as multiple lines with one match of `/n` each doesn’t change anything.

`m` is usually used for regular expressions that match something more like `/\s+$/m` which should match all trailing whitespace. The `m` would be important here as `g` would select all whitespace (without `$`) and only the last line’s whitespace (with `$`)

But I don’t fully understand why you proposed this change, so please feel free to provide an example where this actually makes a difference

packages/kotti-ui/source/kotti-comment/utilities.ts
```diff
@@ -2,4 +2,4 @@ import escape from 'lodash/escape'
 
 export const defaultParser = (message: string) => escape(message)
 export const defaultPostEscapeParser = (message: string) =>
-	message.replace(/\n/g, '</br>')
+	message.replace(/\n/gm, '</br>')
```
FlorianWendelborn, line 5: However, upon seeing this thing again... why is this `</br>` and not either `<br />` or `<br>`?

FlorianWendelborn:
Oh, and if you somehow dislike the current regex and safeParse "solution", you could also decide to implement the alternative I proposed in #253 instead

PR #639, opened 2022-06-21
fix(zod): force resolution to version 3.14.1 due to bug in 3.14.3

carsoli:
commit `f59df39` on https://github.com/colinhacks/zod/releases/tag/v3.14.3 
...adds logic to optimize parsing of types by calling
...a lazy parser when nested schemas are defined.
Unfortunately, the context from the caller function (parser) is not bound
...properly and the lazy parser gets a partial context
This leads to an undefined access attempt because a necessary
...attribute is missing from the context in the lazy parser logic

in short, zod fails internally when parsing nested schemas starting version 3.14.3

### DIFF
--- a/packages/kotti-ui/package.json
+++ b/packages/kotti-ui/package.json
@@ -19,7 +19,7 @@
 		"tippy.js": "6.x",
 		"ts-custom-error": "^3.1.1",
 		"vue-clickaway": "^2.2.2",
-		"zod": "^3.14.1"
+		"zod": "3.14.1"
 	},
 	"description": "Kotti Vue Component UI",
 	"devDependencies": {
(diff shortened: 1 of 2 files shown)


HansAuger (approved):
@carsoli I'll approve. Did you see if there is an open (or already closed) issue for this in the zod repo?  In case not, you may want to use your  current broken state of kotti to document  the bug and send  them data.

carsoli:
> @carsoli I'll approve. Did you see if there is an open (or already closed) issue for this in the zod repo? In case not, you may want to use your current broken state of kotti to document the bug and send them data.

Yes @HansAuger  I will create an issue today

And no, I personally didn't find this but I will revisit before talking to the author :) 
I also had a plan to see if this was somehow implicitly fixed in the next minor version, and therefore, i will test it as well first

Issue #647, opened 2022-07-15
KtAccordion with dynamic content does not resize


santiagoballadares:
**Describe the bug**
KtAccordion with dynamic content does not resize after animation is triggered even when more content is added. Any additional content is cut off.

**To Reproduce**

1. Setup a KtAccordion with dynamic content, e.g.: add a button 'Add Item' that adds content dynamically inside the KtAccordion
2. Make sure KtAccordion is open by default
3. On first render, add content to KtAccordion, e.g.: by clicking on the button 'Add Item'
4. KtAccordion will grow and all its content is visible
5. Collapse and expand KtAccordion (animation will get triggered)
6. Add more content to KtAccordion
7. KtAccordion does not resize and any additional content is hidden

**Expected behavior**
KtAccordion resizes when more content is added and always shows all its content.

**Screenshots**
<img width="615" alt="Screen Shot 2022-07-15 at 11 42 39" src="https://user-images.githubusercontent.com/744091/179269377-839dfe27-a646-43d4-a576-6d1e4e1dcfc9.png">

**Desktop (please complete the following information):**

- OS: MacOS
- Browser: chrome, safari, firefox
- Version: latest on 2022/07/15

carsoli:
closed by #648

PR #648, opened 2022-07-15
fix(#647): fix KtAccordion with dynamic content not resizing 

santiagoballadares:
Cancel animation to clear keyframeEffects and allow container to resize


FlorianWendelborn:
Code LGTM. Didn't test since I'm on mobile but it seems reasonable that it could fix the issue

FlorianWendelborn (dismissed):
Code LGTM. Didn't test since I'm on mobile but it seems reasonable that it could fix the issue

carsoli (approved):

PR #651, opened 2022-07-25
650 Accordion make close/open fully controllable

### LINKED ISSUE #650: Make KtAccordion's open/close state fully controllable
KtAccordion is initialized with an open/closed state and then changes and maintains that autonomously. It would offer more flexibility if that was fully controlled by the parent

**Describe the solution you'd like**
Next to `isClosed` prop (right now only used to initialize KtAccordion in closed state) emit `onIsClosedChange` instead of changing the state internally, to let parent handle change.

`isClosed` might seem like an odd choice for a prop, but it follows the logic of having good defaults, and if you don't want the parent to control this, you will likely be the Accordion be always open.

HansAuger:
Resolves #650 

Can be tested in documentation.

BREAKING CHANGE: isClosed becomes controllable and must always be maintained by parents

~~We usually don't rely on clients using `.sync` but in this case it reduces some boiler plate, so I thought it was good to mention in the docs.~~

### DIFF
--- a/packages/kotti-ui/source/kotti-accordion/KtAccordion.vue
+++ b/packages/kotti-ui/source/kotti-accordion/KtAccordion.vue
@@ -28,17 +28,19 @@ import { makeProps } from '../make-props'
 import { useSlideAnimation } from './hooks'
 import { KottiAccordion } from './types'
 
+//** How much time should be spent on open/close animation in milliseconds */
+const ANIMATION_DURATION = 300
+
 export default defineComponent<KottiAccordion.PropsInternal>({
 	name: 'KtAccordion',
 	props: makeProps(KottiAccordion.propsSchema),
 	setup(props, { emit }) {
 		const contentInnerRef = ref<HTMLElement | null>(null)
-		const duration = 300
 
 		useSlideAnimation(
 			contentInnerRef,
 			computed(() => !props.isClosed),
-			{ duration },
+			{ duration: ANIMATION_DURATION },
 		)
 		return {
 			contentClasses: computed(() => ({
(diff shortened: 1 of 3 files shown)


carsoli:
For your own info @HansAuger https://zelig880.com/vuejs-3-say-goodbye-to-sync-and-welcome-the-brand-new-v-model

`.sync` was generally introduced in vue2 after the core team found out that people want a similar syntactic sugar to `v-model` that works with other props 

because `v-model` was only mapped to `value/@input` pair OR `value/@update:value`

so it was the solution to extending the v-model 

in Vue3, this is deleted and the v-model is powerful
so you can have `x/@update:x` prop/event pairs which can would be the same as the `v-model:x` (syntactic sugar that's easier to remember)

Mentioning all of this, bc I'd avoid using `sync` and instead respect `x/@update:x` in new components/in refactors so that our job is easier when we migrate

HansAuger:
@carsoli good, was never a fan of the `.sync` pattern anyway. I will change it in the documentation then

packages/kotti-ui/source/kotti-accordion/KtAccordion.vue
```diff
@@ -31,23 +31,25 @@ import { KottiAccordion } from './types'
 export default defineComponent<KottiAccordion.PropsInternal>({
 	name: 'KtAccordion',
 	props: makeProps(KottiAccordion.propsSchema),
-	setup(props) {
+	setup(props, { emit }) {
 		const contentInnerRef = ref<HTMLElement | null>(null)
+		const duration = 300
```
carsoli, line 36: since it's a constant, it can move outside the setup and folllow the convention of casing `ANIMATION_DURATION` or something

carsoli (dismissed):

santiagoballadares (approved):

PR #652, opened 2022-07-26
Update to vue 2.7.8

Isokaeder:
[Vue "Naruto"](https://blog.vuejs.org/posts/vue-2-7-naruto.html) is released, making composition api for the greatest part available inside Vue 2.7

This is a naive update to Vue 2.7 and removal of @vue/composition-api, without incremental progression to catch all breaking changes. This is meant as a research and experimentation branch, to answer questions about the best approach to update to vue-v3, and whether we should go there directly or whether vue 2.7 should be an intermediary step.


FlorianWendelborn:
Wow that’s some great news

Issue #657, opened 2022-08-24
Radios and Toggles can look squeezed


HansAuger:
**Describe the bug**
Radio Buttons in `KtRadioGroupField` can look squeezed

<img width="412" alt="Screenshot 2022-08-24 at 12 47 20" src="https://user-images.githubusercontent.com/16950410/186400714-1f194b82-7aaa-4999-96c7-16640a7b037a.png">

**To Reproduce**
Can be seen in the doc page for `KtRadioGroupField` if a big string is entered in the header field

**Additional context**
Also affects `KtToggleGroupField` and `KtToggleField`

Proposed fix: deactivate flex-shrink for radio button

PR #659, opened 2022-08-29
fix(657): tell flex not to shrink its toggle

HansAuger:
See Ticket #657 for reference and how to (not any more) reproduce


FlorianWendelborn (dismissed):
LGTM

FlorianWendelborn (approved):

Issue #273, opened 2020-09-07
Deprecate Old Kotti Form Fields


FlorianWendelborn:
## Removal State

- [ ] `KtCheckbox`
  - [ ] Deprecation Warning
  - [ ] Remove from 3D Viewer
  - [ ] Remove form Yoda
- [ ] `KtDateInput`
  - [ ] Deprecation Warning
  - [ ] Remove from 3D Viewer
  - [ ] Remove form Yoda
- [ ] `KtInput`
  - [ ] Deprecation Warning
  - [ ] Remove from 3D Viewer
  - [ ] Remove form Yoda
- [ ] `KtInputNumber`
  - [ ] Deprecation Warning
  - [ ] Remove from 3D Viewer
  - [ ] Remove form Yoda (https://gitlab.3yourmind.com/dev/yoda/-/merge_requests/397)
- [ ] `KtRadioGroup`
  - [ ] Deprecation Warning
  - [ ] Remove from 3D Viewer
  - [ ] Remove form Yoda
- [ ] `KtRadio`
  - [ ] Deprecation Warning
  - [ ] Remove from 3D Viewer
  - [ ] Remove form Yoda
- [ ] `KtSelect`
  - [ ] Deprecation Warning
  - [ ] Remove from 3D Viewer
  - [ ] Remove form Yoda
- [ ] `KtSwitch`
  - [ ] Deprecation Warning
  - [ ] Remove from 3D Viewer
  - [ ] Remove form Yoda

PR #661, opened 2022-09-05
chore(KtRadio/KtRadioGroup): remove them

carsoli:
no longer used, deprecated since a while


packages/kotti-ui/source/kotti-style/forms/_misc.scss
```diff
@@ -73,10 +73,7 @@
 }
 
 // Disabled fields
-.form-input,
-.form-checkbox,
-.form-switch,
-.form-radio {
```
FlorianWendelborn, line 79: This class is still used in `src/user-panel-frontend/src/components/basket/wizard/BasketWizardPostProcessingSingleSelect.vue`

FlorianWendelborn (dismissed):

packages/documentation/data/menu.ts
```diff
@@ -19,7 +19,6 @@ import {
 	KtNavbar,
 	KtPagination,
 	KtPopover,
-	KtRadio,
```
FlorianWendelborn, line 22: Please mark the 4 relevant tasks as complete in #273. After this PR, we only have `KtInput` to remove, then we can finally release Kotti 3 officially 🚀

FlorianWendelborn: Also, `KtInput` only appears 5 times in the codebase by now. So, the more challenging issue will be the associated classes.

FlorianWendelborn: Oh, I just noticed we forgot to add the inline edit to that document. But still, some good progress towards V3.

carsoli: the inline edit is also wip by me now, would be great to take a look when I submit the final changes today so I can also merge the new one :)

HansAuger (approved):

PR #665, opened 2022-09-07
fix(KtModal): does not show if isOpen prop is set to true initially

carsoli:
unless it changes and retriggers the watcher, the isOpen value
was dismissed in the first iteration
which led to a bug in announcementModal usages, where `isOpen` was true
and never changed throughout the lifecycle of the component.

Fix was to make the watcher immediate, keep track of the initial value
of `isOpen` and wait for tippy to be created and then show it
if the initial value was true.

### DIFF
--- a/packages/kotti-ui/source/kotti-modal/KtModal.vue
+++ b/packages/kotti-ui/source/kotti-modal/KtModal.vue
@@ -35,6 +35,18 @@ export default defineComponent<KottiModal.PropsInternal>({
 		const contentRef = ref<HTMLElement | null>(null)
 		const targetRef = ref<HTMLElement | null>(null)
 		const tippyInstanceRef = ref<Instance | null>(null)
+		/**
+		 * The `onCreate` function of `tippyjs` gets executed after the `KtModal`
+		 * component renders.
+		 *
+		 * This boolean that keeps track of the initial value of `props.isOpen`,
+		 * and passes it once tippy is created.
+		 *
+		 * It can be passed as a prop to tippy.
+		 * @see {@link https://atomiks.github.io/tippyjs/v6/all-props/#showoncreate}
+		 * However, for readability purposes, we manually show inside the `onCreate` function.
+		 */
+		const showOnCreate = ref(false)
 
 		useTippy(
 			targetRef,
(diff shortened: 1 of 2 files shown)


FlorianWendelborn:
@carsoli this PR seems to accidentally contain the `KtFieldInlineEdit` changes

packages/kotti-ui/source/kotti-modal/KtModal.vue
```diff
@@ -73,13 +89,20 @@ export default defineComponent<KottiModal.PropsInternal>({
 
 		watch(
 			() => props.isOpen,
-			(value) => {
+			(shouldOpen, wasOpen) => {
+				if (wasOpen === undefined) {
+					showOnCreate.value = shouldOpen
+					// theoratically, always true
```
FlorianWendelborn, line 95:
```suggestion
					// theoretically, always true
```

packages/kotti-ui/source/kotti-modal/KtModal.vue
```diff
@@ -35,6 +35,18 @@ export default defineComponent<KottiModal.PropsInternal>({
...
+		 * component renders.
+		 *
+		 * This boolean that keeps track of the initial value of `props.isOpen`,
+		 * and passes it once tippy is created.
+		 *
+		 * It can be passed as a prop to tippy.
+		 * @see {@link https://atomiks.github.io/tippyjs/v6/all-props/#showoncreate}
+		 * However, for readability purposes, we manually show inside the `onCreate` function.
+		 */
+		const showOnCreate = ref(false)
```
FlorianWendelborn, line 49: Can't you simply check the current value of `isOpen` and get rid of this variable?

FlorianWendelborn (approved):

packages/kotti-ui/source/kotti-modal/KtModal.vue
carsoli: ye sure whatever

carsoli: on second thought, i prefer this because it's more readabile and doesn't leave room for future mistakes

santiagoballadares (dismissed):

RueRivoli (approved):

PR #668, opened 2022-10-06
feat(KtFieldMultiSelect): support clearOnSelect

Juli0GT:
It is a desired feature by the user-app to clear the user query of the `KtFieldMultiSelect` after option selection.

Therefore, this PR introduces an optional prop `clearOnSelect`, which when set to true, clears the user query after every option selection.

### DIFF
--- a/packages/documentation/pages/usage/components/form-fields.vue
+++ b/packages/documentation/pages/usage/components/form-fields.vue
@@ -454,7 +454,7 @@
 								"
 								formKey="clearOnSelect"
 								isOptional
-								label="clear on select"
+								label="clearOnSelect"
 								type="switch"
 							/>
 							<h4>Additional Slots</h4>
--- a/packages/documentation/pages/usage/components/form-fields.vue
+++ b/packages/documentation/pages/usage/components/form-fields.vue
@@ -448,6 +448,15 @@
 								label="option slot"
 								type="switch"
 							/>
+							<KtFieldToggle
+								v-if="
+									componentDefinition.additionalProps.includes('clearOnSelect')
+								"
+								formKey="clearOnSelect"
+								isOptional
+								label="clear on select"
+								type="switch"
+							/>
 							<h4>Additional Slots</h4>
 							<KtFieldText
 								v-if="
@@ -614,6 +623,7 @@ const components: Array<{
 	{
 		additionalProps: [
 			'actions',
+			'clearOnSelect',
 			'collapseTagsAfter',
 			'hasOptionSlot',
 			'isUnsorted',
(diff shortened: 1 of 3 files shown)


packages/kotti-ui/source/kotti-field-radio-group/KtFieldRadioGroup.vue
```diff
@@ -81,7 +81,6 @@ export default defineComponent<KottiFieldRadioGroup.PropsInternal>({
 
 		const { forceUpdate, forceUpdateKey } = useForceUpdate()
 		const name = ref(`autoGeneratedRadioName${nameIndex++}`)
-
```
carsoli, line 84: looks accidental, readd the line please

carsoli (approved):

carsoli (approved):

carsoli (approved):

santiagoballadares (approved):

FlorianWendelborn:
@Juli0GT I’ll squash the commits that introduced `package-lock.json`. Just deleting it doesn’t get rid of that large file from the git history and slows down git over time

packages/documentation/pages/usage/components/form-fields.vue
```diff
@@ -448,6 +448,15 @@
 								label="option slot"
 								type="switch"
 							/>
+							<KtFieldToggle
+								v-if="
+									componentDefinition.additionalProps.includes('clearOnSelect')
+								"
+								formKey="clearOnSelect"
+								isOptional
+								label="clear on select"
```
FlorianWendelborn, line 457: Should be
```suggestion
								label="clearOnSelect"
```
as this serves as the documentation for how this prop is called as well

packages/kotti-ui/source/kotti-field-select/components/GenericSelectField.vue
```diff
@@ -237,11 +240,14 @@ export default defineComponent({
 			})),
 			inputRef,
 			isDropdownOpen,
+			localQuery,
```
FlorianWendelborn, line 243: This shouldn’t be necessary to expose to the template as you didn’t add any new template code in this PR
```suggestion
```

FlorianWendelborn (dismissed):
LGTM, just 2 minor things you can directly commit through the GitHub PR UI if you agree.

carsoli:
@FlorianWendelborn please approve because now that I've committed the suggestions, we need another reviewer

also please check slack, there's a UI bug going on but it's also on master, unless you see a reason why it's specifically happening for this PR more often.

Afterwards, please go ahead and merge

carsoli (approved):

FlorianWendelborn:
@carsoli I think we should disable automatic review dismissal

FlorianWendelborn (approved):

PR #670, opened 2022-10-12
feat(KtField*Select*): default hoveredIndex to `0` instead of `-1`

Juli0GT:
It is a desired feature by the user-app that:

- When dropdown menu opens, the first item is automatically on focus → user can press Enter key and the item is selected
- User can start typing and the items are refined based on the search → first item of the results remains automatically on focus → user can press Enter key and the item is selected

Therefore, this PR changes the initial `hoveredIndex` from `-1` to `0`

### DIFF
--- a/packages/kotti-ui/source/kotti-field-select/components/Options.vue
+++ b/packages/kotti-ui/source/kotti-field-select/components/Options.vue
@@ -115,8 +115,8 @@ export default defineComponent({
 			return modifiedOptions.sort((a, b) => a.label.localeCompare(b.label))
 		})
 
-		const hoveredIndex = ref(-1)
-		const resetHoveredIndex = () => (hoveredIndex.value = -1)
+		const hoveredIndex = ref(0)
+		const resetHoveredIndex = () => (hoveredIndex.value = 0)
 
 		watch(
 			() => props.isDropdownOpen,


carsoli (dismissed):
about to test though

carsoli:
@FlorianWendelborn https://github.com/3YOURMIND/kotti/pull/670/files#diff-b323717e8f9826727fb935479c5fe49b55ad28979615cff3b1cd61a89b2e73f4R160

this line and the one for handling keyup, if called when the itemCount is `0` would result in `NaN`

Is that okay? I skimmed quickly through the code so maybe you somehow guard against the itemCount being `0` and it's not a bug, but if you don't

that would make the hoveredIndex.value at one point a `NaN` - that okay?

santiagoballadares (dismissed):

FlorianWendelborn:
@carsoli I don’t know what line you’re referring to as the link doesn’t include a line number

carsoli:
@FlorianWendelborn Idk why the link generated is like this, it's R160 as indicated by the last part of the link but it's no longer prefixed with a `#` as the link genration worked before for lines 🤷🏼

carsoli:
@FlorianWendelborn I'm not concerned about the PR, I'm asking in general

the item count can be `0` , right? so in that case the divisor would be 0 and `NaN` is possibly generated

carsoli (approved):

Issue #674, opened 2022-11-14
KtTable have glitchy checkboxes when KtTable has isSelectable to true


RueRivoli:
**Describe the bug**
The checkboxes of selectable tables are glitchy

**To Reproduce**
Go to any KtTable that has the property `isSelectable` prop set to true
You see the screenshot below

**Expected behavior**
Checkboxes are not glitchy

**Screenshots**
![image (2)](https://user-images.githubusercontent.com/17473800/201733289-f1eae5cd-2302-43fb-8547-48709c0522b9.png)

carsoli:
Closed by #684

PR #675, opened 2022-11-14
fix(#674): re-add removed css properties for .form-checkbox .form-icon for KtTable

RueRivoli:
- write back css for .form-checkbox .form-icon necessary in KtTable with selectable rows to avoid glitchy checkboxes in KtTable with selectable rows

### DIFF
--- a/packages/kotti-ui/source/kotti-style/forms/_toggles.scss
+++ b/packages/kotti-ui/source/kotti-style/forms/_toggles.scss
@@ -1,0 +1,57 @@
+.kt-table .form-checkbox {
+	position: relative;
+	display: inline-block;
+	min-height: 1.2rem;
+	padding: (($control-size-sm - $line-height) / 2) $control-padding-x
+		(($control-size-sm - $line-height) / 2)
+		($control-icon-size + $control-padding-x);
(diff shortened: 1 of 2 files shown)


packages/kotti-ui/source/kotti-style/forms/_toggles.scss
```diff
@@ -0,0 +1,6 @@
+.form-checkbox .form-icon {
```
FlorianWendelborn, line 1: Shouldn't this be scoped to only apply to the table?

RueRivoli: hum it can. It's currently necessary only for KtTable if I'm not wrong.
Here I just put it back how it was originally

Juli0GT:
I approve, just @carsoli 's minor change regarding: `kt-table`

Juli0GT (approved):

HansAuger (approved):

FlorianWendelborn:
@carsoli using rebase merge doesn't create a merge commit. I'd strongly recommend using autorebase and a normal merge instead so the commit history is easier to understand and shows which PRs were merged

PR #680, opened 2022-11-25
chore(ci): node-version to latest LTS (18.12.x)

carsoli:
render deployments are failing with this Error

```
Error: error:0308010c:digital envelope routines::unsupported
```

testing to see if this will solve it


santiagoballadares (approved):

README.md
```diff
@@ -147,6 +147,16 @@ yarn run check:stylelint
 yarn run test
```
FlorianWendelborn, line 147: blocking

FlorianWendelborn (changes requested):

README.md
carsoli: ?

.github/workflows/gh-pages.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: '16'
```
FlorianWendelborn, line 13: This change breaks the build for GitHub Pages. https://github.com/3YOURMIND/kotti/actions/runs/3550513262/jobs/5963966183

I also see no evidence that this is a render.com problem. It seems to be happening in node@18 in general, at least on CI, for whatever reason.

On my local macOS Ventura system, I can’t reproduce this issue, but that might be because of openssl versions since this issue is related to those as far as I understand

.github/workflows/ci.yml
```diff
@@ -80,7 +80,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: '16'
```
FlorianWendelborn, line 83: I don’t think it’s a good idea to pin to a non-major version since those aren’t guaranteed to be LTS.

README.md
```diff
@@ -147,6 +147,16 @@ yarn run check:stylelint
...
 ```
 
+### Documentation Deployment & CI
+
+CI jobs config scripts can be found under .github/workflows
+
+Docs from `master` branch are hosted on [gh-pages](https://pages.github.com/), and via [Render](https://render.com/) on each open PR.
```
FlorianWendelborn, line 154: I find the wording confusing as it sounds like GitHub Pages hosts PR environments

```suggestion
Every PR is hosted on [Render](https://render.com) with a link on the pull request conversation. Docs from `master` are additionally hosted on [GitHub Pages](https://pages.github.com/).
```

.nvmrc
```diff
@@ -1 +1 @@
-lts/*
```
FlorianWendelborn, line 1: Why is this pinned to this specific version and not 16 so we at least get some maintenance LTS updates?

FlorianWendelborn (changes requested):

.github/workflows/gh-pages.yml
carsoli: Please read the description

FlorianWendelborn: @carsoli I read the description like 5 times already today.

carsoli: if it also breaks github pages to use node 16.16 then basically only node 16.16 (LTS) and node 18.12 (LTS) got the openssl package upgrades and it would explain why gh-pages breaks starting 16.16 and render breaks for 18.12 (latest LTS specified previously on `.nvmrc`)


My suggestion would be to basically only change the `.npmrc` to 16 then so that render works and keep everything as is because at this point getting the openssl package upgrades breaks both gh-pages and render

carsoli: if you look at the failure logs on render's dashboard u will see that a sub-dependency of render is the one that errors with node version 18

carsoli: you'd have to manually trigger on your branch, because the latest manual trigger I did was on the last commit on my branch (this one)

FlorianWendelborn: > if you look at the failure logs on render's dashboard u will see that a sub-dependency of render is the one that errors with node version 18

I don’t understand why you come to the conclusion that it’s a dependency of render.

.github/workflows/ci.yml
carsoli: 18 then

.github/workflows/gh-pages.yml
FlorianWendelborn: My current understanding is that `webpack@4` is incompatible with `openssl@3`.

One of the major changes of `node@17` was upgrading from `openssl@1.1` to `openssl@3` <https://nodejs.org/es/blog/release/v17.0.0/#openssl-3-0>

`node@16` is no longer LTS as of 2022-10-18 which fits approximately with when Render first showed build failures.

So, `nvm`’s `lts` got updated to `18` around that time, which caused render to install `18` with `openssl@3` and thus failed the webpack build.

.nvmrc
carsoli: becuase i wanted the open ssl changes because it works with render anyways

carsoli: https://nodejs.org/en/blog/release/v16.16.0/

.github/workflows/gh-pages.yml
carsoli: did you check the logs

FlorianWendelborn: Here’s a webpack statement and potential workarounds for this bug: <https://github.com/webpack/webpack/issues/14532#issuecomment-947525539>

FlorianWendelborn: > the same logs you're looking at

I used the link from the PR description

carsoli: it is webpack, I just went back to render's dashboard, you're right.

carsoli: I'll attempt this output.hashFunction: "xxhash64" on this PR then, and keep the upgrades to 18 everywhere, and to lts on `.nvmrc` 

let's see

FlorianWendelborn: @carsoli yeah, if we can get that workaround to work I think that’s the best solution. I’d suggest restoring `lts` wherever possible and pinning the rest to `18` so we actually get security updates

package.json
```diff
@@ -128,7 +128,7 @@
 	"private": true,
 	"repository": "git@github.com:3YOURMIND/kotti",
 	"scripts": {
-		"build:render.com": "yarn run watch:build-kotti-dependencies && yarn workspace @3yourmind/kotti-ui run build && yarn workspace @3yourmind/documentation run nuxt generate",
```
FlorianWendelborn, line 131: @carsoli Can you explain why setting the webpack hash function wasn’t sufficient for this?

render.yaml
```diff
@@ -1,7 +1,7 @@
 services:
   - type: web
     name: kotti
-    env: static
```
FlorianWendelborn, line 4: I think this is a full-text-replace mistake. Seems nonsensical to me

package.json
carsoli: couldn't

carsoli:
This is becoming urgent for render envs @FlorianWendelborn 
I don't want to invest more time in this - we are focusing on migrating our user app this month anyways - and will have core library updates

the same will follow for Kotti. So, all good.

.github/workflows/ci.yml
```diff
@@ -12,7 +12,9 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
```
santiagoballadares, line 13:
```suggestion
      - name: Setup NODE
         uses: actions/setup-node@v2
```

README.md
```diff
@@ -147,6 +147,16 @@ yarn run check:stylelint
 yarn run test
 ```
 
+### Documentation Deployment & CI
+
+CI jobs config scripts can be found under .github/workflows
+
+Every PR is hosted on [Render](https://render.com) with a link on the pull request conversation. Docs from `master` are additionally hosted on [GitHub Pages](https://pages.github.com/).
+
+node-version: Currently we specifically target LTS v18.12.x as the used node-version
```
santiagoballadares, line 156:
```suggestion
node-version: Currently we specifically target LTS v18 as the used node-version
```

carsoli:
closing, not needed thanks to supporting node18 on #814

PR #683, opened 2022-11-29
fix(#882): hovering over an breadcrumb arrow should not change the text's color

RueRivoli:


santiagoballadares (approved):

packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
```diff
@@ -96,13 +96,15 @@ export default defineComponent<KottiBreadcrumb.PropsInternal>({
 		&--is-disabled {
 			color: var(--text-05);
 			cursor: not-allowed;
+			.kt-breadcrumb__list-item__text:hover {
+				cursor: not-allowed;
+				color: var(--text-05);
```
santiagoballadares, line 101:
```suggestion
				color: var(--text-05);
				cursor: not-allowed;
```

packages/kotti-ui/source/kotti-breadcrumb/KtBreadcrumb.vue
```diff
@@ -96,13 +96,15 @@ export default defineComponent<KottiBreadcrumb.PropsInternal>({
...
+				cursor: not-allowed;
+			}
 		}
 
-		&:hover:not(.kt-breadcrumb__list-item--is-disabled) {
-			.kt-breadcrumb__list-item__text {
-				color: var(--interactive-01-hover);
-				cursor: pointer;
-			}
+		&__text:hover {
```
carsoli, line 105: it's just odd that the old implementation is not working..

because we only apply the color change and cursor change IF the parent is not disabled ( parent being `.kt-breadcrumb__list-item` )

PR #685, opened 2022-11-30
feat(KtFilters/#679): Support data-test on each kt-filter__row

Juli0GT:
Support data-test on each kt-filter__row to be split between column, operator, value.

For that, every single component that is rendered from the KtFilters need to accept a `dataTest` prop. The value of this prop is used to overwrite the `data-test` property from the `inputProps`. So the `input` element receives this `data-test`.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8480,9 +8480,9 @@ decimal.js@^10.2.1:
   integrity sha512-V0pfhfr8suzyPGOx3nmq4aHqabehUZn6Ch9kyFpV79TGDTWFmHqUqXdabR7QHqxzrYolF4+tVmJhUG4OURg5dQ==
 
 decode-uri-component@^0.2.0:
-  version "0.2.0"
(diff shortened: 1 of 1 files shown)


packages/kotti-ui/source/kotti-field-currency/types.ts
```diff
@@ -15,6 +15,7 @@ export namespace KottiFieldCurrency {
 		)
 		.extend({
 			currency: z.string(),
+			dataTest: z.string().optional(),
```
FlorianWendelborn, line 18: This should be implemented globally, not per field.

FlorianWendelborn (changes requested):

packages/kotti-ui/source/kotti-field-currency/types.ts
Juli0GT: You are right, now is added into `potentiallySupportedPropsSchema`.

PR #686, opened 2022-12-02
KtField: support dataTest on input level of all KtFields

Juli0GT:
`dataTest` prop now is supported on all KtFields.
`useInputProps` initialise `data-test` property with the value of the `dataTest` prop if available.


carsoli:
Great @Juli0GT :)

carsoli (approved):

PR #689, opened 2022-12-06
refactor(KtFilters): Fix ts errors realted to KtFilters

Juli0GT:
Fix ts errors related to `KtFilters`


packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
```diff
@@ -127,14 +123,19 @@ export default defineComponent<{
 			() => operationOptions.value.length <= 1,
 		)
 		const isValueFieldVisible = computed(
-			() => !isEmptyOperation(props.filter.operation, props.column?.type),
+			() => 	props.column &&
+				props.filter &&
+				!isEmptyOperation(props.filter.operation, props.column?.type),
```
santiagoballadares, line 128:
```suggestion
			() =>
				props.column &&
				props.filter &&
				!isEmptyOperation(props.filter.operation, props.column?.type),
```

packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
```diff
@@ -104,11 +104,7 @@ export default defineComponent<{
 			type: String,
 		},
 		filter: {
-			default: (): KottiFilters.Filter => ({
```
carsoli, line 107: I think this change has an effect on all mentions of `props.filter.*` because it used to be null as default, for all 3 attributes (key, operation, value) but now it's not so fallbacks have to be put in place 

Can you run a check through the whole file and add it 

Alternatively, I can leave suggestions

packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
```diff
@@ -127,14 +123,20 @@ export default defineComponent<{
 			() => operationOptions.value.length <= 1,
 		)
 		const isValueFieldVisible = computed(
-			() => !isEmptyOperation(props.filter.operation, props.column?.type),
+			() =>
+				props.column &&
+				props.filter &&
+				!isEmptyOperation(props.filter.operation, props.column?.type),
```
carsoli, line 129: you added `props.filter &&` but it can just be `!isEmptyOperation(props.filter?.operation` --

stylistic difference but we usually prefer the `optional chaning` over explicit conditions

unless `isEmptyOperations` can accept a `null` - in which case it's correct as is

carsoli: actually on second thought

`isEmptyOperation` technically must be accepting `null` as a first argument because it used to before (props.filter.operation) had a default `null` value

so you can get away with `props.filter?.operation ?? null` as the first argument and remove the explicit check


```
() => !isEmptyOperation(props.filter?.operation ?? null, props.column?.type)
```

packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
carsoli, line 107: https://github.com/3YOURMIND/kotti/pull/689/files#diff-7447735e613d3621591bc6fc61f7aaa43f0feff8480db9da599fdb51af316461R8

you need a v-if on the `KtForm`

because now if the `filter` becomes a `null` prop -- you won't have the formKeys and the form will error

packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
Juli0GT: @carsoli I believe this changes were introduced by @santiagoballadares
`isEmptyOperation` does not accept null for any of the arguments.

I will leave it as it was with the `optional.chaining` syntax.

Juli0GT: Actually `ts` complains in the second argument with the possibility of being undefined. 
Thats why we need the explicit check. So I will leave both explicit checks and remove the `optional.chaining`

carsoli (approved):

packages/kotti-ui/source/kotti-filters/components/FilterRow.vue
```diff
@@ -1,5 +1,6 @@
 <template>
 	<KtForm
+		v-if="filter"
```
santiagoballadares, line 3: This is going to break adding a new filter. When you click add Filter, we render the same component but the prop filter is null because we haven't yet chosen a column.

carsoli: then the v-if would have to be on the level of the fields

If the `filter` is null and the fields use formKeys, they will break as well --

@Juli0GT can u move the v-if to be around the content of the form

you can introduce a layer

like this 
```
<KtForm ....
  <template v-if="filter">
    <KtField* ...>
  </template>
```

Juli0GT:
@carsoli @santiagoballadares @FlorianWendelborn can we merge this?

carsoli (approved):

PR #693, opened 2022-12-29
fix(docs): make folder name of components configurable

Isokaeder:
I noticed that not the links for the `KtField*Select*`s did not work, since the naming scheme of their file paths is not according to our convention. 

To fix this, I added an optional `componentFolder` field into the meta object. If it is set, this will be used instead of deriving the name from the component's name.

To test, look into the render, and check that for the select fields the links lead to the actual documentation pages.
Also, I did not check yet if there are any similar cases, will do it later. If something comes to mind, @ me

### DIFF
--- a/packages/documentation/components/ComponentInfo.vue
+++ b/packages/documentation/components/ComponentInfo.vue
@@ -148,6 +148,7 @@ export default defineComponent<{
 		props?: Record<string, unknown>
 	}
 }>({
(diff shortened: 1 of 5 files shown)


carsoli (approved):

carsoli:
@Isokaeder dateTime, dateRange, dateTimeRange as well

PR #700, opened 2023-01-13
feat(KtFilters/KtFieldToggle): Initial default value is true instead of null

Juli0GT:
Feat(KtFilters/KtFieldToggle): Initial default value is true, not possible to have boolean filter in null state.

When a user opens the filter popover and selects a boolean property filter: 

- By default the property is selected as true (checkbox is selected) → no interaction needed for the user
- When user clicks on the checkbox, state is changed to false (checkbox is unselected)
- It’s not possible to have the boolean filter in a null state

Closes [B3-12375](https://3yourmind.atlassian.net/browse/B3-12375)

[B3-12375]: https://3yourmind.atlassian.net/browse/B3-12375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kotti-ui/source/kotti-filters/utils.ts
+++ b/packages/kotti-ui/source/kotti-filters/utils.ts
@@ -47,7 +47,7 @@ export const getFilterInitialState = (
 					column.operations,
 					KottiFilters.Operation.Boolean.EQUAL,
 				),
-				value: null,
+				value: true,
 			}
 		case KottiFilters.FilterType.CURRENCY:
 			return {


Juli0GT:
@carsoli @santiagoballadares I'm not sure if thats all it needs to be done, I have functionally tested it and seems to work correctly. But there is a function called `getFilterEmptyValue` that still returns `null` for `KottiFilters.FilterType.BOOLEAN` not sure if something needs to be changed there.

santiagoballadares (approved):

santiagoballadares:
> @carsoli @santiagoballadares I'm not sure if thats all it needs to be done, I have functionally tested it and seems to work correctly. But there is a function called `getFilterEmptyValue` that still returns `null` for `KottiFilters.FilterType.BOOLEAN` not sure if something needs to be changed there.

@Juli0GT No need to touch `getFilterEmptyValue`, this is used for `IS_EMPTY` operation which is intended to filter records with `null` values. Also, this is currently not supported in Yoda, hence not used.

Juli0GT:
> > @carsoli @santiagoballadares I'm not sure if thats all it needs to be done, I have functionally tested it and seems to work correctly. But there is a function called `getFilterEmptyValue` that still returns `null` for `KottiFilters.FilterType.BOOLEAN` not sure if something needs to be changed there.
> 
> @Juli0GT No need to touch `getFilterEmptyValue`, this is used for `IS_EMPTY` operation which is intended to filter records with `null` values. Also, this is currently not supported in Yoda, hence not used.

Sweet amazing @santiagoballadares !

Isokaeder (approved):

PR #701, opened 2023-01-17
feat(699): show label suffix next to default slot when label is not g…

santiagoballadares:
Need to show required `*`, optional `(Optional)`, and help text `(?)` next to the content of the `Default slot` that is displayed on the right side of the checkbox/toggle element in KtFieldToggle when using `Default slot` instead of `label` prop.

image-20230103-160436

### DIFF
--- a/packages/kotti-ui/source/kotti-field-toggle/KtFieldToggle.vue
+++ b/packages/kotti-ui/source/kotti-field-toggle/KtFieldToggle.vue
@@ -1,20 +1,29 @@
 <template>
 	<KtField
-		v-bind="{ field }"
+		v-bind="{ field: fieldProps }"
 		:getEmptyValue="() => null"
-		:helpTextSlot="$slots.helpText"
+		:helpTextSlot="showInnerSuffix ? undefined : $slots.helpText"
 	>
 		<div slot="container" class="kt-field-toggle__wrapper">
 			<ToggleInner
 				component="div"
 				:inputProps="inputProps"
-				:isDisabled="field.isDisabled"
+				:isDisabled="fieldProps.isDisabled"
 				:type="type"
-				:value="field.currentValue"
+				:value="fieldProps.currentValue"
 				@input="onInput"
 			>
 				<div class="kt-field-toggle__wrapper__content">
-					<slot name="default" :value="field.currentValue" />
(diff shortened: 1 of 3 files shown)


Juli0GT:
@santiagoballadares MR is missing description

packages/kotti-ui/source/kotti-field-toggle/KtFieldToggle.vue
```diff
@@ -37,20 +47,27 @@ export default defineComponent<KottiFieldToggle.PropsInternal>({
...
 		})
+		const showInnerSuffix = computed(
+			() => Boolean(slots.default) && props.label === null,
+		)
 
 		const { forceUpdate, forceUpdateKey } = useForceUpdate()
 
 		return {
-			field,
+			field: computed(() => ({
```
Juli0GT, line 67: Would give it a different name to the computed property as there is already `const field` in the file

Juli0GT:
Other than the minor comment I left, LGTM

packages/kotti-ui/source/kotti-field-toggle/components/ToggleInnerSuffix.vue
```diff
@@ -0,0 +1,68 @@
...
+		helpTextSlot: { default: () => [], type: Array },
+		hideValidation: { default: false, type: Boolean },
+		isEmpty: { default: false, type: Boolean },
+		isOptional: { default: false, type: Boolean },
+		validation: { default: () => ({}), type: Object },
+	},
+	setup(props: KottiFieldToggle.InnerSuffix) {
+		const validationType = computed(() => props.validation.type)
+		const showValidation = computed(
+			() => !(props.hideValidation || validationType.value === 'empty'),
```
Juli0GT, line 37: If `validation.value === 'empty'` we hide validation?

santiagoballadares: This is correct, validation has `empty` value instead of `null`. See https://github.com/3YOURMIND/kotti/commit/692f636264e0f18f9049b3338c5b9127558d8c61

carsoli:
approved and locally tested

PR #703, opened 2023-01-23
feat(KtPopover): extend options API to allow for checkboxes

carsoli:
replaced `isSelected` attribute on the options object with `isActive`

instead, `isSelected` represents that the checkbox is set to true and only makes sense if `areOptionsSelectable` is passed to `KtPopover`
> introduced `areOptionsSelectable` prop which enables showing the toggle fields ..and making the options selectable.

### DIFF
--- a/packages/kotti-ui/source/kotti-popover/KtPopover.vue
+++ b/packages/kotti-ui/source/kotti-popover/KtPopover.vue
@@ -10,12 +10,25 @@
 					:key="index"
 					:dataTest="option.dataTest"
 					:icon="option.icon"
+					:isActive="option.isActive"
 					:isClickable="option.onClick !== undefined"
 					:isDisabled="option.isDisabled"
-					:isSelected="option.isSelected"
 					:label="option.label"
 					@click.stop="handleItemClick(option)"
-				/>
+				>
+					<template v-if="areOptionsSelectable" slot="option" :option="option">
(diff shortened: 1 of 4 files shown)


packages/kotti-ui/source/kotti-popover/KtPopover.vue
```diff
@@ -125,12 +143,24 @@ export default defineComponent<KottiPopover.PropsInternal>({
 		return {
 			close,
 			contentRef,
+			formKey: computed(() => (formContext === null ? null : 'NONE')),
```
carsoli, line 146: have you tested this in action?

Isokaeder: yes
it works

Isokaeder: We could actually make this a util function that we export as well. It could be useful also for users of kotti.

carsoli: it's kind of an anti-recommendation though :) 

We don't expect users to hack around the form usage. Either a field is used within a form or it's not.. 

You're probably already familiar with the form-within-form hack we have on our user-app and how it's not a good thought in the first place

Isokaeder (approved):

Issue #713, opened 2023-02-10
Update documentation for KtToaster


RueRivoli:
**Is your feature request related to a problem? Please describe.**
warn design the durations for our typical durations for toasters 

**Describe the solution you'd like**
add 2 lines of documentation

**Describe alternatives you've considered**
/

**Additional context**
/

carsoli:
Not to be fixed on kotti, duration is defined on user app

PR #714, opened 2023-02-10
chore(#713): update kt toaster doc

RueRivoli:
Only add a few lines to warn design about the global usage of KtToasters duration in our platform.
Not quite sure it's relevant to put it here though


carsoli:
@RueRivoli should we instead just provide a prop for the time anad have defaults?

carsoli:
Just realised this is set on Yoda's notification store
why would we even put it here @RueRivoli

carsoli:
> Just realised this is set on Yoda's notification store why would we even put it here @RueRivoli

I think we should close both the PR and issue, but please elaborate if you see otherwise

PR #715, opened 2023-02-13
delete(KtTheme): remove deprecated components and quick fix docs

carsoli:
the navbar tokens are all scoped within the Navbar component and can't be overwriten from root level

In any case that's needed in the new design guideline, because the navbar should follow either the brand color or the theme.
> to be made clearer in follow-up PR for theming


FlorianWendelborn (approved):

FlorianWendelborn:
The pipelines are failing due to `color` being an unused dependency now. This was introduced because 10798985532f0844c86388aa463c6f3352954c48 seems to have been directly committed to `master` and thus didn’t go through pipelines

santiagoballadares (approved):

PR #719, opened 2023-02-14
feat(709): add KtFieldFileUpload

santiagoballadares:
Create a new Kotti Field component to upload files.

New component allows selecting files via browser's file selector and drag and drop.

Component properties to control:
    Single or multiple files selection
    File types allowed
    Max file size allowed
    Url for external documentation
    Icon
    Take photos

Design:
https://www.figma.com/file/0yFVivSWXgFf2ddEF92zkf/Kotti-Design-System?node-id=95%3A0&t=ZKKQp8Rmars5Zf3R-0


packages/kotti-ui/source/index.ts
```diff
@@ -36,6 +36,8 @@ import {
 	KtFieldDateTime,
 	KtFieldDateTimeRange,
 } from './kotti-field-date'
+import { KtFieldFileUpload } from './kotti-field-file-upload'
+export * from './kotti-field-file-upload'
```
FlorianWendelborn, line 40: Sort

packages/kotti-ui/source/kotti-field-file-upload/KtFieldFileUpload.vue
```diff
@@ -0,0 +1,143 @@
...
+		onBeforeMount(() =>
+			field.currentValue.forEach((file) =>
+				filesWithStatusOverridden.add(file.id),
+			),
+		)
+
+		return {
+			field: computed(() => ({
+				...field,
+				validation: validation.value ?? field.validation,
```
FlorianWendelborn, line 95: Not 100% sure but likely causes issues with KtFormSubmit

packages/documentation/pages/usage/components/form-fields.vue
```diff
@@ -867,6 +948,10 @@ export default defineComponent({
 			validation: Kotti.Field.Validation.Result['type']
 		}>({
 			additionalProps: {
```
FlorianWendelborn, line 950: `remoteUpload` and potentially others are undocumented

packages/kotti-ui/source/kotti-field-file-upload/KtFieldFileUpload.vue
```diff
@@ -0,0 +1,143 @@
...
+			})),
+			fileInfoList: computed<KottiFieldFileUpload.FileInfo[]>(() =>
+				field.currentValue.map((file) =>
+					!filesWithStatusOverridden.has(file.id) &&
+					props.remoteUpload &&
```
FlorianWendelborn, line 100: Use ?? And ?. operators

packages/kotti-ui/source/kotti-field-file-upload/components/FileItem.vue
```diff
@@ -0,0 +1,270 @@
...
+						if (props.remoteActions?.onDelete)
+							props.remoteActions.onDelete(props.fileInfo.id)
+				}
+			},
+			onClickRetry: () => {
+				if (!props.isDisabled && props.remoteActions?.onRetry)
+					props.remoteActions.onRetry(props.fileInfo.id)
+			},
+			progress: computed(() =>
+				Number.parseInt(props.fileInfo.progress.toFixed(0), 10),
```
FlorianWendelborn, line 179: Is that supposed to be a Math.floor?

packages/kotti-ui/source/kotti-field-file-upload/constants.ts
```diff
@@ -0,0 +1,10 @@
+import { KottiField } from '../kotti-field/types'
+
+export const ISO8601_SECONDS = 'YYYY-MM-DD HH:mm:ss' as const
+
+export const KOTTI_FIELD_FILE_UPLOAD_SUPPORTS: KottiField.Supports = {
+	clear: false,
+	decoration: false,
+	placeholder: false,
+	tabIndex: false,
```
FlorianWendelborn, line 9: I think this should be possible to support here

packages/kotti-ui/source/kotti-field-file-upload/types.ts
```diff
@@ -0,0 +1,125 @@
+import { yocoIconSchema } from '@3yourmind/yoco'
+import { z } from 'zod'
+
+import { KottiField } from '../kotti-field/types'
+
+export namespace KottiFieldFileUpload {
+	export enum Status {
```
FlorianWendelborn, line 7: If possible enums should have the same key and value for debugging purposes

packages/kotti-ui/source/kotti-field-file-upload/types.ts
```diff
@@ -0,0 +1,125 @@
...
+		FINISHED = 'finished',
+		NOT_STARTED = 'notStarted',
+		PROCESSING = 'processing',
+		UPLOADED = 'uploaded',
+		UPLOADING = 'uploading',
+		READY_TO_UPLOAD = 'readyToUpload',
+	}
+	export const statusSchema = z.nativeEnum(Status)
+
+	export enum Validation {
```
FlorianWendelborn, line 19: Same here

packages/kotti-ui/source/kotti-field-file-upload/types.ts
```diff
@@ -0,0 +1,125 @@
...
+		ERROR_MULTIPLE = 'errorMultiple',
+		ERROR_SIZE = 'errorSize',
+		ERROR_TYPE = 'errorType',
+		SUCCESS = 'success',
+	}
+	export const validationSchema = z.nativeEnum(Validation)
+
+	export const fileItemSchema = z.object({
+		blob: z.custom<File>(),
+		id: z.string(),
```
FlorianWendelborn, line 29: If this is a uuid, you can .uuid()

packages/kotti-ui/source/kotti-field-file-upload/types.ts
```diff
@@ -0,0 +1,125 @@
...
+			.default(null),
+		onUpload: z.function().args(z.string()).returns(z.void()),
+	})
+	export type RemoteActions = z.output<typeof remoteActionsSchema>
+
+	export const fileInfoSchema = z.object({
+		id: z.string(),
+		name: z.string(),
+		progress: z.number().default(0),
+		size: z.number(),
```
FlorianWendelborn, line 64: .int().min(0)

packages/kotti-ui/source/kotti-field-file-upload/utils/common.ts
```diff
@@ -0,0 +1,57 @@
+import dayjs from 'dayjs'
+import filesize from 'filesize'
+
+import { ISO8601_SECONDS } from '../constants'
+import { KottiFieldFileUpload } from '../types'
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const nanoid = require('nanoid')
```
FlorianWendelborn, line 8: Should use import

packages/kotti-ui/source/kotti-field-file-upload/utils/common.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const formatFileSize = (fileSize: number): string =>
+	filesize(fileSize, { round: 1, spacer: '' })
```
FlorianWendelborn, line 11:
```suggestion
	filesize(fileSize, { base: 2, round: 1, spacer: '' })
```

Unless necessary, I'd also remove the spacer override to ease readability

packages/kotti-ui/source/kotti-field-file-upload/utils/common.ts
```diff
@@ -0,0 +1,57 @@
...
+				id: nanoid(),
+				name: file.name,
+		  }
+		: {
+				blob: file as File,
+				id: nanoid(),
+				name: `${dayjs().format(ISO8601_SECONDS)}.png`,
+		  }
+
+export const getValidFileTypes = (
```
FlorianWendelborn, line 45: I don't understand what this is supposed to do, so either refactor it or add a JSDoc

packages/kotti-ui/source/kotti-field-file-upload/validators.ts
```diff
@@ -0,0 +1,62 @@
...
+	fileSize: number,
+	maxFileSize: KottiFieldFileUpload.PropsInternal['maxFileSize'],
+): boolean => maxFileSize === null || fileSize < maxFileSize
+
+const isValidFileType = (
+	fileName: string,
+	validFileTypes: Array<string> | null,
+): boolean =>
+	validFileTypes === null ||
+	validFileTypes.includes(fileName.toLowerCase().split('.')[1])
```
FlorianWendelborn, line 13: Fails for `something.png.jpg`

packages/kotti-ui/source/kotti-field-file-upload/components/FileItem.vue
```diff
@@ -0,0 +1,270 @@
...
+			} else status.value = KottiFieldFileUpload.Status.ERROR
+		})
+
+		watch(
+			() => props.fileInfo.status,
+			(remoteStatus) => {
+				if (props.overrideStatus) return
+
+				if (
+					validation.value === KottiFieldFileUpload.Validation.SUCCESS &&
```
FlorianWendelborn, line 127: This looks quite arbitrary

packages/kotti-ui/source/kotti-field-file-upload/components/FileItem.vue
```diff
@@ -0,0 +1,270 @@
...
+				() =>
+					validation.value === KottiFieldFileUpload.Validation.SUCCESS &&
+					[
+						KottiFieldFileUpload.Status.CANCELED,
+						KottiFieldFileUpload.Status.ERROR,
+					].includes(status.value) &&
+					props.remoteActions?.onRetry,
+			),
+			Yoco,
+			status,
```
FlorianWendelborn, line 197: Sort

packages/kotti-ui/source/kotti-field-file-upload/components/ProgressBar.vue
```diff
@@ -0,0 +1,65 @@
...
+	</div>
+</template>
+
+<script lang="ts">
+import { computed, defineComponent } from '@vue/composition-api'
+
+export default defineComponent({
+	name: 'ProgressBar',
+	props: {
+		error: { default: false, type: Boolean },
```
FlorianWendelborn, line 16:
```suggestion
		isError: { default: false, type: Boolean },
```

packages/kotti-ui/source/kotti-field-file-upload/components/ProgressBar.vue
```diff
@@ -0,0 +1,65 @@
...
+		border-radius: var(--unit-1);
+
+		&__fill {
+			display: block;
+			height: 100%;
+			background-color: var(--interactive-03);
+			border-radius: var(--unit-1);
+			transition: width 500ms ease-in-out;
+
+			&--error {
```
FlorianWendelborn, line 51:
```suggestion
			&--is-error {
```

FlorianWendelborn (changes requested):

packages/kotti-ui/source/kotti-i18n/locales/de-DE.ts
```diff
@@ -27,6 +27,48 @@ export const deDE: KottiI18n.Messages = {
...
+			notSupported: 'Keine kompatible Kamera verfügbar',
+		},
+		label: {
+			capture: 'Foto machen',
+			error: 'Error',
+			review: 'Überprüfen',
+			unknown: 'Unbekannt',
+		},
+		statusMsg: {
+			CANCELED: 'Abgebrochen',
```
carsoli, line 51: Minor

this is english-us - I'm not sure if we have a convention for english-uk vs english-us

I personally don't care and didn't even find other instances of cancelled

but if you know a word that distinguishes US from UK english and you see that kotti has a tendency to use either, follow the one it follows

packages/kotti-ui/source/kotti-i18n/locales/de-DE.ts
```diff
@@ -27,6 +27,48 @@ export const deDE: KottiI18n.Messages = {
...
+			PROCESSING: 'Verarbeiten...',
+			UPLOADED: 'Hochgeladen',
+			UPLOADING: 'Hochladen...',
+			READY_TO_UPLOAD: 'Bereit zum Hochladen bei Einreichung',
+		},
+		text: {
+			clickToUpload: 'Zum Hochladen klicken',
+			dragAndDrop: 'oder ziehen und ablegen',
+			learnMore: 'Mehr erfahren',
+			max: 'max.',
```
carsoli, line 64: why is punctuated?

carsoli: is it the abbreviation of maximal?

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
```diff
@@ -0,0 +1,276 @@
...
+	) {
+		const translations = useTranslationNamespace('KtFieldFileUpload')
+
+		const dropAreaRef = ref<HTMLDivElement | null>(null)
+		const isDragging = ref<boolean>(false)
+		const isHovering = ref<boolean>(false)
+		const uploadInputRef = ref<HTMLInputElement | null>(null)
+
+		const emitFiles = (files: FileList | null) => {
+			if (props.isDisabled || !files || files.length === 0) return
```
carsoli, line 113: could be shortened
```
if(props.isDisabled || !files?.length)```

stylistic preference

File: packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue [hunk 22]
```diff
@@ -0,0 +1,276 @@
...
+		dataTest: { default: null, type: String },
+		icon: { default: null, type: String },
+		infoUrl: { default: null, type: String },
+		isDisabled: { default: false, type: Boolean },
+		maxFileSize: { default: null, type: Number },
+		tabIndex: { default: 0, type: Number },
+	},
+	setup(
+		props: Pick<
+			KottiFieldFileUpload.PropsInternal,
```
carsoli, line 92: Would it have made more sense to just scope the props within the namespace of the KottiFieldFileUpload

I will leave the example comment on the types file

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
```diff
@@ -0,0 +1,276 @@
...
+		:tabIndex="_tabIndex"
+		@click.prevent="onClick"
+		@dragleave.prevent.stop="isDragging = false"
+		@dragover.prevent.stop="isDragging = true"
+		@drop.prevent.stop="onDrop"
+		@mouseenter.prevent.stop="isHovering = true"
+		@mouseleave.prevent.stop="isHovering = false"
+	>
+		<div class="kt-field-file-upload__drop-area__description">
+			<i class="yoco" v-text="_icon" />
```
carsoli, line 15: why is it prop-customizable, do we have a use-case to customise the icon from outside?

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
```diff
@@ -0,0 +1,276 @@
...
+				v-if="infoUrl"
+				:href="infoUrl"
+				rel="noopener noreferrer"
+				target="_blank"
+				@click.stop
+				v-text="translations.text.learnMore"
+			/>
+		</div>
+		<input
+			v-show="false"
```
carsoli, line 34: :) out of curiosity

did you run into issues with the css display none along with using a reference?

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
```diff
@@ -0,0 +1,276 @@
...
+			type="file"
+			@change="onSelect"
+			@click.stop
+		/>
+		<div
+			v-if="$slots.content"
+			class="kt-field-file-upload__drop-area__content-slot"
+			@click.stop="onContentSlotClick"
+		>
+			<slot name="content" />
```
carsoli, line 48: 👏🏼

carsoli: Thinking out loud:

do we need to expose anything from this slot to the user-app

i understand why you're using a slot on this level if you detect one on the parent and you have the handler to guard against "disabled" clicks

but if the user-app uses @click.stop on the slot--> it won't bubble up the click event anyways

so what I'm thinking is: "is it better design to just expose the isDisabled prop?"

packages/kotti-ui/source/kotti-i18n/locales/de-DE.ts
santiagoballadares: Didn't find anything about which is preferred in readme. Assumed it is en-us as 'gray' is used instead of 'grey'. There's also this commit f992d3a

packages/kotti-ui/source/kotti-i18n/locales/de-DE.ts
santiagoballadares: Yes, that's an abbreviation as indicated in the design

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
santiagoballadares: This was requested by design

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
santiagoballadares: Just tested, no issues, it's working. 
I just did it this way, will change it.

carsoli: you don't need to change it, I was curious if it'd caused it - but I don't see any reason why it shouldn't be a v-show

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
```diff
@@ -0,0 +1,265 @@
...
+
+		const dropAreaRef = ref<HTMLDivElement | null>(null)
+		const isDragging = ref<boolean>(false)
+		const isHovering = ref<boolean>(false)
+		const uploadInputRef = ref<HTMLInputElement | null>(null)
+
+		const emitFiles = (files: FileList | null) => {
+			if (props.isDisabled || !files || files.length === 0) return
+
+			const { accept, allowMultiple, maxFileSize } = props
```
carsoli, line 108: props is a `reactive`

IF (but since it's not callled useValidateSelection, I'll asume it's a NO) you rely on reactivity of those props, you'd have just destroyed reactivity

carsoli: I'd say regardless - change it, that would be a nasty bug to track down again

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
```diff
@@ -0,0 +1,265 @@
...
+				[KEY_ENTER, KEY_SPACE].includes(event.key) &&
+				dropAreaRef.value === window.document.activeElement
+			) {
+				event.preventDefault()
+				event.stopPropagation()
+				uploadInputRef.value?.click()
+			}
+		}
+
+		onBeforeMount(() => window.addEventListener('keydown', dropAreaListener))
```
carsoli, line 138: I'm not sure if performance-wise it'd be more optimal but we can research 

But I'd personally add the event listener specifically to the keydown.KEY_ENTER, and another for the keydown.SPACE

imagine the amount of times this would get called within a window

technically it's very uncosty because you v-if but **MAYBE** performance-wise it's better to directly watch on the specific event

carsoli: also, out of curiosity

did you try just listening on the element with the ref: `uploadInputRef` directly


```
ref="uploadInputRef"
@keydown.space.stop.prevent="handleDropAreaFocus"
@keydown.enter.stop.prevent="handleDropAreaFocus"
```

carsoli: Sorry - I'm confused :D

the @click on this element is @click.stop without a handler

why do you programatically trigger a click?

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
```diff
@@ -0,0 +1,265 @@
...
+
+		const dropAreaListener = (event: KeyboardEvent) => {
+			if (
+				!props.isDisabled &&
+				[KEY_ENTER, KEY_SPACE].includes(event.key) &&
+				dropAreaRef.value === window.document.activeElement
+			) {
+				event.preventDefault()
+				event.stopPropagation()
+				uploadInputRef.value?.click()
```
carsoli, line 134: https://github.com/3YOURMIND/kotti/pull/719/files#r1108635019

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
```diff
@@ -0,0 +1,265 @@
...
+						: null,
+				]
+					.filter((value) => Boolean(value))
+					.join(' ')
+			}),
+			isHovering,
+			isDragging,
+			onClick: () => {
+				if (!uploadInputRef.value || props.isDisabled) return
+				uploadInputRef.value.click()
```
carsoli, line 173: same question as 

what am I missing

the onClick programatically triggers a click on `uploadInputRef` which does nothing with the click 

https://github.com/3YOURMIND/kotti/pull/719/files#r1108635019

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
```diff
@@ -0,0 +1,265 @@
...
+		}
+
+		onBeforeMount(() => window.addEventListener('keydown', dropAreaListener))
+
+		onUnmounted(() => window.removeEventListener('keydown', dropAreaListener))
+
+		return {
+			_icon: computed(() => props.icon ?? Yoco.Icon.CLOUD_UPLOAD),
+			_tabIndex: computed(() =>
+				props.isDisabled ? null : props.tabIndex ?? 0,
```
carsoli, line 145: can be `-1` instead of null

packages/kotti-ui/source/kotti-field-file-upload/validators.ts
```diff
@@ -0,0 +1,70 @@
+import { KottiFieldFileUpload } from './types'
+
+const isValidFileSize = (
+	fileSize: number,
+	maxFileSize: number | null,
+): boolean => maxFileSize === null || fileSize < maxFileSize
+
+const isValidFileType = (
+	allowedFileTypes: string | null,
```
carsoli, line 9: weird choice of types - why support the extensions argument as a concatenated string vs an array?

carsoli: i mean not even a regex, right? it ends being a lower-case string that we use "include" on

which is very risky

I mean - theoratically - we could end up getting the end of a string and the beginning of another as a format

--

Feels like legacy implementation coming from yoda because i remember seeing this before but I would totally advise against this :/

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
santiagoballadares: Added `@keydown.space` and `@keydown.enter` instead of the listener but they need to go on the DIV container and not the input because it's hidden (intended)

The problem was the label that captures all clicks inside including clicks on the items bellow the drop area. If you remember it also messes up with the focus of the buttons inside. 

Anyway did some refactor, it should be good now

I still need to trigger the input click event on the DIV keydown events to emulate the focus behavior

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
santiagoballadares: Explained on comment above

packages/kotti-ui/source/kotti-field-file-upload/components/DropArea.vue
santiagoballadares: This was due to the label which captures all clicks inside and then forwards the events to the first child inside. It's now refactored.

packages/kotti-ui/source/kotti-field-file-upload/validators.ts
santiagoballadares: Not really a preference, either way will have to check the strings as we could have almost any input
I refactored it to an array pull latest changes and let me know what you think

packages/documentation/pages/usage/components/form-fields.vue
```diff
@@ -1179,13 +1382,55 @@ export default defineComponent({
...
+									return acc
+								},
+								[],
+							),
+						) as Kotti.FieldFileUpload.Remote.Actions,
+						payload: Object.fromEntries(
+							values.value['fileUploadValue'].map((file) => [
+								file.id,
+								{
+									progress: +Math.random().toPrecision(2),
```
Juli0GT, line 1426: progess is randomly calculated? @santiagoballadares

packages/documentation/pages/usage/components/form-fields.vue
```diff
@@ -1231,6 +1470,15 @@ export default defineComponent({
 			isRangeComponent,
 			onSubmit: (values: Record<string, unknown>) =>
 				window.alert(`@submit: ${JSON.stringify(values, null, '\t')}`),
+			remoteUploadPayload: computed(() => {
+				const remoteUpload = componentRepresentation.value.props[
+					'remoteUpload'
+				] as Kotti.FieldFileUpload.Remote.Upload
+
+				return remoteUpload?.payload
+					? JSON.stringify(remoteUpload.payload, null, '	').replace(/"/g, "'")
```
Juli0GT, line 1479: What is being achieved here?

packages/kotti-ui/source/kotti-i18n/locales/de-DE.ts
FlorianWendelborn: @carsoli convention is using en-US, we even had fixes for accidental UK English use in the past

Isokaeder:
The text in the dropdown area is not centered. The anchor element has a margin, is that intentional?

Screenshot 2023-04-18 at 12 23 48

Isokaeder (approved):

packages/documentation/pages/usage/styles/form-fields.scss
```diff
@@ -98,6 +98,7 @@ h3 {
 		&__value {
 			span {
 				font-family: monospace;
+				word-wrap: anywhere;
```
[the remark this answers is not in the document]
santiagoballadares: Replaced with `word-break: break-all;`

packages/kotti-ui/source/kotti-field-file-upload/validators.ts
```diff
@@ -0,0 +1,50 @@
+import { KottiFieldFileUpload } from './types'
+
+export const isSelectingMultipleFilesWhenNotAllowed = (
+	allowMultiple: boolean,
+	selectionSize: number,
+): boolean => !allowMultiple && selectionSize > 1
+
+const isValidFileSize = (fileSize: number, maxFileSize: number): boolean =>
+	fileSize < maxFileSize
```
carsoli, line 9: @santiagoballadares why not `<=`

packages/kotti-ui/source/kotti-field-file-upload/validators.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const isValidFileExtension = (
+	fileName: string,
+	validFileExtensions: Array<string>,
+): boolean => {
+	if (validFileExtensions.length === 0) return true
+
+	const fileExtension = (fileName.split('.').pop() ?? '').trim().toUpperCase()
```
carsoli, line 17: For future reference
`string.prototype.localCompare` could be useful with its options tweaked, because it can ignore casing and punctation (which includes spaces)
check the docs for it

I tried this example and they both return `0` which means they are equal

```
'jpg.'.localeCompare(', JPG', undefined, {ignorePunctuation: true, sensitivity: "accent"})
```

carsoli: in that case you'd just
```
const extension = fileName.split('.').pop() ?? null
if(extension === null) return false

return validFileExtensions.some(e => e.localeCompare(extension, undefined, {ignorePunctuation: true, sensitivity: "accent"}) === 0)
```

packages/kotti-ui/source/kotti-field-file-upload/validators.ts
```diff
@@ -0,0 +1,50 @@
...
+
+	return validFileExtensions
+		.map((extension) => extension.trim().toUpperCase())
+		.includes(fileExtension)
+}
+
+export const isValidUploadedFileStatus = (
+	valueItem: KottiFieldFileUpload.ValueInternal[0],
+) =>
+	valueItem.status === KottiFieldFileUpload.Status.Common.INVALID ||
```
carsoli, line 29: I thought it's called UPLOADED_WITH_ERROR

@santiagoballadares

packages/kotti-ui/source/kotti-field-file-upload/validators.ts
```diff
@@ -0,0 +1,50 @@
...
+	fileSize,
+	maxFileSize,
+}: {
+	extensions: KottiFieldFileUpload.PropsInternal['extensions']
+	fileName: string
+	fileSize: number
+	maxFileSize: KottiFieldFileUpload.PropsInternal['maxFileSize']
+}): KottiFieldFileUpload.Validation => {
+	if (!isValidFileExtension(fileName, extensions))
+		return KottiFieldFileUpload.Validation.INVALID_EXTENSION
```
carsoli, line 44: Question:

If both fail (max size and file extension), you will give priority to the Invalid Extension error first

Aka you don't show an array, you show the first error first then I'd assume you revalidate

Why don't you return an array of all errors - 

Have you aligned with alberto on this

packages/kotti-ui/source/kotti-field-file-upload/types.ts
```diff
@@ -0,0 +1,264 @@
...
+		export const nonRemoteSchema = z.nativeEnum(NonRemote)
+
+		export enum Remote {
+			CANCELED = 'CANCELED',
+			ERROR = 'ERROR',
+			UPLOADING = 'UPLOADING',
+		}
+		export const remoteSchema = z.nativeEnum(Remote)
+
+		export const schema = commonSchema.or(nonRemoteSchema).or(remoteSchema)
```
carsoli, line 30: according to docs, there is no difference between the or and union

have u experienced anything in type validation that was different when executed

packages/kotti-ui/source/kotti-field-file-upload/types.ts
```diff
@@ -0,0 +1,264 @@
...
+			status: Status.schema,
+		})
+		.or(
+			z.object({
+				downloadUrl: z.string().optional(),
+				id: z
+					.string()
+					.uuid()
+					.optional()
+					.default(() => uuidv4()),
```
carsoli, line 54: remember how we saw in Vue that the default doesn't work on that level :) 

because it can only default on a prop level

packages/kotti-ui/source/kotti-field-file-upload/types.ts
```diff
@@ -0,0 +1,264 @@
...
+		}
+		export const commonSchema = z.nativeEnum(Common)
+
+		export enum NonRemote {
+			READY_TO_UPLOAD = 'READY_TO_UPLOAD',
+		}
+		export const nonRemoteSchema = z.nativeEnum(NonRemote)
+
+		export enum Remote {
+			CANCELED = 'CANCELED',
```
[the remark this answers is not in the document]
santiagoballadares: It's US english

packages/kotti-ui/source/kotti-field-file-upload/types.ts
```diff
@@ -0,0 +1,264 @@
...
+			value: valuesSchema.default(() => []),
+		})
+
+	export type Props = z.input<typeof propsSchema>
+	export type PropsInternal = z.output<typeof propsSchema>
+
+	export namespace Events {
+		export type AddFiles = KottiFieldFileUpload.ValueInternal
+
+		export type RemoveFile = string
```
carsoli, line 139: I guess that's the id

packages/kotti-ui/source/kotti-field-file-upload/types.ts
```diff
@@ -0,0 +1,264 @@
...
+		export type SetStatus = {
+			id: string
+			status: Status.All
+		}
+	}
+
+	export namespace ActionButton {
+		export const schema = propsSchema
+			.pick({
+				isDisabled: true,
```
carsoli, line 150: not tabbable? why not tabIndex?

carsoli: Assuming it is indeed tabbable

packages/kotti-ui/source/kotti-field-file-upload/validators.ts
santiagoballadares: Yes, Alberto validated that file extension error has priority.

packages/kotti-ui/source/kotti-field-file-upload/validators.ts
carsoli: I saw the usage of this later

looks to me like you intentionally want to return true even if the extension and size validators fail

which doesn't comply with the last discussion we had about UPLOADED_WITH_ERROR for already uploaded files

packages/kotti-ui/source/kotti-field-file-upload/KtFieldFileUpload.vue
```diff
@@ -0,0 +1,246 @@
...
+
+		watch(
+			() => props.remoteUpload?.payload,
+			(payload) => {
+				if (!payload || isEmpty(payload)) return
+
+				field.currentValue.forEach((fileItem) => {
+					if (
+						!(
+							fileItem.status === KottiFieldFileUpload.Status.Remote.CANCELED ||
```
carsoli, line 180: I understand if you want to skip the setStatus as long as it's uploading

but if you skip the Error and Cancelled, how would they be set from outside the first time..

the first time being when this watcher triggers where newPayload is Canceled but oldPayload is Uploading

carsoli: IMO, unless I'm missing something

you'd only ignore the cancelled and error if they are set AGAIN

you need both the old and new value in the watcher to be able to tell though

packages/kotti-ui/source/kotti-field-file-upload/types.ts
santiagoballadares: id is no longer optional

packages/kotti-ui/source/kotti-field-file-upload/types.ts
```diff
@@ -0,0 +1,264 @@
...
+			status: z
+				.enum([
+					Status.Common.UPLOADED,
+					Status.Remote.CANCELED,
+					Status.Remote.ERROR,
+					Status.Remote.UPLOADING,
+				])
+				.default(Status.Remote.UPLOADING),
+		})
+		export const payloadSchema = z.record(payloadEntrySchema)
```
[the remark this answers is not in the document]
santiagoballadares: payload is a z.record to identify payloads (status, progress) via the key (file Item id)

packages/documentation/pages/utilities.ts
```diff
@@ -42,6 +46,7 @@ export type ComponentValue = {
 	defaultSlot: string | null
 	hasActions: boolean
 	hasHelpTextSlot: boolean
+	hasRemoteUpload: boolean
```
carsoli, line 49: is this still needed?

packages/documentation/pages/utilities.ts
```diff
@@ -69,6 +74,47 @@ export const createActions = (
 		  ]
 		: undefined
 
+export const createRemoteUpload = (
+	hasRemoteUpload: boolean,
```
carsoli, line 78: honestly doesn't feel like the arg is needed , you already check for the component name before calling this function

packages/kotti-ui/source/kotti-field-file-upload/KtFieldFileUpload.vue
```diff
@@ -0,0 +1,214 @@
...
+	props: makeProps(KottiFieldFileUpload.propsSchema),
+	setup(props, { emit }) {
+		const field = useField<KottiFieldFileUpload.ValueInternal>({
+			emit,
+			/**
+			 * We consider this field empty if:
+			 * a. No files are selected, or
+			 * b. All selected files are invalid
+			 */
+			isEmpty: (value) =>
```
carsoli, line 95: i think this can easily be a util that accepts the value

carsoli: bc it's shared between both vue components

packages/kotti-ui/source/kotti-field-file-upload/KtFieldFileUpload.vue
```diff
@@ -0,0 +1,214 @@
...
+						maxFileSize: props.maxFileSize,
+					})
+
+					if (fileItem.status !== fileInfo.status)
+						setStatus({ id: fileItem.id, status: fileInfo.status })
+
+					return fileInfo
+				}),
+		)
+		const showDropArea = computed(
```
carsoli, line 128: you don't need to negate the condition again

showDropArea = computed(() => props.allowMultiple || field.currentValue.length === 0)

packages/kotti-ui/source/kotti-field-file-upload/KtFieldFileUpload.vue
```diff
@@ -0,0 +1,214 @@
...
+						? { ...fileItem, status: payload.status }
+						: fileItem,
+				),
+				{ forceUpdate: true },
+			)
+		}
+
+		onBeforeMount(() => {
+			field.currentValue.forEach((fileItem) => {
+				if (
```
carsoli, line 147: shorter and more readable if
```
![undefined, Kotti.FieldFileUpload.Status.UPLOADED, Kotti.FieldFileUpload.Status.UPLOADED_WITH_ERROR].includes(fileItem.status)
```

up to you 

if you wrote it so ellaborately because of TS issues, then it makes sense, leave it as is :)

packages/kotti-ui/source/kotti-field-file-upload/components/ItemLayout.vue
```diff
@@ -0,0 +1,72 @@
...
+
+<script lang="ts">
+import { defineComponent } from '@vue/composition-api'
+
+export default defineComponent({
+	name: 'ItemLayout',
+	props: {
+		name: { required: true, type: String },
+	},
+	setup() {
```
carsoli, line 26: remove it all together

packages/kotti-ui/source/kotti-field-file-upload/components/TakePhoto/Error.vue
```diff
@@ -0,0 +1,57 @@
...
+			<h2 v-text="translations.label.error" />
+		</div>
+		<div slot="body">
+			<div v-if="error" v-text="errorText" />
+		</div>
+		<div slot="footer">
+			<KtButton
+				class="mr-4"
+				:data-test="dataTest ? `${dataTest}.retakePhoto` : undefined"
+				:label="translations.button.retry"
```
carsoli, line 13: i'm not really sure this would be reactive

can you try changing translations and seeing if the values recompute on demand?

packages/kotti-ui/source/kotti-field-file-upload/components/TakePhoto/Error.vue
```diff
@@ -0,0 +1,57 @@
...
+				type="primary"
+				@click.stop="$emit('retry')"
+			/>
+			<KtButton
+				:data-test="dataTest ? `${dataTest}.closeError` : undefined"
+				:label="translations.button.cancel"
```
carsoli, line 19: same here

packages/kotti-ui/source/kotti-field-file-upload/components/TakePhoto/utils/common.ts
```diff
@@ -0,0 +1,47 @@
+import dayjs from 'dayjs'
+
+const ISO8601 = 'YYYY-MM-DD HH:mm:ss' as const
```
carsoli, line 3: this should be moved to kotti-ui's global constants if possible

I think we had a similar constant for the date fields (maybe unify both of them)

packages/kotti-ui/source/kotti-field-file-upload/index.ts
```diff
@@ -0,0 +1,51 @@
...
+import KtFieldFileUploadRemoteVue from './KtFieldFileUploadRemote.vue'
+import { KottiFieldFileUpload, KottiFieldFileUploadRemote } from './types'
+
+const componentFolder = 'kotti-field-file-upload'
+const url =
+	'https://www.figma.com/file/0yFVivSWXgFf2ddEF92zkf/Kotti-Design-System?node-id=3547%3A8456&t=hhn880dUtUMnmX06-0'
+export const KtFieldFileUpload = attachMeta(
+	makeInstallable(KtFieldFileUploadVue),
+	{
+		addedVersion: '3.0.0-beta.43',
```
carsoli, line 16: well that's recently changed :D

packages/kotti-ui/source/kotti-field-file-upload/components/TakePhoto/utils/common.ts
santiagoballadares: Moved to kotti-ui's global constants

But cannot use in date fields bc they don't use ISO8601 format

packages/kotti-ui/source/kotti-field-file-upload/components/TakePhoto/Error.vue
santiagoballadares: Tested, works fine on demand with all translations

packages/kotti-ui/source/kotti-field-file-upload/components/TakePhoto/Error.vue
santiagoballadares: Tested, works fine on demand with all translations

packages/kotti-ui/source/kotti-field-file-upload/KtFieldFileUpload.vue
santiagoballadares: right, i had that problem before when statuses were split and then merged

packages/documentation/pages/utilities.ts
santiagoballadares: It's still needed to generate the component's code when calling `generateComponentCode` from `form-fiels.vue` (it's similar to hasActions for KtFieldSelects).

packages/documentation/pages/utilities.ts
santiagoballadares: Also needed bc it's called internally by `generateComponentCode` to generate the component's code.

packages/kotti-ui/source/kotti-field-file-upload/index.ts
santiagoballadares: should be the right next version

packages/documentation/pages/utilities.ts
```diff
@@ -69,6 +74,47 @@ export const createActions = (
...
+
+	return remoteUpload
+		? `${[
+				...Object.entries(remoteUpload)
+					.map(([key, value]) => {
+						if (key === 'actions')
+							return [
+								`\t:${key}: {`,
+								...Object.keys(value as Record<string, unknown>).map(
+									(k) => `\t\t${k}: (id: number | string) => {},`,
```
carsoli, line 105: isn't this on Vue level 

the syntax looks like JS to me 

should be `{k}=` not `:`

packages/kotti-ui/source/kotti-field-file-upload/KtFieldFileUploadRemote.vue
```diff
@@ -0,0 +1,246 @@
...
+							KottiFieldFileUploadRemote.Status.INVALID,
+							KottiFieldFileUploadRemote.Status.NOT_STARTED,
+							KottiFieldFileUploadRemote.Status.UPLOADED,
+							KottiFieldFileUploadRemote.Status.UPLOADED_WITH_ERROR,
+						].includes(fileItemStatus)
+					)
+						setStatus({ id, status })
+				})
+			},
+			{ deep: true },
```
carsoli, line 191: This will allow deep mutating from outside rather than replacing the payload object all together whenever a remote is updated

Did you test without this?

packages/kotti-ui/source/kotti-field-file-upload/validators.ts
```diff
@@ -0,0 +1,102 @@
...
+ * @param value The value to evaluate
+ * @returns true if at least one selected file is valid, else false
+ */
+export const isValidValue = <
+	T extends
+		| KottiFieldFileUpload.ValueInternal
+		| KottiFieldFileUploadRemote.ValueInternal,
+>(
+	value: T,
+	type: T extends KottiFieldFileUpload.ValueInternal ? 'NonRemote' : 'Remote',
```
carsoli, line 74: did you end up using the generics for common functions only here

I'd keep it here if that's the only case but I'd otherwise move the 'NonRemote' and 'Remote' as a type in the types file

I recall we tried with an enum - if it doesn't work with enum, just make them two standalone types for similar usages

if it's ony used here - keep it as is.

carsoli (approved):

packages/documentation/pages/utilities.ts
carsoli: Oh no nvm - the payload itself is an object, nvm :)

packages/kotti-ui/source/kotti-field-file-upload/validators.ts
santiagoballadares: yes, it's used only here for `isValidValue` bc in both cases (remote, non-remote) the arguments it takes are the same 

for other cases like `buildFileInfo`, which takes different arguments, are separated

packages/kotti-ui/source/kotti-field-file-upload/KtFieldFileUploadRemote.vue
santiagoballadares: Tested, works without this. It's removed now.

PR #725, opened 2023-03-06
docs(form-field): set initial field component with query parameter

Isokaeder:
A little adjustment to the form-fields path:

- Use `component` query parameter to set initially selected form field component
- when selecting a component, the url gets updated with the right query parameter

Intended effect: share and save specific urls for each component


carsoli:
@Isokaeder I personally have review notifications enabled and others disabled

Please add reviewers when PRs are ready so they don't go stale

Issue #667, opened 2022-09-20
Disable KtPopover if all options are disabled


kpm-berlin:
**Is your feature request related to a problem? Please describe.**
If the `KtPopover` is used in combination with options and a `KtButton`, disabling all options and button does not result in the popover to become disabled.

**Describe the solution you'd like**
Add a `disabled` prop to `KtPopover` that controls the opening/closing of options (similar to button's disabled `prop`)

**Describe alternatives you've considered**
/

**Additional context**
Example:
**Code**
```
<KtPopover :options="moreActionsOptions">
				<KtButton
					:disabled="!toggleMoreActions"
					:label="$t('generic.moreActions')"
				/>
</KtPopover>
```
**Screenshot**
![Screenshot 2022-09-20 at 11 23 52](https://user-images.githubusercontent.com/72387574/191220877-a7ef4f1c-725b-4e57-b6a8-0ec8a42f67ef.png)

PR #727, opened 2023-03-13
feature(#667): disable the trigger div when the KtPopover is disabled

RueRivoli:
This PR aims at disabling the trigger div in a KtPopover component when the prop isDisabled is set to true.
This is needed to disable the "more actions" buttons when all the action items are disabled and then prevent the opening of the popover containing the action items

### DIFF
--- a/packages/kotti-ui/source/kotti-popover/KtPopover.vue
+++ b/packages/kotti-ui/source/kotti-popover/KtPopover.vue
@@ -140,7 +140,8 @@ export default defineComponent<KottiPopover.PropsInternal>({
 					showPopover.value = false
 				},
 				onShow: () => {
-					showPopover.value = true
+					if (!props.isDisabled) showPopover.value = true
+					else return false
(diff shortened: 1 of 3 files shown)


Isokaeder:
I added documentation and approve

Isokaeder (approved):

packages/kotti-ui/source/kotti-popover/KtPopover.vue
```diff
@@ -140,7 +140,8 @@ export default defineComponent<KottiPopover.PropsInternal>({
 					showPopover.value = false
 				},
 				onShow: () => {
-					showPopover.value = true
+					if (!props.isDisabled) showPopover.value = true
+					else return false
```
carsoli, line 144: return false??

RueRivoli: yes to not show the modal

PR #729, opened 2023-03-16
improvement(KtFieldToggleGroup/KtFieldRadioGroup): various fixes and improvements

Isokaeder:
#### Problem
<label> was wrapped around the entire radio/toggle option (header + content slot), so clicking something in the content slot will activate the option. This will especially be a problem when there are [interactive elements in the option](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#accessibility_concerns)
Screenshot 2023-03-16 at 17 20 33

#### Fix
`<label>` only wraps around the actual label. Also there is a new slot that is next to the label, that can be used if you need an interactive element in that area.
Screenshot 2023-03-16 at 17 20 11

#### Problem
Layout breaks for long labels, radio buttons can be squished, radio button, checkbox, switches and the helptext icon are aligned to the center, which makes it look unusual

#### Fix
Elements are aligned to the first line of the option label

### Content default style

Text in the content slot now has a default style (smaller font-size)

## Bugfixes
`KtFieldToggleGroup` supports setting a `dataTest` properties per option, however, there was a bug with the logic there so it was not passed into the DOM

### DIFF
--- a/packages/kotti-ui/source/kotti-field-toggle/KtFieldToggleGroup.vue
+++ b/packages/kotti-ui/source/kotti-field-toggle/KtFieldToggleGroup.vue
@@ -7,20 +7,30 @@
 	>
 		<div slot="container" :class="wrapperClasses">
 			<div v-for="option of optionsWithChecked" :key="option.key">
-				<ToggleInner
-					component="label"
-					:inputProps="getInputProps(option)"
-					:isDisabled="field.isDisabled || Boolean(option.isDisabled)"
-					:type="type"
-					:value="option.value"
-					@input="onInput(option.key, $event)"
-				>
-					<slot name="header" :option="option">
-						<div v-text="option.label" />
-					</slot>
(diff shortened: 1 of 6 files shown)


packages/kotti-ui/source/kotti-field-toggle/KtFieldToggleGroup.vue
```diff
@@ -61,17 +69,12 @@ export default defineComponent<KottiFieldToggleGroup.PropsInternal>({
 
 		return {
 			field,
-			getInputProps: (option: KottiFieldToggleGroup.Entry) => {
-				const fieldDataTest = field.inputProps['data-test']
+			inputProps: computed(() => {
```
RueRivoli, line 72: not ordered list of properties

RueRivoli:
@Isokaeder The mouse cursor is a "hand" when hovering over disabled labels for the `KtRadioGroup` while its the "forbidden" symbol for "KtToggleGroup". I guess it would be consistent to display also "forbidden" for disabled labels.

PR #733, opened 2023-03-20
fix(KtFieldToggle): align label suffix to label's first line

Isokaeder:
Small follow up to #729 

There was another misalignment similar to the ones fixed in #729 between a toggle's label and the label suffix (displayed if the `label` prop's value is `null`

This is how it used to look:
Before fix

This is how it looks now:
After fix


carsoli (approved):

Isokaeder:
I just found out that actually we have the same issue for normal labels

Screenshot 2023-03-21 at 09 45 10

I think this should be handled as well, and possibly in the scope of this MR, as it also has implications on this fix.

PR #734, opened 2023-03-24
fix(KtCard): footer padding/ padding between elements

carsoli:
- fix issues with missing footer padding 
- fix/refact: remove margin/padding manipulation of elements used within the slot (should be controlled by user-app if needed)
- refact: unify slot implementation

- simplify classes that depend on image position
- introduce kt-card mod classes depending on image position to remove kt-card padding
- enhance docs by introducing a select for changeable imgPosition
- changing from md to vue syntax since we have the componentMetadata at the top of the doc page

### DIFF
--- a/packages/documentation/pages/usage/components/card.vue
+++ b/packages/documentation/pages/usage/components/card.vue
@@ -1,121 +1,51 @@
-<template lang="md">
-<ComponentInfo v-bind="{ component }" />
+<template>
+	<div>
+		<ComponentInfo v-bind="{ component }" />
 
-## Structure
-
-<div class="element-example">
-	<KtCard>
-		<div slot="card-header" >
-			<h2>Lorem Ipsum</h2>
-			<b>consectetur adipiscing elit</b>
-		</div>
-		<div slot="card-body" >
-			<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit.
-				Phasellus consequat nisl at nisl condimentum vehicula.
-			</p>
-		</div>
-		<div slot="card-footer">
-			<KtButton label="Button" />
-		</div>
-	</KtCard>
-</div>
-
-The Card component has slots for the `card-header`, `card-body`, and `card-footer`:
-
-```html
-<KtCard>
-	<div slot="card-header">
-		<h2>Lorem Ipsum</h2>
-		<b>consectetur adipiscing elit</b>
-	</div>
-	<div slot="card-body">
-		<p>
-			Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus
-			consequat nisl at nisl condimentum vehicula.
-		</p>
-	</div>
-	<div slot="card-footer">This is the end</div>
-</KtCard>
-```
-
-## Images and Position
-
-<div class="element-example">
-	<KtRow :gap="16">
-		<KtCol :span="8">
-			<KtCard imgUrl="https://picsum.photos/900/300">
-				<div slot="card-header" >
+		<div class="element-example">
+			<KtCard>
+				<div slot="card-header">
 					<h2>Lorem Ipsum</h2>
+					<b>consectetur adipiscing elit</b>
 				</div>
-				<div slot="card-body" >
+				<div slot="card-body">
 					<p>
-						Lorem ipsum dolor sit amet, consectetur adipiscing elit.
-						Phasellus consequat nisl at nisl condimentum vehicula.
+						Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus
+						consequat nisl at nisl condimentum vehicula.
 					</p>
 				</div>
 				<div slot="card-footer">
-					This is the end
+					<KtButton label="Button" />
 				</div>
 			</KtCard>
-		</KtCol>
-		<KtCol :span="8">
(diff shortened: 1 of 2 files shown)


packages/kotti-ui/source/kotti-card/KtCard.vue
```diff
@@ -27,19 +27,13 @@ export default defineComponent<KottiCard.PropsInternal>({
...
-				'kt-card__image-row': true,
-				'kt-card__image-row--is-bottom':
-					props.imgPosition === KottiCard.ImagePosition.BOTTOM,
-				'kt-card__image-row--is-middle':
-					props.imgPosition === KottiCard.ImagePosition.MIDDLE,
-				'kt-card__image-row--is-top':
-					props.imgPosition === KottiCard.ImagePosition.TOP,
+			cardClass: computed(() => ({
+				'kt-card': true,
+				[`kt-card--has-${props.imgPosition}-image`]: true,
```
Isokaeder, line 32: This line has a probably unintended consequence:
Since the default value is `top`, this line will cause the card to have a zero padding on top, see screenshot from the documentation

Screenshot 2023-03-24 at 04 50 47

packages/kotti-ui/source/kotti-card/KtCard.vue
```diff
@@ -53,21 +47,19 @@ export default defineComponent<KottiCard.PropsInternal>({
 	flex-direction: column;
 	word-break: break-all;
```
Isokaeder, line 48: Apparently I added this with my refactor, at least it is not present in the [original css](https://github.com/3YOURMIND/kotti/pull/575/files#diff-0a1dd74771df923f355b00ac7962ef75f54ff5f4567989ff40d0b854417c47c6).

This forces very ugly line breaks (see screenshot in my other comment, can also be seen in [current documentation](https://3yourmind.github.io/kotti/usage/components/card)). I don't know why I did that. I'd say remove or at least use `break-word` instead of `break-all`

Isokaeder (changes requested):

packages/kotti-ui/source/kotti-card/KtCard.vue
carsoli: ah good point, same for bottom, I'll make the condition `hasImage` instead of true

packages/kotti-ui/source/kotti-card/KtCard.vue
carsoli: can do break-word, no worries

packages/kotti-ui/source/kotti-card/KtCard.vue
```diff
@@ -27,19 +27,13 @@ export default defineComponent<KottiCard.PropsInternal>({
...
+				'kt-card': true,
+				[`kt-card--has-${props.imgPosition}-image`]: props.imgUrl,
 			})),
-			footerClass: computed(() => ({
-				'kt-card__footer': true,
-				'kt-card__footer--is-last':
-					props.imgUrl && props.imgPosition !== KottiCard.ImagePosition.BOTTOM,
+			imageRowClass: computed(() => ({
+				'kt-card__image-row': true,
+				[`kt-card__image-row--is-${props.imgPosition}`]: props.imgUrl,
```
carsoli, line 36: technically don't need `props.imgUrl` here because there's a v-if on the template anyways

packages/documentation/pages/usage/components/card.vue
```diff
@@ -1,121 +1,51 @@
...
-	<div slot="card-body">
-		<p>
-			Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus
-			consequat nisl at nisl condimentum vehicula.
-		</p>
+		<KtRow :gap="8">
+			<KtCol :span="12">
+				<KtFieldSingleSelect
+					v-model="imgPosition"
+					hideClear
```
Isokaeder, line 27: Minor: right now the order of options is "bottom - middle - top" which is weird.

Isokaeder (approved):

packages/documentation/pages/usage/components/card.vue
carsoli: Can we ignore it for now, I wanna merge and publish for your labels MR :D

packages/kotti-ui/source/kotti-card/KtCard.vue
```diff
@@ -51,23 +45,21 @@ export default defineComponent<KottiCard.PropsInternal>({
...
 	&__header {
 		order: 2;
-
-		h1,
-		h2,
-		h3,
-		h4,
-		h5,
-		h6 {
-			margin-top: 0;
```
carsoli, line 68: just to clarify, this was removed because we shouldn't control the padding of deeper components, it's the decision of the user app

packages/kotti-ui/source/kotti-card/KtCard.vue
```diff
@@ -79,29 +71,17 @@ export default defineComponent<KottiCard.PropsInternal>({
...
 	&__image-row {
-		padding-top: var(--unit-4);
-
 		&__image {
 			display: block;
 			max-width: 100%;
-			height: auto;
 		}
 
+		margin: var(--unit-4) calc(-1 * var(--unit-4));
```
carsoli, line 80: negative right/left margin to counteract the padding of `kt-card`

PR #737, opened 2023-04-11
fix(KtAvatar): image & fallback to grow in height

Juli0GT:
This MR aims to prevent KtAvatar image to grow in height bigger than its width. This was happening when siblings elements have a bigger height.

### DIFF
--- a/packages/kotti-ui/source/kotti-avatar/KtAvatar.vue
+++ b/packages/kotti-ui/source/kotti-avatar/KtAvatar.vue
@@ -98,6 +98,7 @@ export default defineComponent<KottiAvatar.PropsInternal>({
 	background: var(--ui-02);
 	border: 0.1rem solid var(--white);
 	border-radius: 100%;
+	aspect-ratio: 1/1;
 
 	&__fallback,
 	&__image {
@@ -106,9 +107,8 @@ export default defineComponent<KottiAvatar.PropsInternal>({
 		right: 0;
 		bottom: 0;
 		left: 0;
-
 		width: 100%;
-		height: 100%;
+		aspect-ratio: 1/1;
 	}
 
 	&__fallback {
@@ -130,18 +130,21 @@ export default defineComponent<KottiAvatar.PropsInternal>({
 	&--is-size-small {
 		width: 1.6rem;
 		height: 1.6rem;
+		max-height: 1.6rem;
 		font-size: 1.2rem;
 	}
 
 	&--is-size-medium {
 		width: 2.4rem;
 		height: 2.4rem;
+		max-height: 2.4rem;
 		font-size: 1.8rem;
 	}
 
 	&--is-size-large {
 		width: 3.2rem;
 		height: 3.2rem;
+		max-height: 3.2rem;
 		font-size: 2.4rem;
 	}
 }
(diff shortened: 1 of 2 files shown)


packages/kotti-ui/source/kotti-avatar/KtAvatar.vue
```diff
@@ -129,19 +128,22 @@ export default defineComponent<KottiAvatar.PropsInternal>({
 
 	&--is-size-small {
 		width: 1.6rem;
-		height: 1.6rem;
+		max-height: 1.6rem;
+		aspect-ratio: 1/1;
```
santiagoballadares, line 132: Why repeat this line (`aspect-ratio: 1/1;`) in every size modifier and not move it under `.kt-avatar {`?

packages/kotti-ui/source/kotti-avatar/KtAvatar.vue
```diff
@@ -129,19 +128,22 @@ export default defineComponent<KottiAvatar.PropsInternal>({
...
 		font-size: 1.2rem;
 	}
 
 	&--is-size-medium {
 		width: 2.4rem;
-		height: 2.4rem;
+		max-height: 2.4rem;
+		aspect-ratio: 1/1;
```
santiagoballadares, line 139: Same as above

packages/kotti-ui/source/kotti-avatar/KtAvatar.vue
```diff
@@ -129,19 +128,22 @@ export default defineComponent<KottiAvatar.PropsInternal>({
...
 		font-size: 1.8rem;
 	}
 
 	&--is-size-large {
 		width: 3.2rem;
-		height: 3.2rem;
+		max-height: 3.2rem;
+		aspect-ratio: 1/1;
```
santiagoballadares, line 146: Same as above

packages/kotti-ui/source/kotti-avatar/KtAvatar.vue
```diff
@@ -129,19 +128,22 @@ export default defineComponent<KottiAvatar.PropsInternal>({
 
 	&--is-size-small {
 		width: 1.6rem;
-		height: 1.6rem;
+		max-height: 1.6rem;
```
santiagoballadares, line 131: Why remove `width`? with long content, it'll shrink  the image. Maybe a `min-width` is missing?

santiagoballadares:
@Juli0GT not sure what design expects here.

Should the image shrink with long content? If so, should a `min-width` be introduced?

If not, maybe it's easier to use `flex: 0 0 auto` in `<div class="kt-avatar">` since it's a flex item. This way, `<div class="kt-avatar">` keeps it's size, which is dictated by the `width`s in the size modifiers.

Long content (image disappears if longer):
Screen Shot 2023-04-11 at 23 31 48

packages/kotti-ui/source/kotti-avatar/KtAvatar.vue
Juli0GT: `width` is not removed, I just changed `height` for `max-height` and add the `aspect-ratio: 1/1`

packages/kotti-ui/source/kotti-avatar/KtAvatar.vue
Juli0GT: Same as above

packages/kotti-ui/source/kotti-avatar/KtAvatar.vue
Juli0GT: Same as above.

carsoli (approved):

carsoli:
Still need to test 

@Juli0GT you need to run fix:stylelint on your machine and recommit/push

santiagoballadares (approved):

PR #740, opened 2023-04-21
fix(chrome): make animation of skeletons work for chrome as well

Isokaeder:
I suddenly realized that the skeletons were not doing their shiny animation for my chromium based browser (brave)

This fixes the animation, I made sure that the animation also still works for both Safari and Firefox.


FlorianWendelborn (approved):

carsoli (approved):

carsoli:
Tested on chrome, works

PR #741, opened 2023-04-24
feat(KtButton): Add helptext prop for icon only buttons

Juli0GT:
Support helpText prop exclusively for buttons with an undefined `label` prop (=buttons with icons only or nothing at all)

### DIFF
--- a/packages/kotti-ui/source/kotti-button/KtButton.vue
+++ b/packages/kotti-ui/source/kotti-button/KtButton.vue
@@ -1,5 +1,5 @@
 <template>
-	<div>
+	<div style="display: contents">
 		<button
 			ref="triggerRef"
 			:class="mainClasses"
--- a/packages/kotti-ui/source/kotti-button/KtButton.vue
+++ b/packages/kotti-ui/source/kotti-button/KtButton.vue
@@ -1,22 +1,29 @@
 <template>
-	<button
-		:class="mainClasses"
-		role="button"
-		:type="isSubmit ? 'submit' : 'button'"
-		@click="handleClick"
-	>
-		<i v-if="isLoading" class="kt-circle-loading" />
-		<i v-else-if="icon !== null" class="yoco" v-text="icon" />
-		<span v-if="hasSlot">
-			<slot />
-		</span>
-		<span v-else-if="label !== null" v-text="label" />
-	</button>
+	<div>
+		<button
+			ref="triggerRef"
+			:class="mainClasses"
+			role="button"
+			:type="isSubmit ? 'submit' : 'button'"
+			@click="handleClick"
+		>
+			<i v-if="isLoading" class="kt-circle-loading" />
+			<i v-else-if="icon !== null" class="yoco" v-text="icon" />
+			<span v-if="hasSlot">
+				<slot />
+			</span>
+			<span v-else-if="label !== null" v-text="label" />
+		</button>
+		<div v-if="showHelpText" ref="contentRef" v-text="helpText" />
+	</div>
 </template>
 
 <script lang="ts">
-import { computed, defineComponent } from '@vue/composition-api'
+import { useTippy } from '@3yourmind/vue-use-tippy'
+import { computed, defineComponent, onMounted, ref } from '@vue/composition-api'
+import { roundArrow } from 'tippy.js'
 
+import { TIPPY_LIGHT_BORDER_ARROW_HEIGHT } from '../constants'
 import { makeProps } from '../make-props'
 
 import { KottiButton } from './types'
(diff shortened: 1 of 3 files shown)


packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -1,21 +1,27 @@
...
-		<i v-else-if="icon !== null" class="yoco" v-text="icon" />
-		<span v-if="hasSlot">
-			<slot />
-		</span>
-		<span v-else-if="label !== null" v-text="label" />
-	</button>
+	<div>
+		<button
+			:class="mainClasses"
+			ref="triggerRef"
```
santiagoballadares, line 5:
```suggestion
			ref="triggerRef"
			:class="mainClasses"
```

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -1,21 +1,27 @@
...
+			<span v-else-if="label !== null" v-text="label" />
+		</button>
+		<div v-if="showHelpText" ref="contentRef" v-text="helpText" />
+	</div>
 </template>
 
 <script lang="ts">
-import { computed, defineComponent } from '@vue/composition-api'
+import { computed, defineComponent, onMounted, ref } from '@vue/composition-api'
+import { useTippy } from '@3yourmind/vue-use-tippy'
```
santiagoballadares, line 23:
```suggestion
import { useTippy } from '@3yourmind/vue-use-tippy'
import { computed, defineComponent, onMounted, ref } from '@vue/composition-api'
```

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -25,9 +31,46 @@ export default defineComponent<KottiButton.PropsInternal>({
...
 
+		const isIconButton = computed(
+			() => props.icon !== null && props.label === null,
+		)
+
+		const showHelpText = computed(
+			() => isIconButton.value && props.helpText !== null,
+		)
+
+		useTippy(
```
santiagoballadares, line 47: Instead of setting the trigger to manual (lines 58 to 60), I would do:
```suggestion
		if (showHelpText.value)
			useTippy(
```
This way, Tippy isn't set needlessly for buttons without help text.
And lines 58 to 60 will not longer be needed.

packages/kotti-ui/source/kotti-button/KtButton.vue
carsoli: @Juli0GT we need to reconfigure the liner on your machine :D

Locally please run `yarn run fix` for now

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -1,21 +1,27 @@
...
-		@click="handleClick"
-	>
-		<i v-if="isLoading" class="kt-circle-loading" />
-		<i v-else-if="icon !== null" class="yoco" v-text="icon" />
-		<span v-if="hasSlot">
-			<slot />
-		</span>
-		<span v-else-if="label !== null" v-text="label" />
-	</button>
+	<div>
```
carsoli, line 2: can you add `style="display:contents"`
to avoid any addition of unintentional styles with this div

packages/kotti-ui/source/kotti-button/KtButton.vue
carsoli: I'm not sure this condition would retrigger if the showHelpText changes during the lifecycle of the component though @santiagoballadares - therefore if it (helpText)'s set to null on mounting then gets a value from the user-app side, this might not work at all

hooks shouldn't be wrapped in if conditions on setup level as far as I recall from best practises guidelines as well

santiagoballadares: Agreed @carsoli , didn't consider that this would change after mount

packages/kotti-ui/source/kotti-button/types.ts
```diff
@@ -18,17 +18,32 @@ export namespace KottiButton {
...
 		isBlock: z.boolean().default(false),
 		isLoading: z.boolean().default(false),
 		isMultiline: z.boolean().default(false),
 		isSubmit: z.boolean().default(false),
-		label: z.string().nullable().default(null),
 		size: sizeSchema.default(Size.MEDIUM),
 		type: typeSchema.default(Type.DEFAULT),
 	})
 
+	export const iconButtonSchema = sharedSchema.extend({
```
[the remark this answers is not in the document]
Juli0GT: @carsoli I tried to use `parseSafe` but didn't work. I think is because in the schema definition `label` and `helpText` are defined as `nullable` with `null` as default value, thats why. Even if any of them are passed they are still defined as `null` so the `parseSafe` resolves in success.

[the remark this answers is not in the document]
Juli0GT: I will remove the unused code.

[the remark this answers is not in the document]
carsoli: That's odd.. if you pass the value of a nullable-by-default prop to the component, it shouldn't fallback to null..

In the case of iconButtonSchema, we defined `label` as `z.null()` not `nullable().default(null)` because we actually don't want to pass it

However, i also didn't think through the "implications" of this on Vue

because I don't think Vue handles such a case, so I'm not sure how it'd have internally parsed a label in the case of iconButtonSchema, and I think this is what you meant..

I guess it's much easier to do it with JS - let's not overcomplicate validation of props if there's an available solution

Thank you for removing it and resolving..

packages/kotti-ui/source/kotti-button/types.ts
```diff
@@ -18,17 +18,21 @@ export namespace KottiButton {
...
 		isBlock: z.boolean().default(false),
 		isLoading: z.boolean().default(false),
 		isMultiline: z.boolean().default(false),
 		isSubmit: z.boolean().default(false),
-		label: z.string().nullable().default(null),
 		size: sizeSchema.default(Size.MEDIUM),
 		type: typeSchema.default(Type.DEFAULT),
 	})
 
+	export const propsSchema = sharedSchema.extend({
```
carsoli, line 30: you don't need the split anymore

it can all go in sharedSchema now

it was like this because we were going to have two schemas extend separately from shared but since it's now one schema, please merge it back

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -25,9 +32,46 @@ export default defineComponent<KottiButton.PropsInternal>({
 	name: 'KtButton',
 	props: makeProps(KottiButton.propsSchema),
 	setup(props, { emit, slots }) {
+		const contentRef = ref<Element | null>(null)
+		const triggerRef = ref<Element | null>(null)
+
 		const hasSlot = computed(() => Boolean(slots.default))
```
carsoli, line 38: @Juli0GT i just realised that a button can accept a slot (which can accept text or icons or anything) and I'm not sure what the support is like for those buttons

because they can have a label through the slot, not through the label prop

carsoli (approved):

PR #743, opened 2023-04-27
feat(KtButton): Add iconPosition prop defaulted to left

Juli0GT:
Add iconPosition prop, to allow users to add icon to the right of the label.

Default value is left.

### DIFF
--- a/packages/documentation/pages/usage/components/button.vue
+++ b/packages/documentation/pages/usage/components/button.vue
@@ -105,8 +105,8 @@
     ## Loading
 
     <div class="element-example">
-    	<KtButton isLoading class="mr-16px">Loading button</KtButton>
-    	<KtButton type="primary" class="mr-16px" isLoading>Loading button</KtButton>
+    	<KtButton isLoading class="mr-16px" icon="edit">Loading button</KtButton>
+    	<KtButton type="primary" class="mr-16px" isLoading icon="edit" iconPosition="right">Loading button</KtButton>
     	<KtButton type="secondary" class="mr-16px" isLoading>Loading</KtButton>
     	<KtButton type="danger" class="mr-16px" isLoading>Loading</KtButton>
     </div>
--- a/packages/documentation/pages/usage/components/button.vue
+++ b/packages/documentation/pages/usage/components/button.vue
@@ -65,18 +65,21 @@
     <div class="element-example">
     	<KtButton type="primary" class="mr-4">Edit button</KtButton>
     	<KtButton type="primary" icon="edit" label="Edit Button" class="mr-4" />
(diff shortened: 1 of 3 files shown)


packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -8,11 +8,12 @@
 			@click="handleClick"
 		>
 			<i v-if="isLoading" class="kt-circle-loading" />
-			<i v-else-if="icon !== null" class="yoco" v-text="icon" />
+			<i v-else-if="hasIconLeft" class="yoco" v-text="icon" />
 			<span v-if="hasSlot">
 				<slot />
 			</span>
 			<span v-else-if="label !== null" v-text="label" />
+			<i v-if="hasIconRight && !isLoading" class="yoco" v-text="icon" />
```
carsoli, line 16: technically for the left Icon we remove the icon and replace it with the loading icon (loading comes on left)

but you don't need the `!isLoading` for the right Icon

packages/documentation/pages/usage/components/button.vue
```diff
@@ -65,18 +65,21 @@
...
     	<KtButton type="primary" class="mr-4">Edit button</KtButton>
     	<KtButton type="primary" icon="edit" label="Edit Button" class="mr-4" />
+    	<KtButton type="primary" icon="edit" label="Right Icon Button" iconPosition="right"/>
     	<KtButton type="primary" icon="edit" helpText="This is an icon button"/>
     </div>
 
     * **Label only:** Used in most cases.
     * **Icon and label:** Use when you need to catch the user's attention.
     * **Icon only:** Use when you have limited space, such as when the page needs to fit on a mobile device, and a single icon is enough to convey the meaning,
+    * **iconPosition** prop can be used to place icon to the right of the label. Is left by defaul.
```
santiagoballadares, line 75:
```suggestion
    * **iconPosition** prop can be used to place icon to the right of the label. Is left by default.
```

packages/kotti-ui/source/kotti-button/KtButton.vue
santiagoballadares: @carsoli is it agreed that loading is rendered always on the left side of the button regardless of the icon position?
or should it be placed in the same position as the icon?

Juli0GT: Good point @santiagoballadares probably this is a question for Alberto.

Juli0GT: @santiagoballadares just asked Alberto he agreed to keep it on the left.

santiagoballadares (approved):

PR #744, opened 2023-05-02
fix(KtButtonGroup): styles

carsoli:
due to the introduce of a div around each kt-button, the selectors .. have to be changed on kt-button


santiagoballadares (approved):
FT done

PR #745, opened 2023-05-03
chore(KtInput): delete component and usages

carsoli:


packages/documentation/pages/usage/components/table.vue
```diff
@@ -906,14 +906,14 @@ _Notes_:
...
-				<KtInput
+				<input
 					label="Drag From: "
 					type="number"
 					min="0"
 					:max="columns.length-1"
 					v-model="fromIndex"
 				/>
-				<KtInput
+				<input
```
carsoli, line 916: I have another PR that replaces this with KtFieldNumber in the making but the KtFieldNumber errors on mounting because of values that don't match min and max and this was the easiest replacement for the purpose of KtInput removal

Isokaeder: LGTM

Isokaeder (approved):

PR #747, opened 2023-05-04
refact(kt-comment): Introduce CommentHeader

santiagoballadares:
Introduce CommentHeader.vue and use in KtComment.vue and CommentReply.vue

Includes props to show Internal thread and Edited labels


packages/kotti-ui/source/kotti-comment/components/CommentHeader.vue
```diff
@@ -0,0 +1,90 @@
...
+		}
+	},
+})
+</script>
+
+<style lang="scss" scoped>
+.kt-comment__comment-header {
+	display: flex;
+	justify-content: space-between;
+	width: 100%;
```
carsoli, line 51: do you know why you needed this?

if the parent container is a flex, the children nodes should fit right in

packages/kotti-ui/source/kotti-comment/types.ts
```diff
@@ -58,15 +58,30 @@ export namespace KottiComment {
 		}
 	}
 
+	export type UserData = Pick<PropsInternal, 'userName' | 'userId'>
+
+	export namespace CommentHeader {
+		export const schema = commentSchema
```
carsoli, line 64: why would this be on the schema of the CommentHeader @santiagoballadares 

in any case, this needs to be passed on a comment schema level

so the `isInternalThread` needs to be a prop of the commentSchema

not extended on the schema of the commentHeader alone

same for `isModified`

carsoli (approved):

PR #748, opened 2023-05-04
KtComment: replace KtPopover in CommentActions

carsoli:
As per the new design of KtComment, the popover for delete and edit actions is removed, and instead the action buttons are spread next to the reply..

All the actions are "KtButtons" of type 'text'

The Reply button only shows for the root comment because KtComment doesn't support nested comments beyond depth = 1 
> KtComment only supports one level of replies and therefore the reply button doesn't make sense on the reply comment


santiagoballadares (approved):

santiagoballadares:
LGTM, also FTed

PR #749, opened 2023-05-05
Kt button toggleable type

Juli0GT:
Implements KtButton toggle:

Toggle button is only possible for types: default or text.

Should follow Figma designs: https://www.figma.com/file/0yFVivSWXgFf2ddEF92zkf/Kotti-Design-System?type=design&node-id=4528-10771&t=aooA0sTF2ETWwPx8-0

It adds one new props:
`toggleStatus`

### DIFF
--- a/packages/documentation/pages/usage/components/button.vue
+++ b/packages/documentation/pages/usage/components/button.vue
@@ -81,6 +81,18 @@
     <KtButton type="primary" icon="edit" helpText="This is an icon button"/>
     <KtButton type="primary" icon="edit" label="Right Icon Button" iconPosition="right"/>
     ```
+
+    ## `toggleStatus`
+
+    *  For buttons that are **toggleable**, and can have two different status: "ON" or "OFF"
+    *  This prop is only valid for buttons of type **"default"** or type **"text"**.
+
+     <div class="element-example white">
+    	<KtButton type="default" :icon="toggleDefaultIcon" :toggleStatus="toggleDefaultStatus" @update:toggleStatus="(event) => onToggleDefaultClick(event)" class="mr-4">{{toggleDefaultLabel}}</KtButton>
+    	<KtButton type="text" :icon="toggleTextIcon" :toggleStatus="toggleTextStatus" @update:toggleStatus="(event) => onToggleTextClick(event)" :label="toggleTextLabel" class="mr-4" />
+    </div>
+    </div>
+
     ## `isMultiline`/`isBlock`
 
     For handling long text, we can use the `isMultiline` and `isBlock` properties.
@@ -157,7 +169,8 @@
 
 <script lang="ts">
 import { KtButton } from '@3yourmind/kotti-ui'
-import { defineComponent } from '@vue/composition-api'
+import { Kotti } from '@3yourmind/kotti-ui'
+import { computed, defineComponent, ref } from '@vue/composition-api'
 
 import ComponentInfo from '~/components/ComponentInfo.vue'
 import ShowCase from '~/components/ShowCase.vue'
@@ -169,9 +182,44 @@ export default defineComponent({
 		ShowCase,
 	},
 	setup() {
+		const toggleDefaultStatus = ref<Kotti.Button.ToggleStatus>(
+			Kotti.Button.ToggleStatus.OFF,
+		)
+		const toggleTextStatus = ref<Kotti.Button.ToggleStatus>(
+			Kotti.Button.ToggleStatus.OFF,
+		)
+
 		return {
 			alert: (value: string) => window.alert(value),
 			component: KtButton,
+			onToggleDefaultClick: (status: Kotti.Button.ToggleStatus) => {
+				toggleDefaultStatus.value = status
+			},
+			onToggleTextClick: (status: Kotti.Button.ToggleStatus) => {
+				toggleTextStatus.value = status
+			},
+			toggleDefaultIcon: computed(() =>
+				toggleDefaultStatus.value === Kotti.Button.ToggleStatus.ON
+					? 'check'
+					: 'close',
+			),
+			toggleDefaultLabel: computed(() =>
+				toggleDefaultStatus.value === Kotti.Button.ToggleStatus.ON
+					? 'DEFAULT ON'
+					: 'DEFAULT OFF',
+			),
+			toggleDefaultStatus,
+			toggleTextIcon: computed(() =>
+				toggleTextStatus.value === Kotti.Button.ToggleStatus.ON
+					? 'check'
+					: 'close',
+			),
+			toggleTextLabel: computed(() =>
+				toggleTextStatus.value === Kotti.Button.ToggleStatus.ON
+					? 'TEXT ON'
+					: 'TEXT OFF',
+			),
+			toggleTextStatus,
 		}
 	},
 })
--- a/packages/kotti-ui/source/kotti-button/KtButton.vue
+++ b/packages/kotti-ui/source/kotti-button/KtButton.vue
@@ -48,6 +48,8 @@ export default defineComponent<KottiButton.PropsInternal>({
 			() => isIconButton.value && props.helpText !== null,
 		)
 
+		const isToggle = computed(() => props.toggleStatus !== null)
+
 		useTippy(
 			triggerRef,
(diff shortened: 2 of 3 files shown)


packages/documentation/pages/usage/components/button.vue
```diff
@@ -81,6 +81,18 @@
     <KtButton type="primary" icon="edit" helpText="This is an icon button"/>
     <KtButton type="primary" icon="edit" label="Right Icon Button" iconPosition="right"/>
     ```
+
+    ## `isToggle`
+
+    *  For buttons that are **toggleable**, and can have two different status: "ON" or "OF"
```
carsoli, line 87: or OFF*

packages/documentation/pages/usage/components/button.vue
```diff
@@ -169,9 +182,36 @@ export default defineComponent({
...
+			KottiButton.ToggleStatus.OFF,
+		)
+		const toggleTextStatus = ref<KottiButton.ToggleStatus>(
+			KottiButton.ToggleStatus.OFF,
+		)
+
 		return {
 			alert: (value: string) => window.alert(value),
 			component: KtButton,
+			onToggleDefaultClick: ({
```
carsoli, line 195: the payload that you emit feels weird

packages/documentation/pages/usage/components/button.vue
```diff
@@ -81,6 +81,18 @@
...
     ```
+
+    ## `isToggle`
+
+    *  For buttons that are **toggleable**, and can have two different status: "ON" or "OF"
+    *  This prop is only valid for buttons of type **"default"** or type **"text"**.
+    *  Apart from **isToggle** prop it needs **toggleStatus** prop that can be either **on** or **of**.
+
+     <div class="element-example white">
+    	<KtButton type="default" :isToggle="true" :toggleStatus="toggleDefaultStatus" @click="(event) => onToggleDefaultClick(event)" class="mr-4">Default Toggle button</KtButton>
```
carsoli, line 92: @Juli0GT I'd much rather you have a special event for toggling not just the click

basically

if the prop for toggling on and off is called x , you have an event called update:x

with vue3 this will allow us to do v-model:x so that whoever wants to use a toggle button in the user app can just v-model:x easily and doesn't need an explicit handler

the click can remain as is but not be used

carsoli: my 2nd opinion from an API perspective (usage perspective)

in order to achieve the functionality you want, you ended up having to add 2 props (isToggle and toggleStatus) 

instead

the API just exposes `toggleStatus` and emits `update:toggleStatus`

IF the user defines toggleStatus then it's a toggle button

if the user doesn't define toggleStatus then it's not

so instead of having two props for the feature you have ONE "implication" prop (implication because the status already implies that this is a toggle button)

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -69,11 +69,30 @@ export default defineComponent<KottiButton.PropsInternal>({
 					'KtButton: Guideline Uncompliance: attempted to use helpText with a button that already has a label.',
 				)
 			}
+
+			if (
+				props.isToggle &&
```
carsoli, line 74: if you remove isToggle from props then here you would check

props.toggleStatus !== null

carsoli: you can even have a computed called isToggle 

so isToggle = computed(() => (props.toggleStatus !== null))

packages/kotti-ui/source/kotti-button/types.ts
```diff
@@ -32,8 +39,10 @@ export namespace KottiButton {
 		isLoading: z.boolean().default(false),
 		isMultiline: z.boolean().default(false),
 		isSubmit: z.boolean().default(false),
+		isToggle: z.boolean().default(false),
 		label: z.string().nullable().default(null),
 		size: sizeSchema.default(Size.MEDIUM),
+		toggleStatus: toggleStatusSchema.default(ToggleStatus.OFF),
```
carsoli, line 45: make the default null in that case

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -69,11 +71,29 @@ export default defineComponent<KottiButton.PropsInternal>({
...
-			handleClick: (event) => emit('click', event),
+			handleClick: (event) =>
+				isToggle.value
+					? emit(
+							'update:toggleStatus',
+							props.toggleStatus === KottiButton.ToggleStatus.OFF
+								? KottiButton.ToggleStatus.ON
+								: KottiButton.ToggleStatus.OFF,
+					  )
+					: emit('click', event),
```
carsoli, line 96: @Juli0GT I wouldn't make this a ternary because we still want to emit the click in case the user app needs it

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -240,6 +262,22 @@ export default defineComponent<KottiButton.PropsInternal>({
 				border-bottom-color: var(--button-main-color-dark);
 				border-left-color: var(--button-main-color-dark);
 			}
+			&.kt-button--is-toggle-on {
+				color: var(--text-04) !important;
```
carsoli, line 266: can you actually try without important and see if it works

i think you have higher specifity here, there's no way this doesn't apply and take precedence

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -69,11 +71,30 @@ export default defineComponent<KottiButton.PropsInternal>({
 					'KtButton: Guideline Uncompliance: attempted to use helpText with a button that already has a label.',
 				)
 			}
+
+			if (
+				isToggle.value &&
+				props.type !== KottiButton.Type.DEFAULT &&
+				props.type !== KottiButton.Type.TEXT
```
santiagoballadares, line 78:
```suggestion
				![KottiButton.Type.DEFAULT, KottiButton.Type.TEXT].includes(props.type)
```

packages/kotti-ui/source/kotti-button/types.ts
```diff
@@ -24,6 +24,13 @@ export namespace KottiButton {
 	}
 	export const iconPosition = z.nativeEnum(IconPosition)
 
+	export enum ToggleStatus {
+		ON = 'on',
+		OFF = 'off',
```
santiagoballadares, line 29:
```suggestion
		ON = 'ON',
		OFF = 'OFF',
```

Juli0GT: the other enums defined on this file are in lowercase, I'm fine having them either lowercased or uppercased but I wouldn't mix.

carsoli (dismissed):

packages/documentation/pages/usage/components/button.vue
```diff
@@ -157,7 +169,8 @@
 
 <script lang="ts">
 import { KtButton } from '@3yourmind/kotti-ui'
-import { defineComponent } from '@vue/composition-api'
+import { KottiButton } from '@3yourmind/kotti-ui/dist/esm/kotti-button/types'
```
santiagoballadares, line 172: Why importing from '@3yourmind/kotti-ui/dist/esm/kotti-button/types'?
shouldn't it be?
```
import { Kotti } from '@3yourmind/kotti-ui'
```
and then usage:
```
Kotti.Button.ToggleStatus.OFF
```

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -71,11 +73,29 @@ export default defineComponent<KottiButton.PropsInternal>({
...
+				throw new Error(
+					'KtButton: Guideline Uncompliance: attempted to use toggleStatus with a button of type different than "DEFAULT" or "TEXT"',
+				)
+			}
 		})
 
 		return {
 			contentRef,
-			handleClick: (event: Event) => emit('click', event),
+			handleClick: (event) => {
```
santiagoballadares, line 89: Why is the event type removed?

santiagoballadares (approved):

PR #750, opened 2023-05-05
KtComment: refact comments components

santiagoballadares:
Refact comments components

- Introduce CommentEntry
- Remove CommentReply
- Update KtComment docs


packages/kotti-ui/source/kotti-comment/components/CommentEntry.vue
```diff
@@ -0,0 +1,86 @@
...
+	props: makeProps(KottiComment.Entry.schema),
+	setup(props, { emit }) {
+		const isEditing = ref(false)
+
+		const isReply = computed(() => props.type === KottiComment.EntryType.REPLY)
+
+		return {
+			headerProps: computed(() => ({
+				...pick(props, ['createdAt', 'isModified', 'user']),
+				isInternalThread: !isReply.value ? props.isInternalThread : false,
```
carsoli, line 56: a reply inherits from its parent? why would you return false if it's not a reply

carsoli (approved):

packages/kotti-ui/source/kotti-comment/components/CommentEntry.vue
santiagoballadares: Good catch, i don't need to do this.
So, CommentEntry renders a comment (either parent (first comment) or reply). According to design, only parent comments show `Internal thread` label in the header (the same for the `Reply` button).
Replies are already typed and they don't include isInternalThread, so it'll default to false and not show the `Internal thread` label in the header.
It's updated now.

PR #753, opened 2023-05-10
refact(kt-comment): Refact CommentInlineEdit and KtCommentInput

santiagoballadares:
Refact CommentInlineEdit.vue and KtCommentInput.vue

- Add isInternal and allowInternal flags
- Refact comment docs

### DIFF
--- a/packages/kotti-ui/source/kotti-comment/KtComment.vue
+++ b/packages/kotti-ui/source/kotti-comment/KtComment.vue
@@ -2,31 +2,32 @@
 	<div :class="classes">
 		<CommentEntry
 			v-bind="commentProps"
-			:type="EntryType.POST"
 			@delete="onDelete"
 			@edit="onEdit"
 			@reply="onReply"
 		/>
-		<div class="kt-comment__thread">
+		<div v-if="showCommentThread" class="kt-comment__thread">
 			<CommentEntry
 				v-for="reply in replies"
 				:key="reply.id"
 				v-bind="reply"
 				:dangerouslyOverrideParser="dangerouslyOverrideParser"
+				isReply
 				:parentId="id"
 				:postEscapeParser="postEscapeParser"
-				:type="EntryType.REPLY"
+				:tabIndex="tabIndex"
 				@delete="onDelete"
 				@edit="onEdit"
 			/>
 			<KtCommentInput
 				v-if="userToReply"
-				isInline
+				v-bind="{ isInternal, placeholder, tabIndex, userAvatar }"
+				autofocus
+				isReply
 				:parentId="id"
-				:placeholder="placeholder"
 				:replyToUserId="userToReply.id"
-				:userAvatar="user.avatar"
 				@add="onAdd"
+				@cancel="onCancel"
 			/>
 		</div>
 	</div>
@@ -58,14 +59,17 @@ export default defineComponent<KottiComment.PropsInternal>({
 		return {
 			classes: computed(() => ({
 				'kt-comment': true,
-				'kt-comment--is-internal-thread': props.isInternalThread,
+				'kt-comment--is-internal': props.isInternal,
 			})),
-			commentProps: computed(() => omit(props, 'replies')),
-			EntryType: KottiComment.EntryType,
+			commentProps: computed<KottiComment.Entry.PropsInternal>(() => ({
+				...omit(props, 'replies'),
+				isReply: false,
+			})),
 			onAdd: (payload: KottiComment.Events.Add) => {
 				userToReply.value = null
 				emit('add', payload)
 			},
+			onCancel: () => (userToReply.value = null),
 			onDelete: (payload: KottiComment.Events.Delete) =>
 				emit('delete', payload),
 			onEdit: (payload: KottiComment.Events.Edit) => emit('edit', payload),
@@ -74,6 +78,9 @@ export default defineComponent<KottiComment.PropsInternal>({
 				userToReply.value
 					? [translations.value.replyToLabel, userToReply.value.name].join(' ')
 					: undefined,
+			),
+			showCommentThread: computed(
+				() => props.replies.length > 0 || userToReply.value !== null,
 			),
 			userToReply,
 		}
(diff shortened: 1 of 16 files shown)


packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -21,12 +21,14 @@
 			/>
 			<KtCommentInput
 				v-if="userToReply"
-				isInline
+				autofocus
```
carsoli, line 24: what's that?

santiagoballadares: `autofocus` tells `CommentTextArea` whether it should focus the input on component mount.
When editing or replying, the input should autofocus; but when the page loads, the input inside `KtCommentInput` shouldn't autofocus.

packages/kotti-ui/source/kotti-comment/KtCommentInput.vue
```diff
@@ -1,140 +1,69 @@
...
 	name: 'KtCommentInput',
 	components: {
-		KtAvatar,
-		KtButton,
+		CommentTextArea,
 	},
 	props: makeProps(KottiCommentInput.propsSchema),
 	setup(props, { emit }) {
-		const translations = useTranslationNamespace('KtComment')
+		const isInternal = ref<KottiComment.PropsInternal['isInternal']>(false)
```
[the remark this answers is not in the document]
santiagoballadares: When adding comments or replies there's not values for message or isInternal. So, `KtCommentInput.vue` keeps local variables and sends the values in the @add payload only if the Confirm btn is clicked.

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -2,31 +2,35 @@
...
-				:type="EntryType.REPLY"
+				:tabIndex="tabIndex"
 				@delete="onDelete"
 				@edit="onEdit"
 			/>
 			<KtCommentInput
 				v-if="userToReply"
-				isInline
+				autofocus
+				:isInternal="isInternal"
```
carsoli, line 25: not critical but you could
`v-bind="{ isInternal, tabIndex, userAvatar }"`

packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -75,6 +82,9 @@ export default defineComponent<KottiComment.PropsInternal>({
 					? [translations.value.replyToLabel, userToReply.value.name].join(' ')
 					: undefined,
 			),
+			showCommentThread: computed(
+				() => props.replies.length > 0 || Boolean(userToReply.value),
```
carsoli, line 86: don't prefer Boolean() with objects
because Boolean({}) gives true

so in general i'd go for userReply.value !== null

packages/kotti-ui/source/kotti-comment/KtCommentInput.vue
```diff
@@ -1,140 +1,78 @@
...
-				:label="replyButtonText"
-				type="text"
-				@click="handleSubmitClick"
-			/>
-		</div>
+	<div class="kt-comment-input">
+		<KtAvatar size="sm" :src="userAvatar" />
+		<CommentTextArea
+			v-model="_message"
+			:allowInternal="allowInternal"
```
carsoli, line 6: mostly can be v-bound

v-bind={allowInternal, autoFocus, isReply, placeholder, tabIndex }

carsoli: even the handlers technically because they are following the onEvent convention

packages/kotti-ui/source/kotti-comment/components/CommentInlineEdit.vue
```diff
@@ -1,78 +1,82 @@
...
+		const _message = ref<KottiComment.InlineEdit.PropsInternal['message']>('')
 
 		watch(
 			() => props.isEditing,
-			(newValue) => {
-				if (newValue === true) {
-					inlineValue.value = props.message
+			(isEditing) => {
+				if (isEditing) {
+					_message.value = props.message
```
carsoli, line 44: there's a caveat

if there's a user scenario that would allow the user somehow to trigger the update:isEditing to true twice in a row
then the message will get reset

imagine he clicks post (isEditing is set to true, watcher triggers, so you have your local value set)

then user starts typing

if by any mean, they manage to retrigger update:isEditing then you'd have a bug of resetting your local variable mid typing

and it'd be avoided with a check
```
if(isEditing === props.isEditing) return 
```

carsoli: if it's not possible to end up in this scenario, ignore

packages/kotti-ui/source/kotti-comment/components/CommentTextArea.vue
```diff
@@ -0,0 +1,182 @@
...
+						@click="onToggleInternal"
+					/>
+					<div class="kt-comment__text-area__actions__divider" />
+				</template>
+				<KtButton
+					:disabled="isEmpty"
+					:icon="Yoco.Icon.SEND"
+					:tabIndex="hasFocus ? tabIndex : -1"
+					type="text"
+					@click="onConfirm"
```
carsoli, line 35: you don't need to stop this?

carsoli (approved):

packages/kotti-ui/source/kotti-comment/KtCommentInput.vue
santiagoballadares: Skipping the change for the events for now, otherwise I'll have to either do: 
```
v-on="{
	cancel: onCancel,
	confirm: onConfirm,
	toggleInternal: onToggleInternal,
}"
```
or rename the events to: cancel, confirm, toggleInternal.
If required, I'll do this change on another PR.

packages/kotti-ui/source/kotti-comment/components/CommentInlineEdit.vue
santiagoballadares: It's not really necessary bc the only way to trigger it is via the `Edit` button inside `Actions`, but it's hidden during editing.

Nevertheless, I applied the changed to make it explicit the the internal value shouldn't be mutated during editing.

Isokaeder:
lgtm

Isokaeder (approved):

PR #759, opened 2023-05-26
refact(kt-button): Refact helpText in KtButton

santiagoballadares:
Refact helpText feature in KtButton to fix issues caused by the introduction on a `div` wrapper around `<button />`. Also revert 2 fixes to styles and inherited properties attempting to fix some of the issues caused by the `div` wrapper.

Effectively: 
- Refact: e7e999b8a
- Revert: 6e6d2a1 and 55a46b5

Functional Test required:
1. In usage of KtButton, check that:
    a. Under `Icon` section, `helpText prop can be passed to Icon only buttons that is displayed on button hover.` woks as expected.
    b. Under `Button Group` section, buttons are rendered correctly, i.e. the borders and corners are rounded only on the external sides of the first and last buttons in the group.

2. In the usage of any other component that has a button disabled (e.g. in `KtComment`, the input text area should have the `send` icon button disabled if there's not text), buttons are disabled.

### DIFF
--- a/packages/kotti-ui/source/kotti-button/KtButton.vue
+++ b/packages/kotti-ui/source/kotti-button/KtButton.vue
@@ -1,23 +1,20 @@
 <template>
-	<div class="kt-button-wrapper" style="display: contents">
-		<button
-			ref="triggerRef"
-			v-bind="$attrs"
-			:class="mainClasses"
-			role="button"
-			:type="isSubmit ? 'submit' : 'button'"
-			@click="handleClick"
-		>
-			<i v-if="isLoading" class="kt-circle-loading" />
-			<i v-if="hasIconLeft" class="yoco" v-text="icon" />
-			<span v-if="hasSlot">
-				<slot />
-			</span>
-			<span v-else-if="label !== null" v-text="label" />
-			<i v-if="hasIconRight" class="yoco" v-text="icon" />
-		</button>
-		<div v-if="showHelpText" ref="contentRef" v-text="helpText" />
-	</div>
+	<button
+		ref="helpTextTriggerRef"
+		:class="mainClasses"
+		role="button"
+		:type="isSubmit ? 'submit' : 'button'"
+		@click="handleClick"
+	>
+		<i v-if="isLoading" class="kt-circle-loading" />
+		<i v-if="hasIconLeft" class="yoco" v-text="icon" />
+		<span v-if="hasSlot">
+			<slot />
+		</span>
+		<span v-else-if="label !== null" v-text="label" />
+		<i v-if="hasIconRight" class="yoco" v-text="icon" />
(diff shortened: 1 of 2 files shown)


packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -1,23 +1,20 @@
...
-			</span>
-			<span v-else-if="label !== null" v-text="label" />
-			<i v-if="hasIconRight" class="yoco" v-text="icon" />
-		</button>
-		<div v-if="showHelpText" ref="contentRef" v-text="helpText" />
-	</div>
+	<button
+		ref="helpTextTriggerRef"
+		:class="mainClasses"
+		role="button"
```
carsoli, line 5: why do we need a role button on a button

It is not needed because it is the default value for a button anyways

see https://www.w3.org/TR/2014/REC-html5-20141028/dom.html#aria-role-attribute

or look for this note if the anchor link doesn't work

```
In the majority of cases setting an ARIA [role](https://www.w3.org/TR/2014/REC-html5-20141028/infrastructure.html#attr-aria-role) and/or aria-* attribute that matches the [default implicit ARIA semantics](https://www.w3.org/TR/2014/REC-html5-20141028/infrastructure.html#default-implicit-aria-semantics) is unnecessary and not recommended as these properties are already set by the browser
```

packages/kotti-ui/source/kotti-button/KtButton.vue
```diff
@@ -1,23 +1,20 @@
...
+		:type="isSubmit ? 'submit' : 'button'"
```
carsoli, line 6: from what I understood only if we have "input of type submit" we would need the "role="button"" so maybe that's why the role was there


@FlorianWendelborn do you have an opinion here?

packages/kotti-ui/source/kotti-button/KtButton.vue
carsoli: ahh! good catch

carsoli, line 5:
```suggestion
```

packages/kotti-ui/source/kotti-button/KtButton.vue
FlorianWendelborn: This should definitely stay as-is as it’s a fix for a really ugly and unexpected default HTML behavior where the first button inside a form is automatically assumed to be the submit button. This means that a form like this

```vue
<template>
  <form>
    <input v-model="modelValue" />
    <KtButton label="Delete" @click="deleteMe" />
  </form>
</template>
```

would end-up deleting itself if the user edits the `input` and hits `enter`

See-also: #191 #190 #179

carsoli (approved):

PR #775, opened 2023-06-19
771 KtComment hide reply btn via props

santiagoballadares:
Introduce isReadOnly prop to KtComment to hide all comments actions buttons.

KtCommentInput is not supported.


packages/documentation/pages/usage/components/comment.vue
```diff
@@ -62,6 +64,13 @@
 							label="allowInternal"
 							type="switch"
 						/>
+						<KtFieldToggle
+							formKey="isReadOnly"
+							helpText="Are comment actions buttons visible or not. Only KtComment is supported."
```
carsoli, line 69: what do you mean `only KtComment is supported`

packages/kotti-ui/source/kotti-comment/components/CommentActions.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<div class="kt-comment__actions">
 		<KtButton
-			v-if="!isReply"
+			v-if="!isReadOnly && !isReply"
```
carsoli, line 4: could you not add the `!isReadOnly` on the div instead?

carsoli (approved):

Juli0GT (approved):

Issue #772, opened 2023-06-15
Fix disabled style of `+`/`-` in KtFieldNumber


RueRivoli:
**Describe the bug**
When disabled the `+`/`-` buttons of the `KtFieldNumber` do not have any greyed out style while the class `kt-field-number__button--is-disabled` exists.

**To Reproduce**

1. Go to any KtFieldNumber with a minimum set or (resp with a maximum)
2. When quantity reaches the minimum, the `-` (resp. `+`) button does not appear disabled

**Expected behavior**
disabled `+` and `-` buttons are greyed out, and a cursor: blocker (to confirm with design)

**Screenshots**
<img width="839" alt="Screenshot 2023-06-15 at 11 55 08" src="https://github.com/3YOURMIND/kotti/assets/17473800/7c33250d-4b7c-473b-ab9d-123d89faf036">

**Additional context**
Go to PartDetail of some parts and try to reduce a quantity of any sequence such that the total of sequences from that part is equal to the number of ordered parts. Then the `-` parts is disabled. The greyed out style is applied by yoda not kotti.

PR #776, opened 2023-06-20
fix(KtFieldNumber): when any of the buttons is disabled "cursor: not-allowed" is applied

Juli0GT:
Fix https://github.com/3YOURMIND/kotti/issues/772

### DIFF
--- a/packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
+++ b/packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
@@ -451,6 +451,7 @@ export default defineComponent<KottiFieldNumber.PropsInternal>({
 
 		&--is-disabled {
 			color: var(--text-05);
+			cursor: not-allowed;
 		}
 	}
 


packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
```diff
@@ -411,7 +411,8 @@ export default defineComponent<KottiFieldNumber.PropsInternal>({
 			}
 
 			&--is-disabled {
-				color: var(--text-05);
+				color: vsar(--text-05) !important;
```
carsoli, line 414: typo

santiagoballadares, line 414: @Juli0GT did you FT this?

- Why is `!important` needed for the color? Both, the buttons' icons and the input text already have color `var(--text-05)` when disabled. The problem for the color is mot likely in the usage.

- Adding `cursor: not-allowed;` here doesn't fix the cursor issue. It should be added to this class `kt-field-number__button--is-disabled`.

Also, @carsoli the computed `inputProps` isn't adding the `--is-disabled` class, so this isn't doing anything.

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
```diff
@@ -411,7 +411,8 @@ export default defineComponent<KottiFieldNumber.PropsInternal>({
...
+				cursor: not-allowed;
```
santiagoballadares, line 415: @Juli0GT did you FT this?
Adding `cursor: not-allowed;` here doesn't fix the cursor issue. It should be added to this class `kt-field-number__button--is-disabled`.

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
```diff
@@ -411,7 +411,8 @@ export default defineComponent<KottiFieldNumber.PropsInternal>({
 			}
 
 			&--is-disabled {
```
santiagoballadares, line 413: Also the computed `inputProps` isn't adding the `--is-disabled` class, so this isn't doing anything.
@carsoli fyi

santiagoballadares:
@Juli0GT please squash the commits

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
Juli0GT: This is actually doing something when the whole field is `disabled`, I put the logic in the wrong class.
Now is updated.

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
Juli0GT: The !important is needed because when using the field in a table it overrides the color because some scss aded into `_table.scss`:
```
table {
  	i {
		margin: 0 0.2rem;
		font-size: 1rem !important;
		color: $darkgray-300;
          }
}
```

This code from above overrides the color set in the `--is-disabled` class.

Juli0GT:
> @Juli0GT please squash the commits

@santiagoballadares commits are squashed now. Not sure how it happened to create 3 commits with the same message, as I was amending the commits.

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
santiagoballadares: Yes but that is the usage of `KtFieldNumber`, this component should care about its usage.
The problem is that the table is overriding the css. IMO the code in `_table.scss` should be moved into the corresponding components as scoped. 
@carsoli what do you think?

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
santiagoballadares: Anything inside `kt-field-number__middle__input--is-disabled` is not doing anything because the computed `inputProps` is not adding the `--is-disabled` class.

However, this is out of scope so it can be ignored.

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
carsoli: after another internal discussion, can't be scoped but can cleaned up to remove classes that are no longer used on the user app and to properly target the icons under table-actions only

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
santiagoballadares: Resolving. This unused style is now remove in https://github.com/3YOURMIND/kotti/pull/783

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
santiagoballadares: PR cleaned.
It now contains only the fix for the cursor when hovering on +/- buttons on a disabled KtFieldNumber.

santiagoballadares (approved):

PR #779, opened 2023-06-23
feat(#777): emit blur from KtFieldNumber

RueRivoli:
Emit blur when the focus is out of the whole KtFieldNumber (input + 2 buttons)

### DIFF
--- a/packages/documentation/pages/usage/components/form.vue
+++ b/packages/documentation/pages/usage/components/form.vue
@@ -78,6 +78,12 @@
 					<h2>Personal Details</h2>
 					<KtFormControllerObject formKey="personalDetails">
 						<template #default="{ setValues, values: formObjectValues }">
+							<KtFieldNumber
+								formKey="age"
+								helpText="help for age"
+								label="age"
+								placeholder="how old are you?"
+							/>
 							<KtFieldText
 								formKey="firstName"
 								helpText="help for firstName"
@@ -303,6 +309,7 @@ export default defineComponent({
 		const values = ref({
 			addresses: [{ country: null, houseNumber: null, streetName: null }],
 			personalDetails: {
+				age: 1,
 				firstName: 'John',
 				firstNameFirst: false,
 				fullName: 'Smith,John',
--- a/packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
+++ b/packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
@@ -1,17 +1,21 @@
 <template>
 	<KtField v-bind="{ field }" :helpTextSlot="$slots.helpText">
 		<div
+			ref="wrapperRef"
 			class="kt-field-number"
 			:class="{
 				'kt-field-number--is-hide-change-buttons': hideChangeButtons,
 				'kt-field-number--has-maximum': showMaximum && maximum !== null,
 			}"
+			:tabIndex="-1"
 		>
 			<div
 				v-if="!hideChangeButtons"
+				ref="decrementButtonRef"
 				class="kt-field-number__button"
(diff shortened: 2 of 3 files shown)


RueRivoli:
@carsoli I used your branch `blur-kt-field-number-2`, hope that's fine

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
```diff
@@ -1,18 +1,23 @@
 <template>
 	<KtField v-bind="{ field }" :helpTextSlot="$slots.helpText">
 		<div
+			ref="wrapperRef"
 			class="kt-field-number"
 			:class="{
 				'kt-field-number--is-hide-change-buttons': hideChangeButtons,
 				'kt-field-number--has-maximum': showMaximum && maximum !== null,
 			}"
+			:tabIndex="0"
```
santiagoballadares, line 10: Is this `tabIndex` needed? 
In FT you need to tab twice to get to the `decrement` button and be able to triggered with the `Enter` key.

santiagoballadares:
Did FT and the tab works only in the `form` usage page, not in the `form fields` usage page.
There's one `tabIndex` that seems to be not needed and it makes it needed to tab twice to get to the `decrement` button.
Also, the `decrement` and `increment` buttons don't have any visual indication when the have focus, was this intended? is design aware of this?
It feels weird to have to tab 3 times to get to the `input` (where you can actually type) and 2 times to get out of the field and one is not aware that the buttons inside have the focus.

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
RueRivoli: i removed this tabIndex. When testing with a `KtfieldNumber` not in a first position: after firstname for example. Then we just need to tab once to get in the KtFieldNumber

santiagoballadares (dismissed):
@RueRivoli approved

Please add a toggle to test the blur in the `form fields` page, when enabled just show an alert when the blur happens

packages/documentation/pages/usage/components/form.vue
```diff
@@ -78,6 +78,12 @@
 					<h2>Personal Details</h2>
 					<KtFormControllerObject formKey="personalDetails">
 						<template #default="{ setValues, values: formObjectValues }">
+							<KtFieldNumber
+								formKey="age"
+								helpText="YEARS"
+								label="age"
+								placeholder="quel age as tu?"
```
carsoli, line 85: I'd just fix this to English be compliant with the rest of our docs 

we don't really need helptext, keep the label, and change the placeholder to `how old are you?`

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
```diff
@@ -381,6 +436,9 @@ export default defineComponent<KottiFieldNumber.PropsInternal>({
 		&:not(&--is-disabled):hover {
 			background: var(--ui-02);
 		}
+		&:focus {
```
carsoli, line 439: @RueRivoli did u ask alberto on what we are doing with the buttons?

carsoli (approved):

packages/kotti-ui/source/kotti-field-number/KtFieldNumber.vue
carsoli: @santiagoballadares I requested that he brings this back because it means that the div captures focus in the first place
and so if we remove this, we should also remove the checks from the isEventTarget

carsoli (approved):

santiagoballadares (approved):
FT done, works as expected

Issue #788, opened 2023-07-31
Update KtCard component


RueRivoli:
**Is your feature request related to a problem? Please describe.**
This feature request is necessary to permit the design of Tile (card) view for the catalog. (in addition to the catalog list view)

**Describe the solution you'd like**
- Card height depends on content
- Distinguish 2 type of cards:

> only clickable cards: if the card is meant to be only opened 
> non clickable cards with actions: if the card is meant to contain buttons

- card images needs to have skeletons when loading
- card body needs to have skeletons when loading

- tabulation makes the focus on the card itself or moves forward to the next card or next page component
- shift + tab moves the focus backward to the previous card
- enter: open the card if it's a clickable card or perform the action of the focus element if not

**Describe alternatives you've considered**
/ 

**Additional context**
Find more details on B3-13297

PR #793, opened 2023-08-03
feature(#788) : update KtCard component

RueRivoli:
Updated `KtCard` to make it ready for the Tile View of the catalog.

See different ACs in B3-13297.

Details:
- added props `primaryActionLabel` and `secondaryActionLabel` for the primary and secondary buttons
- I removed the `footer slot` as it complicates things. This is known to design. 
Why? Because we need to know if the card is of type "clickable" or "not-clickable (with buttons actions)".
The hovering and active style differs according to the card type. So, if we permit to add buttons in a footer's slot, we would have to read if there is any one in this slot to know if its a clickable card or not ==> too complicated for a small thing.
So primary and secondary actions are sent via props and the footer slot is removed.
- Header and body slots are still used so any additional skeleton can be added.
 - Emit primaryActionClick/secondaryActionClick when one of the button is clicked.

### DIFF
--- a/packages/documentation/pages/usage/components/card.vue
+++ b/packages/documentation/pages/usage/components/card.vue
@@ -1,7 +1,49 @@
 <template>
 	<div>
 		<ComponentInfo v-bind="{ component }" />
+		<div class="card-view-example">
+			<KtCard
+				v-bind="{ imgPosition }"
+				imgUrl="https://picsum.photos/900/300"
+				primaryActionLabel="Confirm"
+				secondaryActionLabel="Cancel"
+			>
+				<h2 slot="card-header">Lorem Ipsum</h2>
+				<p slot="card-body">
+					Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus
+					consequat nisl at nisl condimentum vehicula.
+				</p>
+			</KtCard>
+		</div>
+		<br />
 
+		<div class="card-view-example-2">
+			<KtCard v-bind="{ imgPosition }" imgUrl="https://picsum.photos/900/300">
+				<h2 slot="card-header">Lorem Ipsum</h2>
+				<p slot="card-body">
+					Moebius ring by pmaegermanis licensed under CC BY 3.0 (Creative
+					Commons https://www.thingiverse.com/thing:1649028
+				</p>
+			</KtCard>
+		</div>
+		<br />
+		<div class="card-view-example-skeleton">
+			<KtCard
+				v-bind="{ imgPosition }"
+				imgUrl="https://picsum.photos/900/300"
+				:isImgLoading="true"
+				:isTextLoading="true"
+				primaryActionLabel="Confirm"
+				secondaryActionLabel="Cancel"
+			>
+				<h2 slot="card-header">Lorem Ipsum</h2>
+				<p slot="card-body">
+					Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus
+					consequat nisl at nisl condimentum vehicula.
+				</p>
+			</KtCard>
+		</div>
+		<br />
 		<div class="element-example">
 			<KtCard>
 				<div slot="card-header">
--- a/packages/kotti-ui/source/kotti-card/KtCard.vue
+++ b/packages/kotti-ui/source/kotti-card/KtCard.vue
@@ -1,16 +1,32 @@
 <template>
-	<div :class="cardClass">
+	<div :class="cardClass" tabindex="0" @click.prevent="handleClick">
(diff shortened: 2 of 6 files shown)


packages/documentation/pages/usage/components/card.vue
```diff
@@ -1,7 +1,49 @@
 <template>
 	<div>
 		<ComponentInfo v-bind="{ component }" />
+		<div class="catalog-tile-view-example">
```
santiagoballadares, line 4: Please rename this to something more generic like:

```suggestion
		<div class="card-view-example">
```

packages/documentation/pages/usage/components/card.vue
```diff
@@ -1,7 +1,49 @@
...
+			<KtCard
+				v-bind="{ imgPosition }"
+				imgUrl="https://picsum.photos/900/300"
+				primaryActionLabel="Command"
```
santiagoballadares, line 8:
```suggestion
				primaryActionLabel="Confirm"
```

packages/documentation/pages/usage/components/card.vue
```diff
@@ -1,7 +1,49 @@
...
+				<h2 slot="card-header">Lorem Ipsum</h2>
+				<p slot="card-body">
+					Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus
+					consequat nisl at nisl condimentum vehicula.
+				</p>
+			</KtCard>
+		</div>
+		<br />
 
+		<div class="catalog-tile-view-example-2">
```
santiagoballadares, line 20: Same as above, make it more generic

packages/kotti-ui/source/kotti-card/KtCard.vue
```diff
@@ -1,36 +1,64 @@
...
+			<slot v-else name="card-header" />
 		</div>
 		<div v-if="$slots['card-body']" class="kt-card__body">
-			<slot name="card-body" />
+			<div v-if="isTextLoading" class="skeleton rectangle sm" />
+			<slot v-else name="card-body" />
 		</div>
-		<div v-if="$slots['card-footer']" class="kt-card__footer">
-			<slot name="card-footer" />
+		<div v-if="$slots['card-footer'] || hasActions" class="kt-card__footer">
```
santiagoballadares, line 15: I though the `footer` slot was removed. Shouldn't it be just?
```suggestion
		<div v-if="hasActions" class="kt-card__footer">
```

packages/kotti-ui/source/kotti-card/types.ts
```diff
@@ -10,6 +10,12 @@ export namespace KottiCard {
 	export const propsSchema = z.object({
 		imgPosition: z.nativeEnum(ImagePosition).default(ImagePosition.TOP),
 		imgUrl: z.string().nullable().default(null),
+		isImgLoading: z.boolean().default(false),
+		isTextLoading: z.boolean().default(false),
```
santiagoballadares, line 14: IS it really necessary to have 2 loadings instead of just 1?

packages/kotti-ui/source/kotti-card/KtCard.vue
RueRivoli: thanks you re right

packages/kotti-ui/source/kotti-card/KtCard.vue
```diff
@@ -1,36 +1,64 @@
...
+			<div v-if="hasActions && !isTextLoading">
```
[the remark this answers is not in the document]
RueRivoli: hmm according to the design we show only 2 skeletons in the card while its loading except if we precise skeletons from yoda directly.

packages/kotti-ui/source/kotti-card/types.ts
RueRivoli: the reason why I created these 2 loading props is because of the requirements on how we want to show next the images/body text:

> For the catalog case, it becomes more important probably that the phases of loading are clarified as follow, depending on the time needed to load the parts thumbnails:

Card group (including metadata - e.g. part name, price, ID, etc.) is visualized in skeleton mode while loading.

If metadata for all cards is loaded before all preview images are loaded, the metadata is displayed for all cards as soon as the last piece of metadata loads. Images continue to load in skeleton

If all images load within the x period (a period of time, usually measured in seconds, that you need to specify depending on the use case), they are shown as soon as the last image loads.

If all images have not finished loading at the end of the x period, the loaded images are shown one by one, and the pending images are visualized in skeleton mode while still loading.

So, apparently we want potentially to show metadata before the image and keep the images as skeleton.

santiagoballadares (approved):

Issue #801, opened 2023-08-14
KtCard improvements


RueRivoli:
**Is your feature request related to a problem? Please describe.**
This issue will contain all the KtCard improvements needed found while working on the catalog Tile View

**Describe the solution you'd like**
First of all: give a border: 1 px solid var(--ui-01) as default for KtCard to avoid increasizing size when hovering

**Describe alternatives you've considered**
/ 

**Additional context**
Other corrections to be potentially added

PR #806, opened 2023-08-15
feature(#801): give a default border to clickable cards

RueRivoli:
- give border a default value of the background color


RueRivoli:
close because will be updated in another PR

Issue #807, opened 2023-08-16
Replace the view icon by the tile_layout icon and changes border background color


RueRivoli:
**Is your feature request related to a problem? Please describe.**
Make the `tile_layout` icon replace the view icon and update the border's background color of ktcard: this is supposed to be transparent

**Describe the solution you'd like**
The view icon is not used anymore so it makes sense to replace the view icon by the tile view icon

**Describe alternatives you've considered**

**Additional context**
Make the `tile_layout` icon replace the view icon

PR #808, opened 2023-08-16
chore(#807): replace svg file for view by tile_layout

RueRivoli:
- replaced view icon by the "tile_view" icon
- modified kt card border's background-color to --ui-background


carsoli:
With the change you did, you need to make sure me or Santiago actually publish, otherwise the commit doesn't make sense

because you change the yoco version here and use it on kotti-ui directly (but it has to be actually published to npm, you can only access it now because you're within the monorepo)

carsoli:
I approve but please resolve the conflicts with caution and don't forget to ask either of us to publish for you @RueRivoli

carsoli (approved):

carsoli (approved):

Issue #812, opened 2023-08-25
Create new value for breakpoint 1920px


RueRivoli:
**Is your feature request related to a problem? Please describe.**
Add a new responsive breakpoint for the tile view equal to 1920px

**Describe the solution you'd like**
Add a new responsive breakpoint for the tile view equal to 1920px

**Describe alternatives you've considered**
/

**Additional context**
Design requests for the tile view to use an additional breakpoint

PR #813, opened 2023-08-25
feature(#812): add breakpoint 3x for 1920 px

RueRivoli:
Added a breakpoint 3x for 1920px to be used in the catalog tile view to make it responsive.

### DIFF
--- a/packages/kotti-ui/source/kotti-style/_variables.scss
+++ b/packages/kotti-ui/source/kotti-style/_variables.scss
@@ -81,6 +81,7 @@ $size-md: 840px !default;
 $size-lg: 960px !default;
 $size-xl: 1280px !default;
 $size-2x: 1440px !default;
+$size-3x: 1920px !default;
 
 $responsive-breakpoint: $size-xs !default;
 


packages/kotti-ui/source/kotti-style/_variables.scss
```diff
@@ -81,6 +81,7 @@ $size-md: 840px !default;
 $size-lg: 960px !default;
 $size-xl: 1280px !default;
 $size-2x: 1440px !default;
+$size-xxl: 1920px !default;
```
carsoli, line 84: do you know if 2x is used anywhere

if xxl needs to be 1920 then you should either make the value of size-2x into 1920 OR you should delete 2x

OR you keep 2x and make xxl --> 3x

Please talk to Alberto about this and involve me if needed

RueRivoli: @carsoli 2x is used in ampi.
I changed xxl to 3x.

santiagoballadares (approved):

carsoli (approved):

PR #817, opened 2023-10-10
feat(kt-value-label): add KtValueLabel

santiagoballadares:
Add KtValueLabel with support for:

- Text values only
- Help text (text and slot)
- Help description
- Unset text
- Validation text
- Default slot

### DIFF
--- a/packages/kotti-ui/source/kotti-value-label/KtValueLabel.vue
+++ b/packages/kotti-ui/source/kotti-value-label/KtValueLabel.vue
@@ -1,0 +1,180 @@
+<template>
+	<div :class="rootClasses" :data-test="dataTest">
+		<template v-if="isLoading">
+			<div
+				v-if="showHeader || $slots.helpText"
+				class="kt-value-label__loading__header skeleton rectangle"
+			/>
+			<div class="kt-value-label__loading__value skeleton rectangle" />
+		</template>
+		<template v-else>
+			<div v-if="showHeader || $slots.helpText" class="kt-value-label__header">
+				<label
+					v-if="hasLabel"
+					class="kt-value-label__header__label"
+					v-text="label"
+				/>
+				<div
+					v-if="hasHelpText || $slots.helpText"
+					class="kt-value-label__header__help-text"
+				>
+					<FieldHelpText :helpText="helpText" :helpTextSlot="$slots.helpText" />
+				</div>
+			</div>
+			<div
+				v-if="helpDescription"
+				class="kt-value-label__help-description"
+				v-text="helpDescription"
+			/>
+			<div class="kt-value-label__value">
+				<i v-if="isUnset" v-text="notSetLabel" />
+				<slot v-else-if="$slots.default" />
+				<span v-else v-text="valueToRender" />
+			</div>
+			<div
+				v-if="!isLoading && validationText !== null"
+				class="kt-value-label__validation-text"
+			>
+				<i class="yoco" v-text="validationTextIcon" />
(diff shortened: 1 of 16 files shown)


packages/kotti-ui/source/kotti-value-label/KtValueLabel.vue
```diff
@@ -0,0 +1,186 @@
+<template>
+	<div :class="rootClasses" :data-test="dataTest">
+		<template v-if="isLoading">
+			<div
+				v-if="hasLabel || hasHelpText || $slots.helpText"
```
carsoli, line 5: since it's repeated in line 5 and 8, i'd refact into a `showLabel` computed value

packages/kotti-ui/source/kotti-value-label/types.ts
```diff
@@ -0,0 +1,29 @@
...
+
+	export const propsSchema = z.object({
+		dataTest: z.string().nullable().default(null),
+		helpDescription: z.string().nullable().default(null),
+		helpText: z.string().nullable().default(null),
+		isLoading: z.boolean().default(false),
+		isUnset: z.boolean().default(false),
+		label: z.string().nullable().default(null),
+		validation: validationSchema.nullable().default(null),
+		value: z.union([z.number(), z.string()]).nullable().default(null),
```
carsoli, line 20: 1. why not boolean? 
2. why not an array of values as well? where the value can be comma-separated or hyphen separated? 
3. why not date? we support DisplayDate

packages/kotti-ui/source/kotti-value-label/types.ts
```diff
@@ -0,0 +1,29 @@
...
+import { KottiField } from '../kotti-field/types'
+
+export namespace KottiValueLabel {
+	export const validationSchema = z.union([
+		KottiField.Validation.errorSchema,
+		KottiField.Validation.successSchema,
+		KottiField.Validation.warningSchema,
+	])
+
+	export const propsSchema = z.object({
```
carsoli, line 12: how do we support DisplayPrice if we don't support a formatter prop?

packages/kotti-ui/source/kotti-value-label/types.ts
```diff
@@ -0,0 +1,29 @@
...
+		dataTest: z.string().nullable().default(null),
```
carsoli, line 13: does it make sense that we pick this from the KtField propsSchema

packages/kotti-ui/source/kotti-value-label/types.ts
santiagoballadares: Current scope was to support text values only with other values (number, currency, date) as follow-up work. We need to align on this.

carsoli: Fair enough

packages/kotti-ui/source/kotti-value-label/index.ts
```diff
@@ -0,0 +1,23 @@
+import { FIELD_META_BASE_SLOTS } from '../kotti-field/meta'
+import { attachMeta, makeInstallable } from '../utilities'
+
+import KtValueLabelVue from './KtValueLabel.vue'
+import { KottiValueLabel } from './types'
+
+export const KtValueLabel = attachMeta(makeInstallable(KtValueLabelVue), {
+	addedVersion: '5.3.0',
```
carsoli, line 8: 5.3.1

carsoli (approved):

PR #826, opened 2023-11-03
chore(storybook): delete storybook package and refact useTippy

carsoli:
and refactor null check on useTippy target argument


FlorianWendelborn (dismissed):

packages/kotti-ui/source/kotti-avatar/KtAvatar.vue
```diff
@@ -47,7 +47,7 @@ export default defineComponent({
 			computed(() => ({
 				appendTo: () => document.body,
 				arrow: roundArrow,
-				content: contentRef.value as NonNullable<typeof contentRef.value>,
+				content: contentRef.value ?? undefined,
```
carsoli, line 50: useTippy has an `onMounted` hook inside it that only binds the tippy instance when the component mounts

so IN THEORY, tippy isn't bound unless the component mounts which means the content shouldn't be null

However, since tippy seems to handle undefined content just fine, we unified all the usages to fallback to undefined if contentRef.value is still null

packages/vue-use-tippy/README.md
```diff
@@ -10,13 +10,13 @@ import { computed, defineComponent, ref } from '@vue/composition-api'
...
 	setup() {
-		const contentRef = ref(null)
+		const contentRef = ref<Element | null>(null)
 
 		useTippy(
 			helpTextTriggerRef,
 			computed(() => ({
 				appendTo: () => document.body,
-				content: contentRef,
+				content: contentRef ?? undefined,
```
carsoli, line 19: fix: contentRef.value

packages/vue-use-tippy/source/index.ts
```diff
@@ -6,12 +6,13 @@ import {
...
 	watch,
 } from '@vue/composition-api'
+
 import castArray from 'lodash.castarray'
 import tippy, { Props, Instance } from 'tippy.js'
 
 type InstanceRefType = Instance<Props>[] | Instance<Props> | null
 type Callback = (t: InstanceRefType) => void
-type Target = Ref<Element | null> | string
+type Target = Ref<Element | null>
```
carsoli, line 15: string was removed from the union type because we use `Target[]` and tippy targets can't be `string[]`

so in order to be able to reuse `Target` as an array type, we split the `string` out of the union and support it directly on argument type

packages/vue-use-tippy/tsconfig.json
```diff
@@ -6,7 +6,7 @@
 		"declarationDir": "dist",
 		"esModuleInterop": true,
 		"experimentalDecorators": true,
-		"lib": ["ESNext", "ES2019"],
+		"lib": ["ESNext", "ES2019", "DOM"],
```
carsoli, line 9: when removing storybook, the DOM lib became EXPLICITLY needed .. because vue-use-tippy relies on the `Element` type which is a DOM lib type

packages/vue-use-tippy/source/index.ts
```diff
@@ -22,27 +23,33 @@ const applyForEvery = (
...
-			if (isRef(targets)) return targets.value
-
-			if (Array.isArray(targets))
-				return targets.map((target) => (isRef(target) ? target.value : target))
-
-			return targets
-		})()
-
-		instance.value = tippy(unwrappedTargets, options.value)
+		if (typeof targets === 'string') {
```
carsoli, line 43: IIFE was removed in favor of a clear control flow for 1) readability
2) typescript compatability
because `tippy` has two signatures:
1. one that accepts `MultipleTargets`
2. one that accepts `SingleTarget`

using an IIFE to return unwrapped targets made both types unioned out of the outcome of that IIFE.. which confused typescript when we had one assignment statement
`instance.value = tippy(...)` because the signatures shouldn't overlap


Could have been fixed differently, for example, by calling `tippy` within the function expression BUT this way is more readable anyways

packages/vue-use-tippy/source/index.ts
carsoli: this was previously a mistake, and thus is actually a FIX

FlorianWendelborn (approved):

PR #828, opened 2023-11-06
feat(kt-filters): add date range shortcuts support

santiagoballadares:
Add support for shortcuts, minimum date, maximum date in DateRange filters.

### DIFF
--- a/packages/documentation/pages/usage/components/filters.vue
+++ b/packages/documentation/pages/usage/components/filters.vue
@@ -78,6 +78,29 @@
     						label="maximum"
     					/>
     				</div>
+    				<div class="field-row">
+    					<KtFieldDate
+    						formKey="dateRangeMaximumDate"
+    						helpText="Support on DATE_RANGE column type only"
+    						isOptional
+    						label="maximumDate"
+    					/>
+    					<KtFieldDate
+    						formKey="dateRangeMinimumDate"
+    						helpText="Support on DATE_RANGE column type only"
+    						isOptional
+    						label="minimumDate"
+    					/>
+    				</div>
+    				<KtFieldToggleGroup
+    					formKey="NONE"
+    					helpText="Support on DATE_RANGE column type only"
+    					isOptional
+    					label="shortcuts"
+    					:options="shortcutsOptions"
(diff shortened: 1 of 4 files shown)


packages/documentation/pages/usage/components/filters.vue
```diff
@@ -78,6 +78,29 @@
     						label="maximum"
     					/>
     				</div>
+    				<div class="field-row">
+    					<KtFieldDate
```
carsoli, line 82: but why not KtFieldDateRange then?

packages/documentation/pages/usage/components/filters.vue
```diff
@@ -120,12 +143,40 @@
...
+import { ISO8601_DAY } from '../../../../kotti-ui/source/constants'
 import { ComponentValue, generateComponentCode } from '../../utilities'
 
 import ComponentInfo from '~/components/ComponentInfo.vue'
 
+const today = (): string => dayjs().format(ISO8601_DAY)
+const getLast = (unit: 'day' | 'month' | 'week' | 'year') =>
+	dayjs().subtract(1, unit).format(ISO8601_DAY)
+
+const Shortcuts: Record<
```
carsoli, line 158: casing: `shortcuts`

packages/kotti-ui/source/constants.ts
```diff
@@ -1,3 +1,4 @@
+export const ISO8601_DAY = 'YYYY-MM-DD' as const
```
carsoli, line 1: does this not exist already as ISO8601? should exist because we have a const for the KtFieldDate format

santiagoballadares: Yes but the format that `element-ui` uses is different and not compatible with the one dayjs uses.
Screenshot 2023-11-07 at 12 31 49
Screenshot 2023-11-07 at 12 39 37

carsoli: Interesting, got it, then I'd just call it `ISO8601`, there's no point in adding _DAY

we add _SECONDS for the other const because it's the full format

carsoli (dismissed):

PR #830, opened 2023-11-09
Rid vue clickaway

Isokaeder:
Replace last usages of vue-clickaway with tippy


packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
```diff
@@ -31,15 +31,14 @@
 					:links="quickLinks"
 				/>
 			</div>
-			<div v-if="mobileMenuToggle" class="kt-navbar__dropdown">
+			<div ref="mobileMenuRef" class="kt-navbar__dropdown">
```
carsoli, line 34: no v-if or v-show?

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
```diff
@@ -72,32 +73,98 @@ export default defineComponent({
...
+				content: mobileMenuRef.value ?? undefined,
+				interactive: true,
+				maxWidth: 'none',
+				offset: [0, 0],
+				onCreate(instance) {
+					tippyInstanceRef.value = instance
+				},
+				placement: 'bottom-start',
+				theme: props.theme
+					? `kt-navbar-${props.theme} kt-navbar-shadow`
```
carsoli, line 102: Florian: if there is a theme, you don't add `kt-navbar` as a standalone class 

IOW misisng `kt-navbar`

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
```diff
@@ -72,32 +73,98 @@ export default defineComponent({
...
 				[`kt-navbar--theme-${props.theme}`]: props.theme !== null,
 			})),
-			clickawayMobileMenu: () => (mobileMenuToggle.value = false),
-			mobileMenuToggle,
-			setIsNarrow: (isNarrow: boolean) => emit('setIsNarrow', isNarrow),
-			toggleMobileMenu() {
-				mobileMenuToggle.value = !mobileMenuToggle.value
+			hideMobileMenu: (event: MouseEvent) => {
+				emit('linkClick', event)
+				tippyInstanceRef.value?.hide()
```
carsoli, line 116: yep that's why

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
```diff
@@ -72,32 +73,98 @@ export default defineComponent({
...
+	--kt-navbar-background: var(--primary-10);
+	--kt-navbar-color: var(--primary-80);
+}
+
+.tippy-box[data-theme~='kt-navbar-shadow'] {
+	box-shadow: 0 2px 4px rgba(0, 0, 0, 0.26);
+}
+</style>
+
+<style lang="scss">
```
carsoli, line 167: did you mean to scope this?

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
```diff
@@ -72,32 +73,98 @@ export default defineComponent({
...
+			setIsNarrow: (isNarrow: boolean) => emit('setIsNarrow', isNarrow),
+			tippyTriggerRef,
+			Yoco,
 		}
 	},
 })
 </script>
 
-<style lang="scss" scoped>
+<style lang="scss">
```
carsoli, line 128: but why not scope this?

if other usages of tippy in kotti rely on the theme, all the styles (bg-color, border-radius, color, etc) will leak, no?

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
```diff
@@ -111,10 +178,7 @@ $narrow-navbar-width: 3.4rem;
 	--navbar-color-light: var(--primary-20);
 	--navbar-color-active: var(--white);
 
-	--user-menu-border: var(--primary-80);
-	--user-menu-background-active: var(--primary-70);
 	--user-menu-background: var(--primary-60);
-	--user-menu-color: var(--primary-10);
 
 	position: relative;
```
carsoli, line 183: is this needed now? there's no need for controlling absolutes/relatives, tippy handles that, no?

packages/kotti-ui/source/kotti-user-menu/KtUserMenu.vue
```diff
@@ -67,58 +69,153 @@ export default defineComponent({
...
+				maxWidth: 'none',
+				offset: [0, 0],
+				onHide: () => {
+					isMenuShow.value = false
+				},
+				onShow: () => {
+					isMenuShow.value = true
+				},
+				theme: context.value.theme
+					? `kt-usermenu-${context.value.theme}`
```
FlorianWendelborn, line 100: Kinda looks like `theme` should default to something like `default` and you can get rid of the conditionals etc. around that then

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
FlorianWendelborn: The theme is specific to the `kt-navbar`. It has to be unscoped because tippy attaches to body

FlorianWendelborn (dismissed):

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
Isokaeder: What Florian says. Scoping would hide all styles from this component, as the tippied content is not appended as a child here but at the html body.

Try it out yourself if you have to see it :-)

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
Isokaeder: I don't think this is needed. But not because of tippy (as this is the Navbar itself, not the dropdown for mobile) but because of `kt-container`.

I can remove, but it is a breaking change for users that do not use `.kt-container`

Isokaeder: sorry, not `.kt-container` is needed, strictly only a flexbox is needed as a container.

carsoli:
Breaking: the css variables were interchangeably used with the `user-menu` variables and have been better scoped by @Isokaeder 

It's a breaking change in case of some reliance on them by a user-app

There is other mentions of the navbar css variables in kotti's code base (not sure how that works but the theming did work)

packages/kotti-ui/source/kotti-navbar/KtNavbar.vue
carsoli: I'm slightly confused by this comment :D, I don't know what you mean

FlorianWendelborn (dismissed):

packages/kotti-ui/source/kotti-user-menu/KtUserMenu.vue
```diff
@@ -67,144 +66,177 @@ export default defineComponent({
...
 
-		&__text {
-			z-index: 2;
-			width: 100%;
-			margin-left: 0.4rem;
-			line-height: 0.8rem;
-			color: var(--user-menu-color);
-		}
+.kt-user-menu-info {
+	position: relative;
```
carsoli, line 178: delete

packages/kotti-ui/source/kotti-user-menu/KtUserMenu.vue
```diff
@@ -67,144 +66,177 @@ export default defineComponent({
...
+	display: flex;
+	flex-direction: row;
+	justify-content: space-between;
+	padding: 0.4rem;
+	margin: -0.4rem;
+	line-height: 1;
+	cursor: pointer;
+	border-radius: 0 0 0.2rem 0.2rem;
```
carsoli, line 186: isn't this supposed to be 0.2 0.2 0 0 ?

carsoli (approved):

Issue #833, opened 2023-11-13
Tags are always capitalized


Isokaeder:
**Describe the bug**
`KtTag` always capitalizes the string passed into the `text` prop.

**To Reproduce**

1. Create Tag with:
```html
 <KtTag text="lower cased"/>
```
2. Will be rendered as:
<img width="114" alt="Screenshot 2023-11-13 at 12 54 32" src="https://github.com/3YOURMIND/kotti/assets/16950410/6804c4a8-aee6-4f87-ad7f-d18c4447df46">

**Expected behavior**
Should be rendered as:
<img width="110" alt="Screenshot 2023-11-13 at 12 55 08" src="https://github.com/3YOURMIND/kotti/assets/16950410/14544c6b-6025-4e90-a5e9-b25f29e216a7">

**Additional context**
This happens because the tag was only used internally and then the capitalization was just kept when we made it a public component.

PR #834, opened 2023-11-13
fix(#833): remove capitalize transform

Isokaeder:
I am not sure why that original `text-transform: capitalize` was there. 

I kept the behavior around for the multi select field specifically, but not sure if it's desired. It can lead to inconsistencies like this.
Screenshot 2023-11-13 at 13 07 22

I could also introduce a prop `isCapitalized` for this, but think it's overkill, as users can easily control this via CSS if desired.

### DIFF
--- a/packages/kotti-ui/source/kotti-field-select/components/GenericSelectField.vue
+++ b/packages/kotti-ui/source/kotti-field-select/components/GenericSelectField.vue
@@ -14,12 +14,14 @@
 					<KtTag
 						v-for="option in visibleSelectedTags"
 						:key="option.value"
+						class="kt-field-select__tag"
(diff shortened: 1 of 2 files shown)


packages/kotti-ui/source/kotti-field-select/components/GenericSelectField.vue
```diff
@@ -14,12 +14,14 @@
 					<KtTag
 						v-for="option in visibleSelectedTags"
 						:key="option.value"
+						class="kt-field-select__input-and-tags__tag"
```
carsoli, line 17: doesn't follow BEM convention, will edit after merging the stylelint MR
team will discuss this, this week in the chapter meet

carsoli (approved):

carsoli:
fixes #833

PR #838, opened 2023-11-20
dep(rollup): bump to rollup 3.x and most plugins

Isokaeder:
Update rollup to v3.x and plugins wherever possible.

The following version bumps were **not** made:
- `rollup` to v4.x  - because it [requires node 18](https://rollupjs.org/migration/#prerequisites) which AFAIU we do [not have activated yet](https://github.com/3YOURMIND/kotti/pull/680) in our CI
- `rollup-plugin-vue` to ^6.0.0 - because it requires Vue3 (but is also [deprecated](https://github.com/vuejs/rollup-plugin-vue#archived))

#### Bundle Size

negligible impact, for CJS the bundle size increased by ~20kB, for ESM it increased by <5kB

#### Typescript

rollup will now fail when errors were found during build (probably due to [this](https://github.com/rollup/plugins/blob/master/packages/typescript/CHANGELOG.md#v600), so could be achieved without bumping

### Side Note

I do not know our long term plan if we want to eventually move to another build system. If so, and this is more or less a distractionf, feel free to close.

### DIFF
--- a/packages/kotti-ui/package.json
+++ b/packages/kotti-ui/package.json
@@ -27,11 +27,11 @@
 		"@3yourmind/sass-node-modules-importer": "*",
 		"@babel/plugin-transform-runtime": "^7.15.0",
 		"@babel/runtime": "^7.15.4",
-		"@rollup/plugin-babel": "^5.1.0",
-		"@rollup/plugin-commonjs": "12",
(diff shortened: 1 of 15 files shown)


carsoli:
We are doing this in the naruto MR

PR #841, opened 2023-11-24
feat(KtField): scope the isLoading skeleton to the input only

santiagoballadares:
Add isLoadingValue prop to show loading state only on the input area of the field

### DIFF
--- a/packages/kotti-ui/source/kotti-field/KtField.vue
+++ b/packages/kotti-ui/source/kotti-field/KtField.vue
@@ -2,7 +2,6 @@
 	<div class="kt-field">
 		<component
 			:is="isGroup ? 'fieldset' : isComponent ? isComponent : 'label'"
-			v-show="!field.isLoading"
 			:class="wrapperClasses"
 			@click="$emit('click', $event)"
 			@mousedown="$emit('mousedown', $event)"
@@ -32,7 +31,11 @@
 				class="kt-field__help-description"
 				v-text="field.helpDescription"
 			/>
-			<div class="kt-field__input-container-wrapper">
+			<div
+				v-if="field.isLoading"
+				class="kt-field__input-container-wrapper-loading skeleton rectangle"
+			/>
+			<div v-show="!field.isLoading" class="kt-field__input-container-wrapper">
 				<div
 					v-if="$slots['container-left']"
 					class="kt-field__input-container__prefix"
@@ -98,13 +101,6 @@
 				{{ validationText }}
 			</div>
 		</component>
-		<div v-if="field.isLoading" class="kt-field__wrapper">
-			<div
-				v-if="hasLabel || hasHelpText"
-				class="kt-field__loading__header skeleton rectangle"
-			/>
-			<div class="kt-field__loading__input-container skeleton rectangle" />
-		</div>
 	</div>
 </template>
 
@@ -241,17 +237,6 @@ we would be able to extend on demand instead of unscoping all field classes -->
 		margin-bottom: 0.8rem;
 	}
 
-	&__loading {
-		&__header {
-			max-width: 200px;
-			height: 20px !important;
-		}
-
-		&__input-container {
-			height: 40px !important;
-		}
-	}
-
 	&__wrapper {
 		display: flex;
 		flex-direction: column;
@@ -367,6 +352,10 @@ we would be able to extend on demand instead of unscoping all field classes -->
 		display: flex;
 		gap: var(--unit-2);
 		align-items: stretch;
+	}
+
+	&__input-container-wrapper-loading {
+		height: 40px !important;
 	}
 
 	// placeholders for slots
--- a/packages/kotti-ui/source/kotti-field/types.ts
+++ b/packages/kotti-ui/source/kotti-field/types.ts
@@ -138,7 +138,7 @@ export namespace KottiField {
 		isDisabled: z.boolean().default(false),
 
 		/**
-		 * Show a skeleton in place of the field
+		 * Show a skeleton in place of the input area of the field
 		 */
 		isLoading: z.boolean().default(false),
 


packages/kotti-ui/source/kotti-field/KtField.vue
```diff
@@ -32,7 +32,10 @@
 				class="kt-field__help-description"
 				v-text="field.helpDescription"
 			/>
-			<div class="kt-field__input-container-wrapper">
+			<div
+				v-show="!field.isLoadingValue"
```
carsoli, line 36: why the v-show?

carsoli:
I don't understand why we need isLoading vs isLoadingValue...

Has the design decided we have two cases?
Also, there is certain implementation on the KtForm that makes the `isLoading` prop one of the inheritable props, so that a user app can pass `isLoading` on a `KtForm` and it would trickle down to all fields in that form contexts and update their `isLoading` here
..The logic would have to be replicated if we want to keep both props

I do, however, find it a bad decision to bloat the API like this.. Design needs to decided what they really want here, they claimed the label is barely affected by the async API requests, why are we keeping both implementations then?

packages/kotti-ui/source/kotti-field/KtField.vue
santiagoballadares: KtFieldDates have hooks that break with `v-if`

carsoli:
> > I don't understand why we need isLoading vs isLoadingValue...
> > Has the design decided we have two cases? Also, there is certain implementation on the KtForm that makes the `isLoading` prop one of the inheritable props, so that a user app can pass `isLoading` on a `KtForm` and it would trickle down to all fields in that form contexts and update their `isLoading` here ..The logic would have to be replicated if we want to keep both props
> > I do, however, find it a bad decision to bloat the API like this.. Design needs to decided what they really want here.
> 
> Yes, design wants 2 different cases for this; and yes, I missed the KtForm stuff. So, drafting this for now.

I see you touched on the `inheritableProps` type
and `useInheritableProperties` 

I don't know why and if I was on an outdated version of the MR when I left the comment that concerned inheriting from the `KtForm` context

there is no issue with this as far as I can see.


Can you relay the reason why the design saw a reason to distinguish both scenarios?

santiagoballadares:
> > > I don't understand why we need isLoading vs isLoadingValue...
> > > Has the design decided we have two cases? Also, there is certain implementation on the KtForm that makes the `isLoading` prop one of the inheritable props, so that a user app can pass `isLoading` on a `KtForm` and it would trickle down to all fields in that form contexts and update their `isLoading` here ..The logic would have to be replicated if we want to keep both props
> > > I do, however, find it a bad decision to bloat the API like this.. Design needs to decided what they really want here.
> > 
> > 
> > Yes, design wants 2 different cases for this; and yes, I missed the KtForm stuff. So, drafting this for now.
> 
> I see you touched on the `inheritableProps` type and `useInheritableProperties`
> 
> I don't know why and if I was on an outdated version of the MR when I left the comment that concerned inheriting from the `KtForm` context
> 
> there is no issue with this as far as I can see.
> 
> Can you relay the reason why the design saw a reason to distinguish both scenarios?

During grooming it was agreed that the loading was going to be changed to only show the loading bar on the value and not the labels but afterwards, design explained that they had from the beginning 2 different loading cases in mind.
Already complained about inaccurate AC during grooming.

For context, this is design's comment on this:

> We have 2 different cases for the loading:
> 
> 1. You are loading the entire field - e.g. on page load of a form, on switching tabs, on opening a drawer, etc -> no changes expected for this, the whole form field is in loading state (label, helptext, input, etc.)
> 
> 2. You are updating a value and a communication to BE is required in the specific moment - e.g. changing value of the inline field, changing a mail input with BE validation, etc. -> then ONLY the input field is in loading state
> 
> The ticket is targeting ONLY case 2, no changes should apply to case 1

carsoli:
> > > > I don't understand why we need isLoading vs isLoadingValue...
> > > > Has the design decided we have two cases? Also, there is certain implementation on the KtForm that makes the `isLoading` prop one of the inheritable props, so that a user app can pass `isLoading` on a `KtForm` and it would trickle down to all fields in that form contexts and update their `isLoading` here ..The logic would have to be replicated if we want to keep both props
> > > > I do, however, find it a bad decision to bloat the API like this.. Design needs to decided what they really want here.
> > > 
> > > 
> > > Yes, design wants 2 different cases for this; and yes, I missed the KtForm stuff. So, drafting this for now.
> > 
> > 
> > I see you touched on the `inheritableProps` type and `useInheritableProperties`
> > I don't know why and if I was on an outdated version of the MR when I left the comment that concerned inheriting from the `KtForm` context
> > there is no issue with this as far as I can see.
> > Can you relay the reason why the design saw a reason to distinguish both scenarios?
> 
> During grooming it was agreed that the loading was going to be changed to only show the loading bar on the value and not the labels but afterwards, design explained that they had from the beginning 2 different loading cases in mind. Already complained about inaccurate AC during grooming.
> 
> For context, this is design's comment on this:
> 
> > We have 2 different cases for the loading:
> > 
> > 1. You are loading the entire field - e.g. on page load of a form, on switching tabs, on opening a drawer, etc -> no changes expected for this, the whole form field is in loading state (label, helptext, input, etc.)
> > 2. You are updating a value and a communication to BE is required in the specific moment - e.g. changing value of the inline field, changing a mail input with BE validation, etc. -> then ONLY the input field is in loading state
> > 
> > The ticket is targeting ONLY case 2, no changes should apply to case 1

even if the whole form is loading in terms of fetching data-from the backend on initial page/form load, the label strings are static translations all across our app

I think design is over-complicating the situation here 🤷🏼

carsoli (approved):

carsoli:
@santiagoballadares made the changes as discussed with design and agreed with you

PR #842, opened 2023-11-27
docs(kt-field-date): Add missing shortcuts documentation

santiagoballadares:
Add missing KtField[Date|DateRange|DateTime|DateTimeRange] shortcuts documentation


packages/documentation/pages/usage/components/form-fields.vue
```diff
@@ -1127,6 +1193,7 @@ export default defineComponent({
 				numberMaximum: null,
 				numberMinimum: null,
 				numberStep: null,
+				shortcuts: null,
```
carsoli, line 1196: shouldn't it be `[]`

santiagoballadares: For the purpose of the docs, this `shortcuts` is a `KtFieldToggleGroup` and not the actual KtFieldDate's `shortcuts` prop. Its value is `Record<string, boolean | null> | null`

carsoli (approved):

PR #846, opened 2023-12-01
Internal Attachments and some fixes

Isokaeder:
### Note

At this stage this branch still contains an additional page in the documentation that tests the flow that this change is trying to support. I left it in to make my changes more convenient to test, considering that the existing documentation is for one broken and for two doesn't make it easy to  document my changes anyway

### Internal Attachments
File value can contain optional flag `isInternal`. If it is `true`, the files are going to be displayed in a slightly different way. It     is the same concept as internal comments

Screenshot 2023-12-01 at 10 08 48

### New Status `HIDDEN

New status `HIDDEN` that can hide files from the list for `KtFieldFileUploadRemote` in cases the consumer needs something like this. For example in the workflow in `file-upload-remote.vue` a modal is asking the user for additional input and while that is open, the files are hidden.

## Bugfixes

- (57a8915022119ea32b0dc0f9b5412b022e051e09) There was a race condition when the watcher in `KtFieldFileUploadRemote.vue` would receive a payload that changed in more than one places and therefore fire multiple emits at once. After the fix only one emit is fired.
- (b8c4f08f22c6c1d085c207aa33d2e6c30d78fece) The progress bar would get out of sync with the actual progress.


RueRivoli:
LGTM

RueRivoli (dismissed):

RueRivoli:
There are a few console errors after the upload of file.
Screenshot 2023-12-01 at 15 51 47
Screenshot 2023-12-01 at 15 51 59

packages/kotti-ui/source/kotti-field-file-upload/components/FileItemRemote.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-	<ItemLayout :name="fileInfo.name">
+	<ItemLayout
+		:class="wrapperClasses"
```
santiagoballadares, line 3: Shouldn't the `ItemLayout` handle its own styling? It is already accepting the new `isInternal` flag which is what these wrapper classes depend on.

packages/kotti-ui/source/kotti-field-file-upload/components/FileItemRemote.vue
```diff
@@ -61,7 +65,7 @@ import ItemLayout from './ItemLayout.vue'
 import ProgressBar from './ProgressBar.vue'
 
 export default defineComponent({
-	name: 'FileItem',
+	name: 'FileItemRemote',
```
santiagoballadares, line 68: Good catch

packages/kotti-ui/source/kotti-field-file-upload/components/FileItemRemote.vue
```diff
@@ -1,5 +1,9 @@
 <template>
```
santiagoballadares, line 1: Line `84` (`? translations.value.statusMsg[status.value]`) may need to be updated. It is indexing `statusMsg` in the translation obj but there's no entry for the new status `HIDDEN`. In theory, it shouldn't happen as these items are not rendered but TS is complaining (maybe pipelines too)

packages/kotti-ui/source/kotti-field-file-upload/KtFieldFileUploadRemote.vue
```diff
@@ -147,6 +151,24 @@ export default defineComponent({
 				newUploadedFiles.add(payload.id)
 		}
 
+		const setStatuses = (
+			payload: Map<string, KottiFieldFileUploadRemote.Status>,
+		) => {
+			field.setValue(
+				field.currentValue.map((fileItem) => {
+					const status =
+						typeof fileItem.id === 'string' && payload.get(fileItem.id)
```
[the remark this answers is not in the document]
Isokaeder: Ah, good point... the problem here was that payload `Kotti.FieldFileUploadRemote.Payload` is typed as `Record<string, _>`, i.e. the strings are assumed to be strings. I took it to mean that the payload would always ever contain ids that we randomly generate (which I think is always a string) and therefore thought this is reasonable.

If I remove this check, there is a typescript complaint like this:
```ts
const status = payload.get(fileItem.id)
                        // ^^^^^^^^^^^
                        // Argument of type 'string | number' is not assignable to parameter of type 'string'.
```

The other way to solve this would be to cast:
```ts
const status = payload.get(String(fileItem.id))
```
I think that is the proper way to handle this.

packages/documentation/pages/usage/file-upload-remote.vue
```diff
@@ -0,0 +1,208 @@
...
+			return MockBackend.upload(
+				file.file,
+				(progress, total) => {
+					const fileStatus = payload.value[fileId]
+					const newProgress = progress / total
+					payload.value = {
+						...payload.value,
+						[fileId]: {
+							...fileStatus,
+							progress: Math.max(newProgress, fileStatus.progress ?? 0),
```
santiagoballadares, line 113:
```suggestion
							progress: Math.min(
								Math.max(newProgress, fileStatus.progress ?? 0),
								1,
							),
```

Isokaeder: I will change it, but my plan is to remove this file for now and reinsert it once I have time to give the upload fields some actual documentation.

RueRivoli (approved):

PR #850, opened 2023-12-05
fix(KtFieldToggle/KtFieldSelects/KtFieldDates/KtFieldFileUpload): isLoading edge cases

carsoli:
- prevent triggering setValue on isLoading for KtFieldToggle
- prevent triggering setValue on isLoading for KtFieldDate, KtFieldDateTime, KtFieldDateRange, KtFieldDateTimeRange
- prevent triggering tippy on isLoading for KtFieldSelect, KtFieldSelectRemote, KtFieldMultiSelect, KtFieldMultiSelectRemote
- throw on KtField if setValue is triggered while isLoading is true (to guide kotti devs)
- disable ktFieldUpload(Remote) input natively on isLoading (prevents opening system file explorer)
- breaking: remove isLoading skeleton margins to avoid weird visual jumping of the field when the isLoading is toggled

### DIFF
--- a/packages/kotti-ui/source/kotti-field-toggle/KtFieldToggle.vue
+++ b/packages/kotti-ui/source/kotti-field-toggle/KtFieldToggle.vue
@@ -72,7 +72,8 @@ export default defineComponent({
 				forceUpdateKey: forceUpdateKey.value,
 			})),
 			onInput: (newValue: boolean | undefined) => {
-				field.setValue(newValue ?? null)
+				if (!field.isDisabled && !field.isLoading)
+					field.setValue(newValue ?? null)
 
 				forceUpdate()
 			},
--- a/packages/kotti-ui/source/kotti-style/_loadings.scss
+++ b/packages/kotti-ui/source/kotti-style/_loadings.scss
@@ -14,11 +14,10 @@
 	}
 }
 
-@mixin skeleton-style($bg-x, $bg-y) {
+%skeleton-style {
 	box-sizing: border-box;
 	display: block;
 	width: 100%;
-	margin: 0.4rem 0;
 	line-height: 1px;
 	background: linear-gradient(
 		to left,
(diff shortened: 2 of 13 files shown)


FlorianWendelborn (dismissed):

packages/kotti-ui/source/kotti-field-toggle/KtFieldToggle.vue
```diff
@@ -72,7 +72,8 @@ export default defineComponent({
 				forceUpdateKey: forceUpdateKey.value,
 			})),
 			onInput: (newValue: boolean | undefined) => {
-				field.setValue(newValue ?? null)
+				if (!field.isDisabled && !field.isLoading)
```
santiagoballadares, line 75: Shouldn't it be as below? Right now, `forceUpdate()` is always executed.
```suggestion
				if (field.isDisabled || field.isLoading) return
```

carsoli: I wanted to go with Santiago's way of doing but you saw a reason not to ? can you explain why forceUpdate() would be needed in that case? Is there an edge case you had in mind where a forceUpdate would be required?

carsoli: That much I can understand but I'm confused as to how the field is even updated in the first place

carsoli: I keep getting confused by this myself
@santiagoballadares see the JS doc on `useForceUpdate`

packages/kotti-ui/source/kotti-style/_loadings.scss
```diff
@@ -14,41 +14,42 @@
 	}
 }
 
-@mixin skeleton-style($bg-x, $bg-y) {
+%skeleton-style {
 	box-sizing: border-box;
 	display: block;
 	width: 100%;
-	margin: 0.4rem 0;
```
carsoli, line 21: intentional

FlorianWendelborn (dismissed):
LGTM

packages/kotti-ui/source/kotti-style/_loadings.scss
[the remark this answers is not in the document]
carsoli: yep

packages/kotti-ui/source/kotti-field-toggle/KtFieldToggle.vue
santiagoballadares: TBH I'm confused, this: 

> whenever the field updates itself it needs to be forceUpdated

is clear, but I don't follow this:

> to restore the old value

shouldn't it be: `to make the native HTML input display the new value`?

I get that whenever the field's value is updated it needs to be forceUpdated but here when the field is loading all interactions are/should be blocked so that the field's value does not get updated in the first place, so no need to forceUpdated it bc the value didn't change, no?

Anyway, I suppose there's no harm in forceUpdating the field in either case, it'll just re-render.

packages/kotti-ui/source/kotti-style/_loadings.scss
carsoli: Design team agreed to remove this and instead make sure that two direct loadings have a separation margin of 0.4rem and not 0.8rem (which maintains the previous case)

carsoli:
@santiagoballadares I will need to test all the skeleton on yoda

santiagoballadares:
FT done

All good, only `KtFieldFileUploadRemote` needs fixing

santiagoballadares (approved):
LGTM

PR #852, opened 2023-12-07
feat(KtField): restrict clickable field area

carsoli:
we POCed a much simpler approach to restricting the clickable area of the KtField, simpler= doesn't introduce as many changes, what do you think? @santiagoballadares 

We tested with KtField*Select & KtFieldUpload* & KtFieldDates & with the ToggleGroup(which is just the `fieldset`)


carsoli:
Duplicates #849

santiagoballadares:
Closing, #849 was merged

PR #856, opened 2023-12-22
fix(kt-comment): remove blur behavior to prevent unwanted behavior wh…

santiagoballadares:
…ere new and edited text is lost on blur event


RueRivoli (approved):
@santiagoballadares this seems to work fine. Comment content is not lost when we click outside of the field.

Issue #864, opened 2024-01-09
Make the dropzone of KtFieldFileUpload and KtFieldFileUploadRemote hideable


RueRivoli:
**Is your feature request related to a problem? Please describe.**
In some cases, we would like to hide the dropzone of the KtFieldFileUpload s to prevent the user to upload new files/documents
This will happen in the catalog detail page where it shouldn't be an editable page 

**Describe the solution you'd like**
Display the list of uploaded files as usual but hide the dropzone.
So I imagine a property isDropzoneHidden or something similar

**Describe alternatives you've considered**
/ 

**Additional context**
See the B3-14040 ticket which relates to the implementation of the catalog detail page

PR #866, opened 2024-01-10
feature(#864): add a property hideDropArea for both KtFieldFileUpload…

RueRivoli:
This code changes aims at adding a property hideDropArea for both components KtFieldFileUpload and KtFieldFileUploadRemote
This would permit us to display attachments in a "read-only" way for the new catalog detail page.


Isokaeder (approved):
LGTM

PR #872, opened 2024-01-15
refactor(KtField*Select*): Use $props on Template

FlorianWendelborn:
..instead of manually exposing `props` object


carsoli (approved):

carsoli:
This caused issues with Vue 2.7 because reexposing props imploded the template types

PR #873, opened 2024-01-15
refactor(kotti): Avoid Using `export * as` Syntax

FlorianWendelborn:
It’s relatively modern syntax and doesn’t always have perfect tooling support 
Removed because it caused problems in the Vue 2.7 branch

Co-Authored-By: Carol Soliman <17387510+carsoli@users.noreply.github.com>


FlorianWendelborn (commented):
LGTM

carsoli (approved):

PR #890, opened 2024-01-18
feat(docs): add class to separate flex rows

carsoli:
#850 caused a UI bug on docs for `loadings` component, since it removed `margin-bottom`
> introduce a class to add a row-gap between loading skeletons where needed

removal of margin-bottom was intentional
> see comments on original PR

### DIFF
--- a/packages/documentation/pages/usage/components/loadings.vue
+++ b/packages/documentation/pages/usage/components/loadings.vue
@@ -20,7 +20,7 @@
     		<div class="column col-2">
     			<div class="skeleton square" />
     		</div>
-    		<div class="column col-8">
+    		<div class="column col-8 col-with-row-gap">
     			<div class="skeleton rectangle md" />
     			<div class="skeleton rectangle md" />
     			<div class="skeleton rectangle md" />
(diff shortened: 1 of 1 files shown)


packages/kotti-ui/source/kotti-style/_layout.scss
```diff
@@ -125,6 +125,12 @@
 	margin-right: auto;
 }
 
+.col-with-row-gap {
```
FlorianWendelborn, line 128: I don’t think we should be exposing more classes from Kotti.

FlorianWendelborn (changes requested):

packages/kotti-ui/source/kotti-style/_layout.scss
carsoli: change amended

carsoli: to the wrong file, another change about to be pushed

packages/kotti-ui/source/kotti-style/_loadings.scss
```diff
@@ -152,3 +152,4 @@
 		}
 	}
 }
+
```
FlorianWendelborn, line 155: Already had newline at EOF

FlorianWendelborn (approved):

carsoli:
@FlorianWendelborn need reapproval

FlorianWendelborn (approved):

PR #893, opened 2024-02-15
fix(kotti-ui): path to style.css

carsoli:
- remove dead code from vite.config
- fix README to point to correct paths of style.css


FlorianWendelborn (approved):
LGTM

PR #894, opened 2024-02-20
feat(split-button): add split button

santiagoballadares:
Add new component <KtSplitButton />

- Change button docs from markdown template to html template
- Add KtSplitButton to button docs


packages/kotti-ui/source/kotti-split-button/types.ts
```diff
@@ -0,0 +1,51 @@
...
+import { createLooseZodEnumSchema } from '../zod-utilities/enums'
+
+export module KottiSplitButton {
+	export enum Type {
+		DEFAULT = KottiButton.Type.DEFAULT,
+		PRIMARY = KottiButton.Type.PRIMARY,
+	}
+	export const typeSchema = createLooseZodEnumSchema(Type)
+
+	export const Size = {
```
carsoli, line 14: why a const?

packages/kotti-ui/source/kotti-split-button/types.ts
```diff
@@ -0,0 +1,51 @@
...
+		...KottiButton.Size,
```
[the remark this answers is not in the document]
carsoli: because of this confusion, you tried to re-export the enum but you couldn't destruct an enum inside so you made it a const

and the util we have relies on actually having a native `enum`

packages/kotti-ui/source/kotti-button/hooks/use-actions-tippy.ts
```diff
@@ -0,0 +1,73 @@
...
+import { Ref, computed, ref, watch } from 'vue'
+
+import { TIPPY_VERTICAL_OFFSET } from '../../constants'
+
+export const useActionsTippy = ({
+	isDisabled,
+	isLoading,
+}: {
+	isDisabled: Ref<boolean>
+	isLoading: Ref<boolean>
```
Isokaeder, line 12: It may be a good idea to watch on these and hide the tippy when the values change to true.

Isokaeder (changes requested):

Isokaeder:
For the secondary style, it is strange to me that the divider line between the buttons disappears when hovering the primary button. However, when hovering the secondary button, the primary line persists

Screenshot 2024-03-13 at 21 29 51

Just a tad inconsistent, if you ask me. If design wants that, so be it. But perhaps clarify with them.

santiagoballadares:
> For the secondary style, it is strange to me that the divider line between the buttons disappears when hovering the primary button. However, when hovering the secondary button, the primary line persists
> 
> Screenshot 2024-03-13 at 21 31 08 Screenshot 2024-03-13 at 21 32 22
> Just a tad inconsistent, if you ask me. If design wants that, so be it. But perhaps clarify with them.

@Isokaeder Design confirmed that this is the desired behavior.

PR #902, opened 2024-03-15
fix(build): make yoco and vue-use-tippy esm by default

Isokaeder:
We had to do this because:
    - nuxt@2 does not yet support reading "exports" from package json
    - for an unknown reason nuxt resolved vue to an esm build in vue-use-tippy leading to two instances of vue being loaded (dual package hazard)
    - Trying to fix the build issue itself yielded no results
    - after many experiments the best approximation of a fix was to turn all
      subpackages into modules

We also reduced the likelihood of this happening again by checking if the documentation builds without errors


FlorianWendelborn (approved):
LGTM

PR #907, opened 2024-04-03
fix(KtFieldToggle/RadioGroup): have unique id per input element

Isokaeder:
- for both components only one value was generated that was passed to all child input elements
- now we generate a unique id per child input element


FlorianWendelborn (approved):
LGTM

PR #915, opened 2024-05-13
ci(github): Add GitHub Actions Workflow for Automatic Release

FlorianWendelborn:


internals/scripts/source/publish.ts
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable no-console */
+import { $, semver } from 'bun'
+import { z } from 'zod'
+
+const packagesToConsider = [] as string[]
```
FlorianWendelborn, line 5: Right now there’s no packages to publish. They should be enabled on a case-by-case basis.

PR #923, opened 2024-05-28
fix(eslint-config): change json key order to natural, case-insensitive

Isokaeder:
- update dependencies
- remove debug-level
- add names to configuration objects


FlorianWendelborn (approved):
LGTM

PR #924, opened 2024-05-29
chore(legacy-branch): cherry pick relevant commits from master

Isokaeder:
Prepare legacy branch by cherry picking all relevant changes from master after `5.6.0`

- Change some metas (readme, gitignore)
- pin versions of kotti packages to avoid an eventual unwanted leak of new releases


FlorianWendelborn (approved):
LGTM

PR #928, opened 2024-06-04
refact(KtTable): turn all files in  kotti-table/logic into typescript

Isokaeder:
types have been purposefully left obtuse for now


FlorianWendelborn (approved):
LGTM

PR #931, opened 2024-06-05
fix(SSR): explicitly load KtNavbar as a component

Isokaeder:
A change in the template of NavBar.vue produced a very hard to trace down bug which caused the navbar to not render in the documentation. Although the source of this bug could not be made out, it seemed to have to do with how nuxt is resolving components. Adding an explicit import of KtNavbar in the file (and therefore not relying on the kotti-ui plugin for nuxt) fixes this.

As the file change that triggered this is very unrelated and it would be impossible to see this very unrelated effect in code review, a test was added that should at least stop the very same event to occur


FlorianWendelborn (approved):
LGTM

PR #943, opened 2024-06-11
fix(exports): expose svg folder

Isokaeder:
Tell node that this package supports importing svg files directly:

```ts
import charts from '@3yourmind/yoco/svg/charts.svg'
```


FlorianWendelborn (approved):
Seems good enough

PR #949, opened 2024-06-18
fix(KtTable): use attrs instead of domProps for specific attributes

Isokaeder:
Vue offers two different parameters to pass html attributes into rendered elements, `attrs` and `domProps`. But they are not interchangable, in fact each seems to accept a predefined set of attributes and will do nothing if the user passes an attribute that the paramenter does not accept. It is unclear what the logic behind this is, so the changes here have been found by experimentation

Edit:
After some investigation, this seems to be at least a hint at what is considered to be passed as `attrs` https://github.com/vuejs/vue/blob/13f4e7dc03e2caed900ac70ff8b8fe58dda45663/packages/server-renderer/src/util.ts#L27

### DIFF
--- a/packages/kotti-ui/source/kotti-table/components/TableBody.ts
+++ b/packages/kotti-ui/source/kotti-table/components/TableBody.ts
@@ -46,10 +46,6 @@ export const TableBody: any = {
 
 						return [
 							h(TableRow, {
-								domProps: {
-									// eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-									'data-test': `table:element:${row.number}:${row.title}`,
-								},
 								key,
 								props: {
 									row,
--- a/packages/kotti-ui/source/kotti-table/components/TableBodyExpandRow.ts
+++ b/packages/kotti-ui/source/kotti-table/components/TableBodyExpandRow.ts
@@ -18,9 +18,7 @@ export const TableBodyExpandRow: any = {
 				h(
 					'td',
 					{
-						domProps: {
-							colSpan,
-						},
+						attrs: { colSpan },
 						class: 'kt-table__expanded-cell',
 					},
 					[renderExpand?.(h, { row, data: row, rowIndex })],
--- a/packages/kotti-ui/source/kotti-table/components/TableBodyLoadingRow.ts
+++ b/packages/kotti-ui/source/kotti-table/components/TableBodyLoadingRow.ts
@@ -20,9 +20,14 @@ export const TableBodyLoadingRow: any = {
 		const { colSpan, render } = this
 
 		return h('tr', {}, [
-			h('td', { domProps: { colSpan }, class: 'kt-table__loader' }, [
-				render(h),
-			]),
+			h(
+				'td',
+				{
+					attrs: { colSpan },
+					class: 'kt-table__loader',
+				},
+				[render(h)],
+			),
 		])
 	},
 }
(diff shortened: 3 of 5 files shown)


packages/kotti-ui/source/kotti-table/components/TableBody.ts
```diff
@@ -46,10 +46,6 @@ export const TableBody: any = {
 
 						return [
 							h(TableRow, {
-								domProps: {
```
FlorianWendelborn, line 49: Seems like a bug?

packages/kotti-ui/source/kotti-table/components/TableBodyExpandRow.ts
```diff
@@ -18,9 +18,7 @@ export const TableBodyExpandRow: any = {
 				h(
 					'td',
 					{
-						domProps: {
-							colSpan,
```
FlorianWendelborn, line 22: https://github.com/vuejs/vue/blob/13f4e7dc03e2caed900ac70ff8b8fe58dda45663/packages/server-renderer/src/util.ts#L6

packages/kotti-ui/source/kotti-table/components/TableBodyLoadingRow.ts
```diff
@@ -20,9 +20,14 @@ export const TableBodyLoadingRow: any = {
 		const { colSpan, render } = this
 
 		return h('tr', {}, [
-			h('td', { domProps: { colSpan }, class: 'kt-table__loader' }, [
-				render(h),
-			]),
+			h(
+				'td',
+				{
+					domProps: { colSpan },
```
FlorianWendelborn, line 26: https://github.com/vuejs/vue/blob/13f4e7dc03e2caed900ac70ff8b8fe58dda45663/packages/server-renderer/src/util.ts#L6

packages/kotti-ui/source/kotti-table/components/TableRow.ts
```diff
@@ -231,9 +235,11 @@ export const TableRow: any = {
 								[
 									h('label', { class: 'form-checkbox' }, [
 										h('input', {
-											domProps: {
+											attrs: {
 												checked: isSelected,
 												disabled: isDisabled,
+											},
+											domProps: {
 												type: 'checkbox',
```
FlorianWendelborn, line 243: https://github.com/vuejs/vue/blob/13f4e7dc03e2caed900ac70ff8b8fe58dda45663/packages/server-renderer/src/util.ts#L14

FlorianWendelborn (changes requested):

packages/kotti-ui/source/kotti-table/components/TableBody.ts
Isokaeder: It's moved to TableRow.ts

packages/kotti-ui/source/kotti-table/components/TableBodyExpandRow.ts
Isokaeder: fixed and verified in a local build

packages/kotti-ui/source/kotti-table/components/TableBodyLoadingRow.ts
Isokaeder: fixed and verified in a local build

FlorianWendelborn (approved):
LGTM

PR #952, opened 2024-06-21
hack(vue-use-tippy): make sure tippy is always a function

Isokaeder:
In certain situations `import tippy from "tippy.js"` will not produce a function, because tippy will be imported as cjs when it should be imported as esm.

This hack tries to ensure correctness at runtime since a proper fix at build time has not been found and we prefer to keep both builds as export options.


FlorianWendelborn (approved):
LGTM

PR #954, opened 2024-06-24
chore(eslint): enforce default imports for dependencies that are not verifiably modern

Isokaeder:
After investigating https://github.com/3YOURMIND/kotti/pull/952 and similar occurrences with packages that may be resolved as CJS, we pinned down the reason in node and how they are trying to make sense of importing CJS from ESM.

Resources:
- https://nodejs.org/docs/v20.15.0/api/esm.html#interoperability-with-commonjs
- https://github.com/nodejs/node/tree/main/deps/cjs-module-lexer

tl;dr: When in doubt, use default imports for dependencies that either use commonjs only, or do not declare "exports" in their `package.json`

In this PR we therefore enforce default imports for many of our dependencies, excluding the ones where we are empircally convinced that they are ok (like `zod` or `vue`)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -5,6 +5,25 @@ import eslintConfig3YD from '@3yourmind/eslint-config'
 import { fileURLToPath } from 'url'
 import path from 'node:path'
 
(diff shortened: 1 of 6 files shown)


FlorianWendelborn (approved):
LGTM

PR #973, opened 2024-07-23
fix(element-ui): externalize translations

Isokaeder:
create helper function to handle eventual default keys when importing element-ui's translation objects


FlorianWendelborn (approved):
LGTM

Issue #970, opened 2024-07-18
KtCommentInput: force internal comments


RueRivoli:
**Is your feature request related to a problem? Please describe.**
In the new feature draft quotes, we want the people to leave only internal comments. 
BE receiving internal comments know that they need to send emails only to service users and not for the customer.

**Describe the solution you'd like**
Make the lock icon not switchable to the unlock icon (for public comment)

**Describe alternatives you've considered**
none

**Additional context**

PR #974, opened 2024-07-24
feature(#970) : force internal comments for KtCommentInput (and KtComment)

RueRivoli:
This merge request intends to block comments into the internal mode meaning that the user cannot switch to public.
This prop is meaningful only if the property allowInternal is true.

### DIFF
--- a/packages/kotti-ui/source/kotti-comment/KtComment.vue
+++ b/packages/kotti-ui/source/kotti-comment/KtComment.vue
@@ -25,7 +25,7 @@
 			/>
 			<KtCommentInput
 				v-if="userToReply"
-				v-bind="{ isInternal, placeholder, tabIndex, userAvatar }"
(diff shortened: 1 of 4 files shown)


packages/kotti-ui/source/kotti-comment/KtComment.vue
```diff
@@ -25,7 +25,7 @@
 			/>
 			<KtCommentInput
 				v-if="userToReply"
-				v-bind="{ isInternal, placeholder, tabIndex, userAvatar }"
+				v-bind="{ forceInternal, isInternal, placeholder, tabIndex, userAvatar }"
```
FlorianWendelborn, line 28: Where does this variable come from?

Isokaeder (approved):
FT okay

FlorianWendelborn (approved):
LGTM

Isokaeder:
This resolves #970

PR #977, opened 2024-07-24
dep(zod): lock zod version to v3.22.5

Isokaeder:
after initially locking to v3.23.6 didn't fix the original issue and some subsequent experimentation, v3.22.5 turned out to be the right choice

see also -> <https://github.com/colinhacks/zod/issues/3435>


FlorianWendelborn (approved):
LGTM

PR #984, opened 2024-08-07
test(kt-accordion): add snapshot test

santiagoballadares:


Isokaeder (approved):

FlorianWendelborn (approved):
LGTM, although I'm a bit worried about those data attributes changing. Let's see

PR #988, opened 2024-08-30
fix(kt-table-legacy): fix row selection reactivity not working

santiagoballadares:
Refact row selection and selection checkboxes styles to fix row selection reactivity not working.

Also improve KtTableLegacy documentation.


kpm-berlin (approved):

Isokaeder (approved):
LGTM

PR #995, opened 2024-09-16
chore(kotti-i18n): add crowdin translations

santiagoballadares:
- add config file
- cleanup spanish translations
- fix some keys sorting

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,7 @@ vite.config.ts.timestamp*
 
 # package.json tarball script
 /tarballs
+
+# crowdin (ignore all but the source file)
+packages/kotti-ui/source/locales/*.json
+!packages/kotti-ui/source/locales/input.json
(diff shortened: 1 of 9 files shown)


Isokaeder (approved):
I am not against merging this.

However, it seems like an in-between state. The translations are not used yet from actual code, there is no documentation about how to interact with crowdin, there is no steps documented within kotti how we plan to get there.

Maybe it's not necessary, there seems to be an agreement in the team that I am not aware of since it happened during my vacation. But for me it would be useful to have some list of todo items somewhere (gitlab issue, a comment on this PR, a markdown somewhere in the codebase, all are ok), so that we have a easy to understand state of this task.

santiagoballadares:
> I am not against merging this.
> 
> However, it seems like an in-between state. The translations are not used yet from actual code, there is no documentation about how to interact with crowdin, there is no steps documented within kotti how we plan to get there.
> 
> Maybe it's not necessary, there seems to be an agreement in the team that I am not aware of since it happened during my vacation. But for me it would be useful to have some list of todo items somewhere (gitlab issue, a comment on this PR, a markdown somewhere in the codebase, all are ok), so that we have a easy to understand state of this task.

@Isokaeder PR updated
1. The agreement was not to integrate Crowdin but to copy/paste translations from a Crowdin project to Kotti. This has not changed.
2. Added instructions to the README file

kpm-berlin (approved):

PR #996, opened 2024-09-17
feat(KtFileUpload): create KtFileUpload component

Isokaeder:
This is meant to replace KtFieldFileUploadRemote specifically, since it suffers from severe bugs since vue was updated to v2.7 and was hard to use even before that.

The PR has some open TODOs:
- The new component is missing proper documentation for the new component. 
- ~~`packages/kotti-ui/source/kotti-file-upload/types.ts` contains duplicated types with its counterparts from `kotti-field-file-upload`~~
- In general, one can consider to merge the two folders together, since KtFileUpload uses a lot of the components from `kotti-field-file-upload` - although it is not technically a field.

However, since this is urgent, I would propose to merge as is, I will deliver another PR amending the documentation as quickly as possible.


FlorianWendelborn (approved):
LGTM considering it’s urgent, no major issues just some minor stuff we should improve someday

PR #1022, opened 2024-11-22
fix(imports): use file path for lodash import

Isokaeder:
Note: we are using this import pattern `'lohash/xxx.js'` for all other imports as well. Not using it can lead to not found errors in certain setups.


santiagoballadares (approved):

FlorianWendelborn (commented):
LGTM

Issue #1005, opened 2024-10-17
Reveal/obscure password in KtFieldPassword


RueRivoli:
**Is your feature request related to a problem? Please describe.**
It would be great to have the functionality to reveal the password in a KtFieldPassword. 
With that, the user can check the value.

**Describe the solution you'd like**
In the new authentication pages, we have 2 options:
- show fields "password", "confirm password" 
- have one field "password" with an eye icon (for example) so that the user can check the value

**Describe alternatives you've considered**
/

**Additional context**
This would be helpful for the sign up and set password pages.

Isokaeder:
Closed by https://github.com/3YOURMIND/kotti/pull/1027

PR #1027, opened 2024-11-27
feature(#1005): Hide or reveal passwords

RueRivoli:
This MR aims at creating a click handler on KtField right icons so that KtFieldPassword can hide/show password when the user clicks on the eye icon.


Isokaeder:
I am wondering about why this feature affects all kotti fields, since this feature should only affect KtFieldPassword according to the issue #1005 

Wouldn't it be easier to add a boolean field to only KtFieldPassword that lets you opt in (or better opt out) of this toggle. Something like:

```html
<!-- Will have the eye icon to toggle the obscure/reveal password -->
<KtFieldPassword />

<!-- Will not have that toggle -->
<KtFieldPassword hideVisibilityToggle />
```

packages/yoco/svg/eye_slashed.svg
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<svg width="800px" height="800px" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Isokaeder, line 2: Those "800px" look a bit sus. Not sure if it has an actual effect, but I would change both to "24px" (as for most of yoco icons) just to be safe.

Isokaeder (approved):
LGTM

PR #1029, opened 2024-12-03
Kt standard table wip

santiagoballadares:


packages/documentation/pages/usage/components/standard-table.vue
```diff
@@ -169,13 +174,84 @@ export default defineComponent({
...
+				filterFn: (row, columnId, filterValue) =>
+					row.getValue(columnId) === filterValue,
+				getData: (row: TodoRow) => row.priority,
+				id: 'priority',
+				isSortable: true,
+				label: 'Priority',
+			},
+			{
+				display: { decimalPlaces: 2, type: 'numerical' },
+				filterFn: (row, columnId, filterValue) => {
```
santiagoballadares, line 188: `filterFn` is exposed in the `Column` type. Right now its implementation is on the user side. Could be moved into the standard table hook at the expense of leaving less flexibility in terms of the implementation/logic for the filters/values in the table.

packages/kotti-ui/source/kotti-table/standard-table/hooks.ts
```diff
@@ -48,9 +58,23 @@ export const useKottiStandardTable = <ROW extends AnyRow>(
 } => {
 	const params = computed(() => paramsSchema.parse(_params.value))
 
+	const appliedFilters = ref<KottiStandardTable.AppliedFilter[]>([])
+	/**
+	 * https://github.com/TanStack/table/discussions/4670
+	 */
+	const globalFilter = ref<KottiFieldText.Value>(' ')
```
santiagoballadares, line 65: Apparently Column filters aren't applied until the Global filter is applied. Maybe this was fixed on a more recent version of tanstack table

packages/kotti-ui/source/kotti-table/table/hooks.ts
```diff
@@ -54,6 +58,14 @@ export const paramsSchema = z
 	.object({
 		//TODO:
 		// actions : list of buttons (based on baseOptionSchema)
+		columnFilters: z
+			.object({
+				id: z.string(),
+				// Zod schema `z.unknown()` incorrectly infers the property as optional
```
santiagoballadares, line 64: Zod has issues with the unknown type

PR #1032, opened 2024-12-10
fix(KtPagination): disable back button on first page

Isokaeder:
this was introduced by making pagination 1-indexed


FlorianWendelborn (approved):
LGTM

PR #1034, opened 2024-12-15
fix(KtTable): fix styling issues with the table

Isokaeder:
- slide in sorting indicator if header is right-aligned
- add spacing to header to make it properly center-aligned if sorting indicator is present
- show inline loading state if rows are present
- put custom styles into td/th elements to support relative widths


Isokaeder:
- (optional) selected rows should have blue background
- action icons need bigger click area
- put indicator on the left if column is right aligned
- double indicator on hover if sortable but unsorted header is hovered

Isokaeder:
design for sorting indicator here: https://www.figma.com/proto/0yFVivSWXgFf2ddEF92zkf/Kotti-Design-System?node-id=7408-4784&t=KpfQ8pCwwv8UoQN1-1&scaling=min-zoom&content-scaling=fixed&page-id=95%3A0&starting-point-node-id=7408%3A4784

FlorianWendelborn (approved):
LGTM

PR #1041, opened 2025-01-16
build(vite): bump to latest

Isokaeder:


Isokaeder:
probably not a good idea to this before moving to vue3 because `@vitejs/vue2` is not maintained

Issue #1026, opened 2024-11-26
Provide icons chevron_down_double and chevron_up_double


RueRivoli:
**Is your feature request related to a problem? Please describe.**
We want to have more visual for the KtFieldSingleSelect options.
The specific usage would be the part priorities: very high, high, medium, low, very low

**Describe the solution you'd like**
Possibility to have icon in KtFieldSingleSelect options, eventually for KtFieldMultiSelect

**Describe alternatives you've considered**
none

**Additional context**
This is proposed to have among part priorities fields

RueRivoli:
This task is finally only about adding icons: "chevron_up_double" and "chevron_down_double".
This will enable us to have more distinct icons to display for part priorities: "which part to print first".

santiagoballadares:
#1044

PR #1044, opened 2025-01-17
feature(#1026): add chevron_down_double and chevron_up_double icons

RueRivoli:
This pull request consists in adding 2 icons.


santiagoballadares:
@RueRivoli LGTM but it needs:
- yoco version to be updated: packages/yoco/package.json
- kotti-ui dep version updated: packages/kotti-ui/package.json

packages/kotti-ui/package.json
```diff
@@ -98,5 +98,5 @@
 	"style": "./dist/style.css",
 	"type": "module",
 	"types": "./dist/index.d.ts",
-	"version": "8.3.1"
+	"version": "8.4.0"
```
santiagoballadares, line 101: This version is the Kotti-UI version, you don't need to update this unless you want a new Kotti version.

What is still missing, is the Yoco dependency here (in this file). It should match the new yoco version, so: `"@3yourmind/yoco": "^2.10.0"`

Isokaeder (approved):

packages/kotti-ui/package.json
Isokaeder: I felt inclined to take over and do that in @RueRivoli's stead.

PR #1056, opened 2025-02-11
fix(kt-table): unset successfullyDroppedColumnId after column re-order

santiagoballadares:
Unset successfullyDroppedColumnId after column re-order

### DIFF
--- a/packages/kotti-ui/source/kotti-table/KtTable.vue
+++ b/packages/kotti-ui/source/kotti-table/KtTable.vue
@@ -15,6 +15,7 @@
 						:data-test="header.dataTest"
 						:draggable="header.isDraggable"
 						:style="header.style"
+						@animationend="handleAnimationEnd"
 						@click="(e) => handleHeaderClick(e, header)"
 						@dragenter.prevent
 						@dragleave.prevent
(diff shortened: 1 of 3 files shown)


santiagoballadares:
@Isokaeder apparently there's an event `animationend` that could be used to clear `successfullyDroppedColumnId` but an event listener would be needed on every cell.

`
cellElement.addEventListener('animationend', () => {
  successfullyDroppedColumnId.value = null
})
`
What do you think?

Isokaeder:
> @Isokaeder apparently there's an event `animationend` that could be used to clear `successfullyDroppedColumnId` but an event listener would be needed on every cell.
> 
> `cellElement.addEventListener('animationend', () => { successfullyDroppedColumnId.value = null })` What do you think?

Yeah, that sounds a little cleaner. I would go about it like this:

- expose `unsetDroppedColumnId` method on the `internal` api of the table hook
```ts
const tableContext: TableContext<ROW, COLUMN_ID> = computed(() => ({
	// ...other exposed methods
	unsetDroppedColumnId: () => {
	        // avoid eventual redraws since method will be called once per row
		if (successfullyDroppedColumnId.value === null) return 
		successfullyDroppedColumnId.value = null
	},
}))
```
- handler function in `KtTable.vue`
```ts
const handleAnimationEnd = () => {
	tableContext.value.internal.unsetDraggedColumnId()
}
```
- add event to td element
```html
<td
	v-for="cell in row.cells"
	...
	@animationend="handleAnimationEnd"
	@dragenter.prevent
	@dragleave.prevent
	@dragover.prevent="(e) => handleCellDragOver(e, cell.columnId)"
>
```

Isokaeder (approved):
Works as expected. Approved!

PR #1062, opened 2025-02-17
chore(eslint): use namespace instead of module

santiagoballadares:


Isokaeder:
I remember the inverse change maybe from two years ago, where all `namespace` were replaced by `module`. As far as I remember the reason back then was that modules are the [recommended way to organize code](https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html#using-modules) according to the official typescript documentation, and that we don't use namespace specific features like merging them across files.

Is there a concrete reason now to revert this change?

Isokaeder:
Aye. Yeah, there seems to be a bit of confusion about module vs namespace. See also [here](https://typescript-eslint.io/rules/no-namespace/) and [here](https://typescript-eslint.io/rules/prefer-namespace-keyword) 

🤷‍♂️

Personally I don't care. I wish there was a singular or at least clear way to do namespaces. Let's take `namespace`, simply because at least then we can activate the [`@typescript-eslint/prefer-namespace-keyword`](https://typescript-eslint.io/rules/prefer-namespace-keyword) rope in the scope of this MR as well, and force some consistency.

Isokaeder (changes requested):
reactivate `@typescript-eslint/prefer-namespace-keyword` by removing line 127 in `eslint.config.ts`

Isokaeder (approved):

Issue #1059, opened 2025-02-14
Interactive via keyboard, focus state missing from KtFieldToggle, KtFieldToggleGroup, KtFieldRadioGroup


santiagoballadares:
**Is your feature request related to a problem? Please describe.**
`KtFieldToggle`, `KtFieldToggleGroup`, `KtFieldRadioGroup` do not have:

- Interactivity via keyboard
- Focus state

**Describe the solution you'd like**
To be defined.

**Describe alternatives you've considered**

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio

**Additional context**

PR #1016, opened 2024-11-18
Update styles of toggles, switches, and radios

Isokaeder:
- Figma for checkboxes/toggles: https://www.figma.com/design/0yFVivSWXgFf2ddEF92zkf/Kotti-Design-System?node-id=6815-8673&node-type=section&t=xaKVnXiFtzcOEdF0-0
- Figma for radios: https://www.figma.com/design/0yFVivSWXgFf2ddEF92zkf/Kotti-Design-System?node-id=6815-8672&node-type=section&t=xaKVnXiFtzcOEdF0-0
- Figma for switches: https://www.figma.com/design/0yFVivSWXgFf2ddEF92zkf/Kotti-Design-System?node-id=6815-8671&node-type=section&t=xaKVnXiFtzcOEdF0-0

PR #1064, opened 2025-02-20
feat(kotti-fields): add hover state to KtFields

santiagoballadares:


Isokaeder (approved):
Looks good.

Question: There is no hover state introduced for non-standard-looking inputs like toggles or radios. Is that a TODO, or is it not intended for those elements to have a hover state?

santiagoballadares:
> Looks good.
> 
> Question: There is no hover state introduced for non-standard-looking inputs like toggles or radios. Is that a TODO, or is it not intended for those elements to have a hover state?



> Looks good.
> 
> Question: There is no hover state introduced for non-standard-looking inputs like toggles or radios. Is that a TODO, or is it not intended for those elements to have a hover state?

Not implemented for `KtFieldToggle`, `KtFieldToggleGroup`, `KtFieldRadioGroup` because blur cannot happen without focus. These fields don't have focus state or interactivity via keyboard. Here's the issue https://github.com/3YOURMIND/kotti/issues/1059

I also saw this old PR https://github.com/3YOURMIND/kotti/pull/1016, not sure what its status is. 

I asked Design and agreed on leaving these fields for another follow-up PR.

Issue #1057, opened 2025-02-12
Change label of itemsPerPage selector


Isokaeder:
Now reads "Rows per page"

Should read "Items per page"

PR #1066, opened 2025-02-24
fix(#1057): change selector label to be more generic

Isokaeder:


Isokaeder:
noice. Then I can close this one.

PR #1067, opened 2025-02-24
fix(KtFileUpload): optionally hide file size

Isokaeder:
In some cases we want to not display file size or do not have access to it. This MR removes the necessity to pass file's size to file info array entries


santiagoballadares (approved):
LGTM

PR #1068, opened 2025-02-25
fix(kotti-fields): dont emit blur event when interacting with the con…

santiagoballadares:
…tent inside the field container

### DIFF
--- a/packages/kotti-ui/source/kotti-field/hooks.ts
+++ b/packages/kotti-ui/source/kotti-field/hooks.ts
@@ -3,6 +3,7 @@ import type { Ref } from 'vue'
 import {
 	computed,
 	inject,
+	onBeforeMount,
 	onMounted,
 	onUnmounted,
 	reactive,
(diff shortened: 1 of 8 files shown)


packages/kotti-ui/source/kotti-field/hooks.ts
```diff
@@ -348,3 +350,72 @@ export const useForceUpdate = (): {
...
+	field,
+	fieldTarget,
+	findEventTarget,
+	valuePlaceholder,
+}: {
+	emit: (event: 'blur', value: T | null) => void
+	field: KottiField.Hook.Returns<T>
+	fieldTarget: Ref<Array<HTMLElement | null> | HTMLElement | null>
+	findEventTarget?: (target: EventTarget | null) => EventTarget | null
+	valuePlaceholder?: T | null
```
Isokaeder, line 381: `valuePlaceholder` is not the best name. a placeholder would be something that would be given, if the original value is null, but this does the opposite. I'd call this `valueOverride` or `valueReplacement`

Isokaeder (approved):
LGTM except for naming issue

Isokaeder:
FT ok as well

PR #1077, opened 2025-03-10
feat(KtTable): make header toggle indeterminate if only some rows are selected

Isokaeder:
make proper use of now indeterminate toggle state

Screenshot 2025-03-10 at 14 01 27


santiagoballadares (approved):
LGTM

PR #1084, opened 2025-03-13
style(KtFieldNumber): make increment/decrement buttons smaller

Isokaeder:


santiagoballadares (approved):
LGTM

After FT, the buttons have the same dimensions as before (32px x 32px) so not smaller but I see less padding on the sides.

I also noticed that their data-test has the autogenerated form key. Probably it's not relevant for this PR.

Screenshot 2025-03-13 at 12 47 44

PR #1117, opened 2025-09-03
feat: allow KtPagination to have no total prop

Isokaeder:
This is for cases where it is not known how many pages are available

Note:
With this change I also did a small change to `KtStandardTable`, making it controllable if its pagination would have an upper bound or not by setting the `paginationOptions.rowCount` parameter to either a number or `null`.


santiagoballadares (approved):
LGTM

PR #1131, opened 2025-09-23
feat(yoco): add filled_square icon

santiagoballadares:


packages/yoco/source/index.ts
```diff
@@ -67,6 +67,7 @@ export namespace Yoco {
 		FILLED_CIRCLE_CHECK = 'filled_circle_check',
 		FILLED_CIRCLE_CROSS = 'filled_circle_cross',
 		FILLED_CIRCLE_MINUS = 'filled_circle_minus',
+		FILLED_SQUARE = 'filled_square',
```
Isokaeder, line 70: Not a deal breaker, but I noticed that our naming is a bit confusing when it comes to filled icons:

| sparse version | filled version |
| --- | --- |
| circle_check | filled_circle_check |
| circle_cross | filled_circle_cross |
| circle_minus | filled_circle_minus |
| <none> | filled_triangle_attention |
| bookmark | bookmark_solid |
| heart | heart_solid |
| star | star_solid |
| user | user_solid |
| <none> | filled_square ? |

Sometimes we use `filled_` prefix, sometimes we use `_solid` suffix to describe the variation.

Not gonna fix it, but I think in this case `square_solid` is more in line with the existing names. but either way it's a bit messy.

PR #1149, opened 2025-10-08
feat(KtFieldSelects): allow popup width to auto size via min-width an…

santiagoballadares:
…d max-width

### DIFF
--- a/packages/kotti-ui/package.json
+++ b/packages/kotti-ui/package.json
@@ -98,5 +98,5 @@
 	"style": "./dist/style.css",
 	"type": "module",
 	"types": "./dist/types/index.d.ts",
-	"version": "8.10.0"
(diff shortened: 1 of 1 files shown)


Isokaeder (changes requested):
I think the solution is reasonable. However, I am not convinced that introducing an extra prop brings real value. I could see a better approach of just having one modifier, `atLeastSameWidth` instead of `sameWidth`, that would basically force the popper width to be basically `max(160px, width_of(input_reference))` (pseudo code)

In other words, arguably the `sameWidth` modifier is flawed in that it forces the dropdown to be smaller than it should be if the input container is small. I think we can fix that bug without introducing a new prop.

As for the documentation, we can still make this bahavior testable by adding a switch that will set the max-width of the input as it already happens.

